PR #9, opened 2022-03-22
Types: Support Recast, Reactions and Follows

varunsrin:
* Add new types to Recast, Reactions and Follows
* Document types more clearly
* Refactor types and typeguards to make them easier to read


src/types/index.ts
```diff
@@ -0,0 +1,147 @@
...
+/** A message body that represents a public broadcast from a user. */
+export type CastAddMessageBody = {
+  replyTo?: string;
+  text: string;
+  type: 'cast-add';
+};
+
+/** A message body that removes a previous broadcast from a user. */
+export type CastDeleteMessageBody = {
+  targetCastHash: string;
```
varunsrin, line 79: self note: this should be a URI for consistency with Recast, Reaction, Follow bodies

src/types/index.ts
```diff
@@ -0,0 +1,147 @@
...
+// ===========================
+
+/** A Cast Message */
+export type Cast = SignedMessage<CastMessageBody>;
+
+export type CastMessageBody = CastAddMessageBody | CastDeleteMessageBody | CastRecastMessageBody;
+
+/** A message body that represents a public broadcast from a user. */
+export type CastAddMessageBody = {
+  replyTo?: string;
```
varunsrin, line 72: self note: should be named URI so it's clear that we are using a URI type here

varunsrin: also add a URI type that explains in some detail how a URI is generated or at least links to a spec.

danromero (commented):
Overall looks good. A couple of minor comments from me.

src/types/index.ts
```diff
@@ -0,0 +1,147 @@
...
+ * @sequence - the sequence number of the message
+ * @signedAt - the unix timestamp of when the message was signed
+ * @username - the Farcaster username of the user that signed the message.
+ */
+type Message<T = MessageBody> = {
+  body: T;
+  prevHash: string;
+  rootBlock: number;
+  // TODO: research whether this is more akin to a version vector.
+  sequence: number;
```
varunsrin, line 33: this is basically a type of simple logical clock called a  [Lamport Timestamp](https://en.wikipedia.org/wiki/Lamport_timestamp) 

it be called `index`, `sequence`, `clock`, `logicalClock` or a very literal `lamportTimestamp`

src/types/index.ts
```diff
@@ -0,0 +1,147 @@
...
+  signedAt: number;
+  username: string;
+};
+
+type MessageBody = RootMessageBody | CastMessageBody | CastRecastMessageBody | ReactionMesageBody | FollowMessageBody;
+
+// ===========================
```
varunsrin, line 40: if a root does not have a type, will indexers know where to place it when it is received?

src/types/index.ts
```diff
@@ -0,0 +1,176 @@
...
+ * @message - the message that is being signed.
+ * @hash - the keccak256 hash of the message.
+ * @schema - the schema of the message.
+ * @signature - the ecdsa signature of the hash of the message.
+ * @signer - the ethereum address whose private key was used to create the signature
+ */
+export type SignedMessage<T = MessageBody> = {
+  hash: string;
+  message: Message<T>;
+  schema: string; // TODO: should the schema be validated by nodes?
```
nickcherry, line 13: What are some example values of `schema`?

src/types/index.ts
nickcherry: What is the scenario where a root wouldn't have a type?

src/types/index.ts
```diff
@@ -0,0 +1,147 @@
...
+//  Cast Types
+// ===========================
+
+/** A Cast Message */
+export type Cast = SignedMessage<CastMessageBody>;
+
+export type CastMessageBody = CastAddMessageBody | CastDeleteMessageBody | CastRecastMessageBody;
+
+/** A message body that represents a public broadcast from a user. */
+export type CastAddMessageBody = {
```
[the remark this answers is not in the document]
nickcherry: I think that should be a client/back-end concern

src/types/index.ts
varunsrin: currently roots do not have a type and the node infers it when the next message is received (which has a type). 

explicit types are probably the right way to go for simplicity (at the expense of some storage)

src/types/index.ts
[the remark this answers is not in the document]
varunsrin: On reflection, I'm proposing that we add an explicit Attachment object. 

```typescript
type Attachments = {
   items: string[];  // upto 2 max
}
```

The casting client is responsible for stripping URL's from the message and adding them into the attachment field. If the user adds more than two it can give them a visual indicator of which two will be attached. The displaying client will add both attachments to the bottom of the post, irrespective of where the URLs were posted.



Pros: 

* Attachment type can be extended to hold metadata to display URI's in a special way (e.g. show summary view of this NFT instead of the detail view). 

* If there are many URI's, we give clients a way to explicitly denote which two URL's should be expanded (since we do not want to expand an infinite number of URI's). This can easily be increased to any number in the future, though decreasing it is harder. 

Cons: 

* More work for the casting client to structure the message correctly. 
* More rigidity / complexity in the protocol definition. 

@nickcherry  wdyt?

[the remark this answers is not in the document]
nickcherry: Firstly, I think I want to retract my previous suggestion that we make attachments a back-end/client concern. My primary reasoning is that the alternative would require us to embed all links into cast bodies, and URLs could be prohibitively long unless the client uses a URL shortener, which introduces centralization + fragility. 

That being said, along with Varun's points, I think I'm into the idea of an explicit `Attachments` array. I think this is what you meant, but just want to clarify that I think those attachments should always be strings at the protocol level. Then back-ends / clients can optionally upgrade those strings to objects with useful metadata (e.g. title, preview image). This approach would keep the protocol relatively simple + light, while giving clients flexibility in how they interpret those URLS. Another nice property is that it would give clients the ability to retroactively enhance URLs, relevant for example if someone shared a link referencing a relatively unknown protocol (e.g. `dan.something_other_than_eth`) that later became popular.

RE the attachment count, my bias would be to keep it simple and have either a hard limit (i.e. 1, 2, or 3) or no limit, rather than introducing complexity by letting user's configure client-specific display preferences that would only be applied if both the caster and the reader were using the same node (I think).

src/client.ts
```diff
@@ -53,8 +54,11 @@ class Client {
     const item = {
       message: {
         body: {
+          _attachments: { items: [] },
+          _text: text,
```
nickcherry, line 58: Why the `_` prefixes?

nickcherry: Ah...
> the text of the Cast, the underscore prefix indicates that it is not hashed into the message.

src/types/index.ts
varunsrin: nonce?

src/types/index.ts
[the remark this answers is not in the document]
varunsrin: 👍  - agreed on all points

varunsrin:
@nickcherry and @sds - I updated the schema prop based on your feedback. we've now: 

* merged type (which was in the body of each message) with schema (which was top level) 
* schema is now a typed string value for each unique version of the post. 

i think this is a nice improvement over the previous setup where we had both schema and type which were sort of duplicating data. thumbs up if you agree with thi s

PR #24, opened 2022-05-06
added lexicographical hash comparing for roots

koloz193:
per https://github.com/farcasterxyz/node-playground/issues/21 we want to devise an algorithm for predictable root hash processing. We compare hashes based on their ascii value taking the one with the lower hash.

### DIFF
--- a/src/engine/engine.root.test.ts
+++ b/src/engine/engine.root.test.ts
@@ -197,7 +197,47 @@ describe('addRoot', () => {
       expect(engine.getCastAdds(username)).toEqual([]);
     });
 
-    // test('succeeds if the root block is identical but lexicographical hash value is greater', async () => {});
+    test('succeeds if the root block is identical but lexicographical hash value is lower', async () => {
+      const transient = { transient: { privateKey: alicePrivateKey } };
+
+      const signerChange = {
+        blockNumber: 99,
+        blockHash: Faker.datatype.hexaDecimal(64).toLowerCase(),
+        logIndex: 0,
+        address: aliceAddress,
+      };
+
+      engine.addSignerChange('bob', signerChange);
+
+      const root1 = await Factories.Root.create(
+        {
+          data: { rootBlock: 100, username: 'bob', signedAt: Date.now() },
+        },
+        transient
+      );
+
+      const root2 = await Factories.Root.create(
+        {
+          data: { rootBlock: 100, username: 'bob', signedAt: Date.now() + 5 * 60 * 1000 },
+        },
+        transient
+      );
+
+      if (root1.hash > root2.hash) {
+        engine.addRoot(root1);
+        expect(engine.getRoot('bob')).toEqual(root1);
+
+        engine.addRoot(root2);
+        expect(engine.getRoot('bob')).toEqual(root2);
+      } else {
+        engine.addRoot(root2);
+        expect(engine.getRoot('bob')).toEqual(root2);
+
+        engine.addRoot(root1);
+        expect(engine.getRoot('bob')).toEqual(root1);
+      }
+    });
+
     // test('fails if the root block is identical but lexicographical hash value is lower', async () => {});
 
     test('fails if the root is a duplicate', async () => {
--- a/src/engine/engine.root.test.ts
+++ b/src/engine/engine.root.test.ts
@@ -11,6 +11,8 @@ describe('addRoot', () => {
   let root100: Root;
   let root110: Root;
   let root90: Root;
+  let root1: Root;
+  let root2: Root;
   let transient: { transient: { privateKey: string } };
 
   let alicePrivateKey: string;
@@ -36,6 +38,20 @@ describe('addRoot', () => {
     );
 
     root90 = await Factories.Root.create({ data: { rootBlock: 90, username: 'alice' } }, transient);
+
+    root1 = await Factories.Root.create(
+      {
+        data: { rootBlock: 100, username: 'bob', signedAt: Date.now() },
+      },
+      transient
+    );
+
+    root2 = await Factories.Root.create(
+      {
+        data: { rootBlock: 100, username: 'bob', signedAt: Date.now() + 5 * 60 * 1000 },
+      },
+      transient
+    );
   });
 
   // Every test should start with a valid signer for alice.
@@ -50,6 +66,8 @@ describe('addRoot', () => {
       address: aliceAddress,
     };
     engine.addSignerChange('alice', aliceRegistrationSignerChange);
+
+    engine.addSignerChange('bob', aliceRegistrationSignerChange);
   });
 
   describe('input validation: ', () => {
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -50,3 +50,24 @@ const removeProps = (obj: Record<string, any>): void => {
     });
   }
 };
+
+export const lexicographicalCompare = (a: string, b: string): number => {
```
varunsrin, line 54: looks great! 

1. this method could benefit from unit tests that thoroughly cover the behavior, in case someone wants to refactor it in the future. at the very least, it would be good to cover: 

- null inputs (either or both) 
- different length inputs
- tests with numeric, capital case and lowercase hexes 


2. can you add a comment describing the general behavior of the util?

yarn.lock
```diff
@@ -36,72 +36,72 @@
...
+  "resolved" "https://registry.npmjs.org/@babel/helper-environment-visitor/-/helper-environment-visitor-7.16.7.tgz"
+  "version" "7.16.7"
   dependencies:
     "@babel/types" "^7.16.7"
 
 "@babel/helper-function-name@^7.16.7":
-  version "7.16.7"
-  resolved "https://registry.yarnpkg.com/@babel/helper-function-name/-/helper-function-name-7.16.7.tgz#f1ec51551fb1c8956bc8dd95f38523b6cf375f8f"
-  integrity sha512-QfDfEnIUyyBSR3HtrtGECuZ6DAyCkYFp7GHl75vFtTnn6pjKeK0T1DB5lLkFvBea8MdaiUABx3osbgLyInoejA==
+  "integrity" "sha512-QfDfEnIUyyBSR3HtrtGECuZ6DAyCkYFp7GHl75vFtTnn6pjKeK0T1DB5lLkFvBea8MdaiUABx3osbgLyInoejA=="
```
varunsrin: is this yarn.lock change necessary?

src/engine/engine.root.test.ts
```diff
@@ -197,7 +197,47 @@ describe('addRoot', () => {
       expect(engine.getCastAdds(username)).toEqual([]);
     });
 
-    // test('succeeds if the root block is identical but lexicographical hash value is greater', async () => {});
+    test('succeeds if the root block is identical but lexicographical hash value is lower', async () => {
+      const transient = { transient: { privateKey: alicePrivateKey } };
+
+      const signerChange = {
```
varunsrin, line 203: do you need this here? i think it's being added in the `beforeEach` block

src/engine/engine.root.test.ts
```diff
@@ -197,7 +197,47 @@ describe('addRoot', () => {
...
+        blockNumber: 99,
+        blockHash: Faker.datatype.hexaDecimal(64).toLowerCase(),
+        logIndex: 0,
+        address: aliceAddress,
+      };
+
+      engine.addSignerChange('bob', signerChange);
+
+      const root1 = await Factories.Root.create(
```
varunsrin, line 212: one challenge with this approach is that when we generate the root, we don't know which one will have the lexicographically higher hash. so the current approach is to use the javascript compare to determine which one is higher, and then run the lexicographical compare and assert that they match. to your earlier point over text, it's probably not a good idea to rely on javascript's behavior here. 

an alternative approach could be to pre-generate the roots and hard code them in. this way we know exactly which hash value should be higher, and then we can assert that directly.

varunsrin (changes requested):

src/utils.test.ts
```diff
@@ -62,3 +62,50 @@ describe('hashFCObject', () => {
...
+    const cmp = lexicographicalCompare('hello', 'helloa');
+    expect(cmp).toBeLessThan(0);
+  });
+
+  test('compare strings second input has additional char', async () => {
+    const cmp = lexicographicalCompare('helloa', 'hello');
+    expect(cmp).toBeGreaterThan(0);
+  });
+
+  test('compare strings second input has additional char', async () => {
```
varunsrin, line 102: the description for this test in incorrect

src/engine/engine.root.test.ts
```diff
@@ -34,6 +38,20 @@ describe('addRoot', () => {
     );
 
     root90 = await Factories.Root.create({ data: { rootBlock: 90, username: 'alice' } }, transient);
+
+    root1 = await Factories.Root.create(
```
varunsrin, line 42: i would also add a comment line here explaining that you're setting different `signedAt` values to cause the hash to change, since it may not be obvious to the reader.

src/engine/engine.root.test.ts
```diff
@@ -48,6 +66,8 @@ describe('addRoot', () => {
       address: aliceAddress,
     };
     engine.addSignerChange('alice', aliceRegistrationSignerChange);
+
+    engine.addSignerChange('bob', aliceRegistrationSignerChange);
```
varunsrin, line 70: the design here has the root signed by the username `@bob` but using `@alice`'s keys to sign the roots (see transient property). we are also then adding a signer for bob with alice's address. this logically works but may be confusing to the reader, since it implies that different users should share signing keys, which is not the desired behavior.  

we can simplify this by: 

1. changing `root1` and `root2` to be have username === alice
2. changing their blocks to be much higher numbers (say 200, 210) so they replace any previously seeded blocsk
3. renaming them to `root200` and `root210` to maintain consistency with the naming scheme 
4. removing this signer change

i think the tests should still work correctly after this, and it avoids the use of bob entirely

src/engine/engine.root.test.ts
```diff
@@ -197,8 +217,37 @@ describe('addRoot', () => {
       expect(engine.getCastAdds(username)).toEqual([]);
     });
 
-    // test('succeeds if the root block is identical but lexicographical hash value is greater', async () => {});
-    // test('fails if the root block is identical but lexicographical hash value is lower', async () => {});
+    test('succeeds if the root block is identical but lexicographical hash value is lower', async () => {
```
varunsrin, line 220: i think this should say `lexicographical hash value is higher`

src/engine/engine.root.test.ts
```diff
@@ -19,7 +23,7 @@ describe('addRoot', () => {
     const keypair = await Factories.EthAddress.create({});
     alicePrivateKey = keypair.privateKey;
     aliceAddress = keypair.address;
-    const transient = { transient: { privateKey: alicePrivateKey } };
+    transient = { transient: { privateKey: alicePrivateKey } };
```
varunsrin, line 26: nit: do you need to remove the const here?

varunsrin (changes requested):
Looks great, thanks for all your work on this!

Need a few more changes before we can merge, please see comments for details.

src/engine/engine.root.test.ts
varunsrin: this should still say higher i think? either way, lets merge and fix later, it's a small nit

src/utils.test.ts
varunsrin: i'll fix this up post merge

varunsrin (approved):

varunsrin:
@koloz193 you'll need to sign your commits to merge into this repository, see [this guide](https://farcasterxyz.notion.site/Setting-Up-GPG-56fcd0aa909744e6ae942c7cc8996e37)

PR #34, opened 2022-06-18
feature: change signing algorithm to Ed25519, hashing algorithm to Blake2b

zachterrell57:


varunsrin (changes requested):
Looks great! 

I left some feedback on some refactors that would make this more readable and perhaps reduce some instances of async calls that can be safely avoided.

src/debugger.ts
```diff
@@ -151,10 +136,10 @@ const Debugger = {
         table.push({ [nodeName]: Debugger._visualizeState(messages, color) });
       }
 
-      nodeStatus.push(username);
-      nodeStatus.push(table.toString());
+      console.log(username);
```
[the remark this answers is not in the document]
zachterrell57: I reverted it but then cherry-picked the changes back onto my commit, I'm not sure why it didn't commit that change as well

src/engine/engine.reaction.test.ts
```diff
@@ -12,17 +14,18 @@ describe('mergeReaction', () => {
...
   let reaction: Reaction;
+  let transient: { transient: { privateKey: Uint8Array } };
   const subject = () => engine._getActiveReactions(username);
 
   beforeAll(async () => {
-    const keypair = await Factories.EthAddress.create({});
-    alicePrivateKey = keypair.privateKey;
-    aliceAddress = keypair.address;
+    const publicPrivateKeys = await generatePublicPrivateKeys([username]);
+    const privateKeyBuffer = publicPrivateKeys.get(username)?.get('privateKey') as Uint8Array;
```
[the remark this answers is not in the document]
zachterrell57: the ed library is very particular about wanting uint8 values, I tried to keep it all hex but the ed library rejected it

src/utils.ts
```diff
@@ -75,3 +79,25 @@ export const hashCompare = (a: string, b: string): number => {
...
+
+    name.set('privateKey', privateKey);
+    name.set('publicKey', publicKey);
+    publicPrivateKeys.set(instanceName, name);
+  }
+
+  return publicPrivateKeys;
+};
+
+export const convertToHex = async (text: Uint8Array): Promise<string> => {
```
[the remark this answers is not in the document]
zachterrell57: I want to say I tried this and it resulted in cryptic error from `ed` in relation to the byte sequence, so I decided to keep all functions within the `ed` ecosystem. Your overall point still stands, but in this particular instance having this function be async does not increase async calls anywhere else since every function that calls this method is async iirc

src/utils.test.ts
```diff
@@ -14,17 +15,18 @@ describe('hashFCObject', () => {
     },
   };
 
-  const keccakSimpleObject = '0xe65bac56b00ed02f3672b7f1e0ed5e77f7f0a56d51c0fbdf1af696ae140006fa';
+  const blake2bSimpleObject =
```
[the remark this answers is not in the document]
zachterrell57: Yes I replicated it in replit

yarn.lock
```diff
@@ -3122,6 +3122,7 @@
   "version" "27.5.1"
   dependencies:
     "@jest/core" "^27.5.1"
+<<<<<<< HEAD
```
sds, line 3125: This looks like a rebase conflict marker. One simple approach if you don't want to address is to remove your `yarn.lock` and just run `yarn install` again to generate a new one. This could result in other changes but given the nature of this repo that's likely fine.

src/client.ts
```diff
@@ -39,7 +35,7 @@ class Client {
     };
 
     item.hash = await hashMessage(item);
-    item.signature = await sign(item.hash, this.privateKey);
+    item.signature = await signEd25519(item.hash, this.privateKey);
```
sds, line 38: Something to consider to make this code more flexible to changes in signature implementation in future would be to make your `KeyPair` type a class instead, and define a ed25519 implementation (`Ed25519KeyPair extends KeyPair`) with a `sign`/`verify` methods that you can call.

This allows you to write the rest of the simulator code without the simulator needing to be aware of the underlying algorithm. All the simulator needs to do is work with `KeyPair`s which implement the `sign` and `verify` methods, and then the rest of the code takes care of itself.

sds (commented):
This is a much nicer refactoring, nice work! Minor comment about a lingering rebase conflict marker, and some thoughts on how to further refactor (but not a blocker).

src/client.ts
varunsrin: i like this suggestion a lot

src/engine/engine.cast.test.ts
```diff
@@ -12,17 +14,18 @@ describe('mergeCast', () => {
...
+  let transient: { transient: { privateKey: Uint8Array } };
   const subject = () => engine._getCastAdds(username);
 
   beforeAll(async () => {
-    const keypair = await Factories.EthAddress.create({});
-    alicePrivateKey = keypair.privateKey;
-    aliceAddress = keypair.address;
+    const keyPair = await generateEd25519KeyPair();
+    const privateKeyBuffer = keyPair.privateKey;
+    alicePrivateKey = await convertToHex(privateKeyBuffer);
```
varunsrin, line 23: nit: i would add some line breaks so this is a bit easier to parse

src/engine/index.ts
```diff
@@ -110,27 +111,26 @@ class Engine {
...
       const username = cast.data.username;
 
       const signerChanges = this._users.get(username);
       if (!signerChanges) {
         return err('mergeCast: unknown user');
       }
-
-      if (!this.validateMessage(cast).isOk()) {
-        return this.validateMessage(cast);
+      if (!(await this.validateMessage(cast)).isOk()) {
```
varunsrin, line 122: this is a bit hard to read since there are so many clauses, i would recommend splitting into a separate line and using the isErr() method to avoid the negation.  

```suggestion
     const isCastValidResult = await this.validateMessage(cast); 
      if (isCastValidResult.isErr()) {
```

src/engine/index.ts
```diff
@@ -110,27 +111,26 @@ class Engine {
...
-
       let castSet = this._casts.get(username);
       if (!castSet) {
         castSet = new CastSet();
         this._casts.set(username, castSet);
       }
-
       return castSet.merge(cast);
     } catch (e: any) {
+      console.error(e);
```
varunsrin, line 133: if you want to keep log lines because they are useful for debugging, i would add some method context so its easier for the reader to see. something like: 

```suggestion
      console.error(`engine.MergeCast` + e.toString());
```

src/engine/index.ts
```diff
@@ -167,8 +167,8 @@ class Engine {
         return err('mergeReaction: unknown user');
       }
 
-      if (!this.validateMessage(reaction).isOk()) {
-        return this.validateMessage(reaction);
+      if (!(await this.validateMessage(reaction)).isOk()) {
```
varunsrin, line 170: same comment as above about simplifying the clause

varunsrin (changes requested):

varunsrin (approved):

varunsrin:
nice work @zachterrell57!

PR #35, opened 2022-07-15
Verifications implementation v1

pfletcherhill:
Implements verifications according to Proposal A' in this notion doc: https://pfletcherhill.notion.site/Reconciling-Verification-and-Signer-message-schema-53e262b6aca74b0da6ecde8836bce828

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,4 +1,3 @@
-import { measureMemory } from 'vm';
 import * as FC from '~/types';
 import { hashMessage, signEd25519, convertToHex, hashFCObject } from '~/utils';
 
@@ -36,7 +35,7 @@ class Client {
       body: {
         blockHash: ethblockHash,
         schema: 'farcaster.xyz/schemas/v1/root',
-      } as FC.RootBody,
+      },
       rootBlock: ethBlockNum,
       signedAt: Date.now(),
       username: this.username,
@@ -113,7 +112,7 @@ class Client {
     return await hashFCObject({
       username: this.username,
       externalAddressUri,
-    } as FC.VerificationClaim);
+    });
   }
 
   async makeVerificationAdd(
@@ -129,7 +128,7 @@ class Client {
         externalSignature,
         externalSignatureType: 'secp256k1-eip-191',
         claimHash,
-      } as FC.VerificationAddBody,
+      },
       rootBlock: root.data.rootBlock,
       signedAt: Date.now(),
       username: this.username,
@@ -142,7 +141,7 @@ class Client {
       body: {
         schema: 'farcaster.xyz/schemas/v1/verification-remove',
         verificationAddHash: verificationAdd.hash,
-      } as FC.VerificationRemoveBody,
+      },
       rootBlock: root.data.rootBlock,
       signedAt: Date.now(),
       username: this.username,
--- a/src/client.ts
+++ b/src/client.ts
@@ -140,7 +140,7 @@ class Client {
     const message = await this.makeMessage({
       body: {
         schema: 'farcaster.xyz/schemas/v1/verification-remove',
-        verificationAddHash: verificationAdd.hash,
+        claimHash: verificationAdd.data.body.claimHash,
       },
       rootBlock: root.data.rootBlock,
       signedAt: Date.now(),
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,5 +1,5 @@
 import * as FC from '~/types';
-import { hashMessage, signEd25519, convertToHex } from '~/utils';
+import { hashMessage, signEd25519, convertToHex, hashFCObject } from '~/utils';
 
 class Client {
   publicKey: Uint8Array;
@@ -129,33 +129,34 @@ class Client {
     return item;
   }
 
-  makeVerificationClaim(externalAddressUri: FC.URI): FC.VerificationClaim {
-    return {
+  async makeVerificationClaimHash(externalAddressUri: FC.URI): Promise<string> {
+    return await hashFCObject({
       username: this.username,
       externalAddressUri,
-    };
+    } as FC.VerificationClaim);
   }
 
   async makeVerificationAdd(
-    claim: FC.VerificationClaim,
-    root: FC.Root,
-    signedClaim: string
+    externalAddressUri: FC.URI,
+    claimHash: string,
+    externalSignature: string,
+    root: FC.Root
   ): Promise<FC.VerificationAdd> {
     const schema = 'farcaster.xyz/schemas/v1/verification-add' as const;
     const externalSignatureType = 'secp256k1-address-ownership' as const;
     const signedAt = Date.now();
     const signer = await convertToHex(this.publicKey);
 
-    // TODO: what is rootBlock and why is it needed?
     const rootBlock = root.data.rootBlock;
 
     const message = {
       data: {
         body: {
           schema,
-          verificationClaim: claim,
-          externalSignature: signedClaim,
+          externalAddressUri,
+          externalSignature,
           externalSignatureType,
+          claimHash,
         },
         rootBlock,
         signedAt,
@@ -171,6 +172,10 @@ class Client {
 
     return message;
   }
+
+  // async makeVerificationRemove(): Promise<FC.VerificationRemove> {
+  //   // TODO
+  // }
 }
 
 export default Client;
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,3 +1,4 @@
+import { measureMemory } from 'vm';
 import * as FC from '~/types';
 import { hashMessage, signEd25519, convertToHex, hashFCObject } from '~/utils';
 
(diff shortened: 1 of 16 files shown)


src/simulator/basicSimulator.ts
```diff
@@ -77,6 +79,14 @@ class BasicSimulator extends Simulator {
...
 
+  async generateRandomVerification(client: Client, root: FC.Root): Promise<FC.VerificationAdd> {
+    const randomEthWallet = ethers.Wallet.createRandom();
+    const ethAddress = randomEthWallet.address;
+    const claimToSign = await client.makeVerificationClaimHash(ethAddress);
+    const externalSignature = await randomEthWallet.signMessage(claimToSign);
+    return client.makeVerificationAdd(ethAddress, claimToSign, externalSignature, root);
+  }
+
   async generateMessages(client: Client) {
```
pfletcherhill, line 90: I only added a verification to the basic simulator. Should I try adding them to others too?

src/utils.ts
```diff
@@ -43,11 +43,6 @@ export const signEd25519 = async (text: string, key: Uint8Array): Promise<string
   return convertToHex(signature);
 };
 
-/** Signs message with secp256k1 elliptic curve */
-export const signSecp256k1 = (text: string, key: utils.SigningKey): string => {
```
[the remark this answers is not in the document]
varunsrin: im probably going to use it to simulate signing from custody addresses, so i asked zach to leave it in there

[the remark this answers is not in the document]
pfletcherhill: Okay, I'll re-add.

Practically speaking though, a custody address is an ethereum address, so would it ever sign using this method instead of an EIP 191 supported approach?

src/simulator/index.ts
```diff
@@ -47,15 +47,19 @@ abstract class Simulator {
...
         node.mergeRoot(message);
       } else if (isReaction(message)) {
         node.mergeReaction(message);
-      } else {
+      } else if (isCast(message)) {
         node.mergeCast(message);
+      } else if (isVerification(message)) {
+        node.mergeVerification(message);
+      } else {
+        // TODO: decide how to handle unknown message type
```
varunsrin, line 62: if it's typed as a Message, and the branches cover all sub-types of messages, aren't we guaranteed that this won't happpen?

src/sets/verificationsSet.ts
```diff
@@ -0,0 +1,100 @@
...
+   * Private Methods
+   */
+
+  private add(message: VerificationAdd): Result<void, string> {
+    if (!isVerificationAdd(message)) {
+      return err('VerificationsSet.add: invalid message format');
+    }
+
+    if (this._deletes.get(message.hash)) {
+      return err('VerificationsSet.add: verification was deleted');
```
varunsrin, line 53: nit: "is already deleted" to match the language below

src/client.ts
```diff
@@ -18,115 +19,135 @@ class Client {
...
-    item.hash = await hashMessage(item);
-    item.signature = await signEd25519(item.hash, this.privateKey);
-
-    return item;
+  async makeRoot(ethBlockNum: number, ethblockHash: string): Promise<FC.Root> {
+    const message = await this.makeMessage({
+      body: {
+        blockHash: ethblockHash,
+        schema: 'farcaster.xyz/schemas/v1/root',
+      } as FC.RootBody,
```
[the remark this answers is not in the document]
pfletcherhill: No, you don't, you're right. Will clean these up.

src/client.ts
```diff
@@ -18,115 +19,135 @@ class Client {
...
     };
+    const message = await this.makeMessage(messageData);
+    return message as FC.Reaction;
+  }
+
+  async makeVerificationClaimHash(externalAddressUri: FC.URI): Promise<string> {
+    return await hashFCObject({
+      username: this.username,
+      externalAddressUri,
+    } as FC.VerificationClaim);
```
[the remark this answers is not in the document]
pfletcherhill: This method returns a promise of the hash of the verification claim (not the claim object), but I can remove the `as FC.VerificationClaim`.

src/client.ts
```diff
@@ -18,115 +19,135 @@ class Client {
...
+        externalAddressUri,
+        externalSignature,
+        externalSignatureType: 'secp256k1-eip-191',
+        claimHash,
+      } as FC.VerificationAddBody,
+      rootBlock: root.data.rootBlock,
+      signedAt: Date.now(),
+      username: this.username,
+    });
+    return message as FC.VerificationAdd;
```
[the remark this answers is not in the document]
pfletcherhill: When I remove this typescript complains with `Message<Body> is not assignable to type `VerificationAdd`. Because `makeMessage` returns `Promise<FC.Message<FC.Body>>`, I think I have to be explicit about `FC.VerificationAdd` here. Do you have another idea/recommendation?

src/client.ts
```diff
@@ -18,115 +19,135 @@ class Client {
...
+    claimHash: string,
+    externalSignature: string,
+    root: FC.Root
+  ): Promise<FC.VerificationAdd> {
+    const message = await this.makeMessage({
+      body: {
+        schema: 'farcaster.xyz/schemas/v1/verification-add',
+        externalAddressUri,
+        externalSignature,
+        externalSignatureType: 'secp256k1-eip-191',
```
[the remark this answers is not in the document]
pfletcherhill: I added it to [this PR](https://github.com/farcasterxyz/protocol/pull/16), though do you want me to write it in notion too?

src/simulator/index.ts
pfletcherhill: At this moment, yeah. But I think it's still good practice to have some sort of comment here for when new message types are added and we haven't updated this function yet.

src/client.ts
```diff
@@ -18,115 +19,135 @@ class Client {
...
+    const message = await this.makeMessage({
+      body: {
+        schema: 'farcaster.xyz/schemas/v1/verification-remove',
+        verificationAddHash: verificationAdd.hash,
+      } as FC.VerificationRemoveBody,
+      rootBlock: root.data.rootBlock,
+      signedAt: Date.now(),
+      username: this.username,
+    });
+    return message as FC.VerificationRemove;
```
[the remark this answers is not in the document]
pfletcherhill: See reply above

varunsrin (changes requested):

src/client.ts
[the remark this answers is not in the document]
varunsrin: nope, that's good enough

src/client.ts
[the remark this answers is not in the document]
varunsrin: ah i forgot that returned a less concrete type, this is fine

src/sets/verificationsSet.ts
```diff
@@ -0,0 +1,100 @@
...
+  private add(message: VerificationAdd): Result<void, string> {
+    if (!isVerificationAdd(message)) {
+      return err('VerificationsSet.add: invalid message format');
+    }
+
+    if (this._deletes.get(message.hash)) {
+      return err('VerificationsSet.add: verification is already deleted');
+    }
+
+    if (this._adds.get(message.hash)) {
```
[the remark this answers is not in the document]
varunsrin: Your solution seems better than mine since it doesn't cause state divergence. 

But is there a way we can extend your solution to avoid duplicates? That would be a much better experience for clients and also stop people from accidentally flooding their hubs with multiple proofs.

src/simulator/basicSimulator.ts
varunsrin: i think we should add them to all, though you can do it in a followup PR if you prefer.

varunsrin:
we should also wait for @sds to review before merging this in.

src/sets/verificationsSet.ts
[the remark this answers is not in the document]
pfletcherhill: Implemented a potential solution here: https://github.com/farcasterxyz/hub/pull/35/commits/cd51cfef44cb7bcfcbd00165bc6301d44f03fcb4

Summary:
* Use `claimHash` as index for both adds and deletes sets
* Prevents duplicate claims (i.e. verifications of same address)
* Hides external address once deleted because `VerificationRemove` only contains hash
* Updated set `add` and `delete` logic to check `signedAt` timestamp and give preference to latest add or delete
* User can still add a verification, delete it, then re-add the same address later on

Only downside I can think of is that the address isn't as hidden as it would be if deletes were indexed by the message hash, because someone with the address could hash it and check the set.

src/sets/verificationsSet.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    return err('VerificationsSet.merge: invalid message format');
+  }
+
+  /**
+   * Private Methods
+   */
+
+  private add(message: VerificationAdd): Result<void, string> {
+    if (!isVerificationAdd(message)) {
```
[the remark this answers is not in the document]
pfletcherhill: Fair, I'll remove it

src/sets/verificationsSet.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    return err('VerificationsSet.merge: invalid message format');
+  }
+
+  /**
+   * Private Methods
+   */
+
+  private add(message: VerificationAdd): Result<void, string> {
+    const {
```
varunsrin, line 53: nit: since we only really need claimHash, signedAt and message, would it be easier to read if we avoided destructuring and instead did something like 

```
const claimHash = message.data.body.claimHash
const signedAt = message.data.signedAt
...
```

src/sets/verificationsSet.test.ts
```diff
@@ -0,0 +1,207 @@
...
+        expect(adds()).toEqual([add2]);
+      });
+
+      test('fails with an earlier timestamp than existing add message', async () => {
+        expect(set.merge(add2).isOk()).toBe(true);
+        expect(set.merge(add1).isOk()).toBe(false);
+        expect(adds()).toEqual([add2]);
+      });
+
+      describe('with same timestamp', () => {
```
varunsrin, line 75: is it possible to get a claim with the same timestamp, same claim hash and a different hash?

varunsrin (approved):
Left minor comments, LGTM otherwise

src/engine/index.ts
```diff
@@ -186,6 +194,51 @@ class Engine {
...
+  }
+
+  /** Get claimHashes of known active verifications for a username */
+  getVerificationClaimHashes(username: string): string[] {
+    const verificationsSet = this._verifications.get(username);
+    return verificationsSet ? verificationsSet.getClaimHashes() : [];
+  }
+
+  /** Get claimHashes of all known verifications for a username */
+  getAllVerificationClaimHashes(username: string): string[] {
```
sds, line 214: Where do we see this method and `getVerificationClaimHashes` being used in practice?

src/engine/index.ts
```diff
@@ -369,6 +445,39 @@ class Engine {
     // TODO: validate targetUri, schema
     return ok(undefined);
   }
+
+  private async validateVerificationAdd(message: VerificationAdd): Promise<Result<void, string>> {
+    const { externalAddressUri, externalSignature, externalSignatureType, claimHash } = message.data.body;
+
+    if (externalSignatureType !== 'secp256k1-eip-191')
```
sds, line 452: I might be missing something, but we're not actually implementing EIP 191 signatures in this PR, right? (i.e. we're not [prepending with `0x19`](https://eips.ethereum.org/EIPS/eip-191#specification) etc.)

What does storing signatures in EIP 191 get us? I realize the spec is short and I have done a cursory read, but curious how you're thinking about its usage here.

sds (commented):
Looking great so far. Minor comments/questions inline.

src/simulator/basicSimulator.ts
pfletcherhill: Yeah, I'll do a separate simulation PR.

src/engine/index.ts
varunsrin: I'll let @pfletcherhill address the first question. 

For the second one, mist wallets generate EIP-191 signatures if I'm not mistaken. So we will *have* to assume that they are in that form.

src/client.ts
```diff
@@ -18,115 +18,135 @@ class Client {
...
-        },
-        rootBlock,
-        signedAt,
-        username: this.username,
+    const messageData = {
+      body: {
+        active,
+        // TODO: When we implement URI generation, this should be generated from the cast
+        targetUri: targetCast.hash,
+        type: 'like' as const,
```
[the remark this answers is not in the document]
pfletcherhill: Because `ReactionBody.type` in this case (or anytime we have a `schema` attribute) is a string literal type. So we have to reassure typescript that a variable like this one won't change by adding `as const`.

src/engine/index.ts
pfletcherhill: I initially added these methods because other sets had similar methods elsewhere in the engine. But I'd imagine that these would be useful for a client to list the ethereum addresses that an account has verified. They'd get all the claim hashes via `getVerificationClaimHashes` and then map over them with `getVerification` to get the addresses.

src/engine/index.ts
pfletcherhill: I thought the `signMessage` and `verifyMessage` methods from ethers (which we're using) did implement EIP 191. But I'll look into that and confirm.

pfletcherhill: The goal for using EIP 191 is to support as many external Ethereum wallets as possible. I think most wallets support EIP 191 now, so if we use the same standard for verifying the external signature we can avoid worrying about which wallet the user has for their external address.

src/sets/verificationsSet.test.ts
pfletcherhill: Hmm the only reason I can think of would be the account signs the same claim using two different external wallets, which implement different versions of EIP 191. So there may be two valid yet different `externalSignature` values.

src/sets/verificationsSet.ts
pfletcherhill: We use `message.hash` as well, so it would be three lines. I prefer destructuring, but it's up to you.

varunsrin: no strong opinion here, will resolve

src/engine/index.ts
pfletcherhill: Upon further investigation, ethers (and every major wallet) implements EIP 191 version `0x45` (i.e. prepending `"\x19Ethereum Signed Message:\n" + len(message)` before signing the message). EIP 191 versions `0x00` and `0x01` are different, and ethers does not support those in `verifyMessage`. So I suggest we change this type to `eip-191-0x45` or `eip-191-personal-sign` to be extra clear. @varunsrin or @sds, have a preference?

varunsrin: no strong preference, but slight lean towards 0x45 for precision

sds: Supportive of the shorter form as well for conciseness reasons as well. Thanks for digging in!

src/sets/verificationsSet.test.ts
varunsrin: good point. in that case, i think we should explicitly fail non EIP 191 signatures, either in the set or the engine. it prevents downstream clients from having to test for both.

varunsrin (changes requested):

src/sets/verificationsSet.test.ts
pfletcherhill: So now that we're restricting it to EIP 191 version `0x45`, all valid external signatures of the same claim should be identical. Do you think it's better to remove this test and the block in `VerificationSet` that handles it? Or leave it knowing there may be a future change (supporting more external signature types) that makes it relevant again?

varunsrin (approved):

PR #36, opened 2022-07-15
Add SignerSet implementation

drshrey:
**Context:**
https://farcasterxyz.notion.site/Signer-Set-85706ad04aea4bb6a3ac4e30f07bcd09

**What this is** 
- Initial SignerSet implementation
- SignerSet unit tests

**What this is not**
- SignerSet integrated with Hub


package.json
```diff
@@ -40,6 +41,7 @@
     "lint-staged": "^12.3.6",
     "nodemon": "^2.0.15",
     "prettier": "^2.6.0",
+    "secp256k1": "^4.0.3",
```
varunsrin, line 44: does ethereum

varunsrin, line 44: does `ethereum-cryptography`, which we already use, have a secp256k1 signing system that we can use instead of adding a new lib?

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,191 @@
...
+    this.delegates.push(child);
+  }
+
+  removeDelegate(pubkey: string): boolean {
+    return false;
+  }
+}
+
+class Signer {
+  public custodyAddressRoot: SignerNode;
```
varunsrin, line 70: nit: shall we call this the custodyAddressSigner instead of root?

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,191 @@
...
+  private signers: Signer[];
+
+  constructor() {
+    this.signers = [];
+  }
+
+  // Verification is done by the Ethereum blockchain i.e. the on-chain register/transfer event
+  // that triggers the addition of a Signer is proof the custody address being added owns the associated
+  // account id
+  public addSigner(custodyAddressPubkey: string): boolean {
```
varunsrin, line 144: prefer the nomenclature `addCustodySigner` and `addDelegateSigner` or just omitting the Signer part entirely

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,191 @@
...
+  EcdsaSecp256k1 = 'ecdsa-secp256k1',
+  Ed25519 = 'ed25519',
+}
+
+export enum HashAlgorithm {
+  Keccak256 = 'keccak256',
+  Blake2b = 'blake2b',
+}
+
+class SignerNode {
```
varunsrin, line 51: would we want to design this such that we can traverse back up the tree? or do you think that's not going to be required with the logic we have?

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,191 @@
+import { Signature } from 'ethers';
+
+export interface SignerAddition {
```
pfletcherhill, line 3: The rest of the codebase doesn't use the `{ message: {...}, envelope: {...} }` schema yet. I'd look at the `types/index.ts` file and define these new types as extensions of existing ones (i.e. `Message` and `Body`). Then it will be easier in the future for someone to migrate all types that extend `Message` to this schema at once.

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,191 @@
...
+  message: {
+    body: {
+      parentKey: string;
+      childKey: string;
+      schema: string;
+    };
+    account: number;
+  };
+  envelope: {
```
pfletcherhill, line 12: A reminder that the envelope attributes should be standardized across all message types. So these would be `hash`, `signature`, and `signer` in the current implementation of `Message` and represent the parent's signature. The child's signature, hash, etc should be included in the body.

src/sets/signerSet/signerSet.ts
pfletcherhill: Nit: in [my verifications PR](https://github.com/farcasterxyz/hub/pull/35) I called the new verification message `VerificationAdd`, so we should change this to `SignerAdd` or change both to `<...>Addition`.

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,191 @@
...
+    hash: string;
+    hashType: HashAlgorithm;
+    parentSignature: string;
+    parentSignatureType: SignatureAlgorithm;
+    parentSignerPubkey: string;
+  };
+}
+
+export enum SignatureAlgorithm {
```
pfletcherhill, line 41: Would be nice to move this to `types/index.ts` and use it across other message types

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,191 @@
...
+        return node;
+      }
+    }
+
+    return null;
+  }
+}
+
+class SignerSet {
+  private signers: Signer[];
```
pfletcherhill, line 135: I'm not sure an array is the best data structure here. Maybe something like this instead for easier lookups? (This is mostly borrowed from the notion spec and other sets in the codebase.)
```
adds : Map<signerPubKey, SignerAdd>; // childPubKey from SignerAdd
deletes: Map<signerPubKey, SignerRemove>;
edges: Map<parentPubKey, Set<childPubKey>>;
custodySigners: Set<signerPubKey>;
```
`custodySigners` gives you the roots. You can look up any signer in the tree (and its parent) via `adds` and check that signers haven't already been removed via `deletes`. `edges` gives you children for a given parent.

You'd have to keep `edges` in sync with the `SignerAdd` and `SignerRemove` messages stored in `adds` and `deletes`.

package.json
drshrey: Ah yes, https://github.com/ethereum/js-ethereum-cryptography#secp256k1-curve; will change to use this.

src/sets/signerSet/signerSet.ts
drshrey: `SignerAdd` sounds good and that's a good idea, I'll follow that type extension pattern.

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+    // generate custodyAddressPubkey
+    const rootKey2 = newSecp256k1Key();
+    const signerRootPubkey2 = secp.getPublicKey(rootKey2);
+    const signerRootEncodedPubkey2 = Buffer.from(signerRootPubkey2.toString()).toString('base64');
+
+    expect(signerSet.addSigner(signerRootEncodedPubkey2)).toEqual(true);
+    expect(signerSet.numSigners()).toEqual(2);
+  });
+
+  test('fails when same root is tried to be added twice', async () => {
```
sds, line 30: I don't have a strong opinion in the context of the simulator, but what are your thoughts on making the addition of a signer idempotent rather than fail when adding a signer that was already added?

Is there an advantage to failing (instead of being a no-op) when the same root is added twice?

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+        childSignatureType: SignatureAlgorithm.EcdsaSecp256k1,
+        childSignerPubkey: childEncodedPubkey,
+      },
+    };
+
+    const addWorked = signerSet.addDelegate(signerAddition);
+    expect(addWorked).toEqual(true);
+  });
+
+  test('fails when delegate exists in another signer', async () => {
```
sds, line 90: Regarding Varun's early comment from Slack about having all operations be local to a single account, some questions:

- Will we fail if the delegate exists in another delegate signer for a different account? Or only if in the same account?
- Given the potential for messages to arrive out-of-order on different hubs, we need a way to determine a [total ordering](https://en.wikipedia.org/wiki/Total_order) so that both hubs converge to the same end state. Does the current logic ensure that the "correct" delegated signer "wins"? (it's possible the spec does not adequately specify who should win in this context, though I _think_ the winner is currently defined as whichever message with a higher lexicographic hash).

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+
+    const removeWorked = signerSet.removeDelegate(signerRemove);
+    expect(removeWorked).toEqual(true);
+
+    // This will fail since delegate has been revoked
+    addWorked = signerSet.addDelegate(signerAddition);
+    expect(addWorked).toEqual(false);
+  });
+
+  test('fails when child is an existing node', async () => {
```
sds, line 279: Related to an earlier question, but should this fail or simply be idempotent? (I'm personally not sure what makes the most sense in the context of the simulator)

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+        parentSignatureType: SignatureAlgorithm.EcdsaSecp256k1,
+        parentSignerPubkey: signerRootEncodedPubkey,
+      },
+    };
+
+    const removeWorked = signerSet.removeDelegate(signerRemove2_1);
+    expect(removeWorked).toEqual(false);
+  });
+
+  test('fails because delegate has been revoked', async () => {
```
sds, line 426: Repeating just for sake of discussion and my own notes—no need to respond separately.

Should this case simply be a no-op?

sds (commented):
Some minor comments and questions—please don't consider blockers, but more topics of potential conversation,

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    return false;
+  }
+
+  // TODO - optional
+  public revokePastSigners(): boolean {
+    return false;
+  }
+
+  public numSigners(): number {
```
varunsrin, line 226: nit: is this just used for testing? if so, we should prefix it with an underscore just like we do in the other signers

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,235 @@
...
+    return false;
+  }
+
+  // Verification is done by the Ethereum blockchain i.e. the on-chain register/transfer event
+  // that triggers the addition of a Signer is proof the custody address being added owns the associated
+  // account id
+  public addSigner(custodyAddressPubkey: string): boolean {
+    if (this.pubkeyExistsInSignerSet(custodyAddressPubkey)) {
+      console.error('node with key value already exists in SignerSet');
+      return false;
```
varunsrin, line 165: prefer using the `Result` type when returning from all public set functions since it allows specifying an error string. you can see an example of this in the castSet

this is particularly useful in tests where you can assert that the error matches the error you expect, and not some other error condition.

src/sets/signerSet/signerSet.test.ts
varunsrin, line 90: nit: "in another signer" is confusing terminology -- what we are checking is that it does not exist in another custody signer's tree

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+        childSignatureType: SignatureAlgorithm.EcdsaSecp256k1,
+        childSignerPubkey: childEncodedPubkey,
+      },
+    };
+
+    addWorked = signerSet.addDelegate(signerAddition);
+    expect(addWorked).toEqual(false);
+  });
+
+  test('fails when claimed parent does not exist', async () => {
```
varunsrin, line 167: should we have separate tests for when the parent is a custody (tree root) and delegate? or is the logic similar enough that it isn't needed?

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+
+    // generate custodyAddressPubkey
+    const rootKey = newSecp256k1Key();
+    const signerRootPubkey = secp.getPublicKey(rootKey);
+    const signerRootEncodedPubkey = Buffer.from(signerRootPubkey.toString()).toString('base64');
+
+    signerSet.addSigner(signerRootEncodedPubkey);
+    expect(signerSet.numSigners()).toEqual(1);
+
+    const childKey = newSecp256k1Key();
```
varunsrin, line 178: nit: might be useful to make a helper that just returns an encoded key and call it in a single line, since this is done a lot in tests

src/sets/signerSet/signerSet.test.ts
varunsrin, line 426: is `revoke` a different action from `remove` ?

src/sets/signerSet/signerSet.ts
varunsrin: as discussed in our chat @drshrey, I think you'll have to go with @pfletcherhill's approach (or something similar) to get the conflict resolution logic working

src/sets/signerSet/signerSet.ts
drshrey: Sounds good and agreed on omitting `Signer` suffix. In the context of the current implementation, it's redundant with the `SignerSet` term (there is nothing other than Signers that we should be storing in there for now).

src/sets/signerSet/signerSet.ts
drshrey: resolving since this is now outdated with the current graph design

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+import { randomBytes } from 'crypto';
+const secp = require('ethereum-cryptography/secp256k1');
+
+const FarcasterSchemaUrl = 'farcaster.xyz/schemas/v1/signer-authorize';
+
+describe('create signer set', () => {
+  test('happy path', async () => {
+    const signerSet = new SignerSet();
+    // generate custodyAddressPubkey
+    const rootKey = newSecp256k1Key();
```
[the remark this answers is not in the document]
drshrey: Got it, I'll go with `custodySigner` since that's what I call it in the core logic

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
...
+/* eslint-disable no-restricted-imports */
+import SignerSet, { SignerAdd, SignatureAlgorithm, HashAlgorithm, SignerRemove } from './signerSet';
+import { blake2b } from 'ethereum-cryptography/blake2b';
+import { randomBytes } from 'crypto';
+const secp = require('ethereum-cryptography/secp256k1');
+
+const FarcasterSchemaUrl = 'farcaster.xyz/schemas/v1/signer-authorize';
+
+describe('create signer set', () => {
+  test('happy path', async () => {
```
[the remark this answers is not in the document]
drshrey: Agreed, will change these test labels to be more descriptive of the actual test itself

src/sets/signerSet/signerSet.ts
drshrey: Agreed and this is done!

drshrey: The one change I made from the above spec is naming deletes to `revoked` to make it explicit what the nature of that set's nodes contains (revoked elicits a more direct connection of how the node got there vs "deleted")

src/sets/signerSet/signerSet.ts
drshrey: Done (and yep, just used for testing)

src/sets/signerSet/signerSet.test.ts
```diff
@@ -0,0 +1,503 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
varunsrin, line 1: why is this necessary?

src/sets/signerSet/signerSet.ts
varunsrin: +1, i'd at least move its location to types/ in this PR

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+/*   
+  SignerSet manages the account's associated custody addresses authorized 
+  for Signing and their corresponding delegates.
+*/
+class SignerSet {
+  // new code
+  /* adds is a Map of (delegatePubkey, SignerAdd) */
+  private adds: Map<string, SignerAdd>;
+  /* revoked is a Map of (delegatePubkey, SignerRemove) */
+  private revoked: Map<string, SignerRemove>;
```
varunsrin, line 61: prefer the terminology remove to revoke -- it lines up with how people describe CRDTs

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+  /* edges is a Map of (parentPubkey (delegate or custody signer), Set<childPubkey>) */
+  private edges: Map<string, Set<string>>;
+  /* custodySigners is a set of custody signer pubkey values */
+  private custodySigners: Set<string>;
```
varunsrin, line 65: remember that when we introduce the custody signer revocation message, we will need to revoke certain signers in causal order (i.e. block order) 

so we may need to store these in an array or at least in a form that preserves the blocknumber. 

if this is too big a refactor to finish tomorrow, fine to avoid but maybe call this out in the comments.

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  /* addCustody adds a custody signer  */
+  public addCustody(custodySignerPubkey: string): Result<void, string> {
+    if (this._nodeWithPubkeyExists(custodySignerPubkey)) {
+      return err(`${custodySignerPubkey} already exists`);
+    }
+
+    this.custodySigners.add(custodySignerPubkey);
+    this.edges.set(custodySignerPubkey, new Set<string>());
```
varunsrin, line 81: edges implies that each item in the set is an edge in the tree/graph, but each item is really a set of children of a node in the tree. two questions: 

1. is this data structure preferable to storing a list of edges?
2. if so, what should we call it?

src/sets/signerSet/signerSet.test.ts
drshrey: I see `revoke` as the user action and `remove` as the internal action. For the client, they are **revoking** delegate access. The tactical nature of how that's done inside signer set is through a set of removal operations.

Using higher-order terms like `revoke` implicitly provides usage/ application context so hence why revoke is mentioned here

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+    // either as a delegate or custody signer
+    if (
+      this.revoked.has(proposedParentPubkey) ||
+      (!this.adds.has(proposedParentPubkey) && !this.custodySigners.has(proposedParentPubkey))
+    ) {
+      return err(`unable to use ${proposedParentPubkey}`);
+    }
+
+    // Concurrent Edge Case handling for when the delegate exists under a different parent
+    // See the "Conflicting Parents" section in https://farcasterxyz.notion.site/Signer-Set-85706ad04aea4bb6a3ac4e30f07bcd09
```
varunsrin, line 111: this spec is temporary, i wouldn't link to it as a permanent resource.

if some of the content is important, we should bring it into here or the protocol repo.

src/sets/signerSet/signerSet.test.ts
drshrey: The failure case of the parent not being a custody signer or another delegate is handled with the [same logic here](https://github.com/farcasterxyz/hub/pull/36/files#diff-1b97142d99baef4457c1ee2d7a58cf927584d76586bd9156fd21382c06328ba1R101-R108), so I don't think we need a separate test.

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+    if (existingSignerAdd !== undefined && existingSignerAdd.envelope.parentSignerPubkey !== proposedParentPubkey) {
+      // check if current edge has a higher lexicographical hash than the proposed edge
+      const existingParentPubkey = existingSignerAdd.envelope.parentSignerPubkey;
+      // compute hash of existing parentPubkey + childPubkey edge
+      const existingEdgeHash = blake2BHash(`${existingParentPubkey}${delegate}`);
```
varunsrin, line 116: why not compare the hashes of the SignerAdd object, which are already calculated and present?

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+      const proposedEdgeHash = blake2BHash(`${proposedParentPubkey}${delegate}`);
+
+      if (existingEdgeHash < proposedEdgeHash) {
```
varunsrin, line 119: we have a function called `hashCompare` in utils.ts which should be used for this, and which is used in other sets. it uses a specific string comparison algorithm that is defined to work across different languages in the same manner.

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+        // add delegate edge to proposed parent
+        const added = this._addDelegateEdge(proposedParentPubkey, delegate);
+        if (!added) {
+          return err(`could not add delegate ${delegate} to parent ${proposedParentPubkey}`);
+        }
+
+        // replace existing add with new SignerAdd structure
+        this.adds.set(delegate, delegateAddition);
+        return ok(undefined);
+      } else {
```
varunsrin, line 141: can this conditional nesting be reducing by reordering to run the else clause first? so something like... 

```ts
if (existingEdgeHash < proposedEdgeHash){
    return err(`delegate ${delegate} already exists under ${existingParentPubkey}`);
}

const subtreeRemoved = this._removeSubtree(delegate);
if (!subtreeRemoved) {
  return err(`could not remove delegate subtree ${delegate}`);
}
...
...
return ok(undefined)

```

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    this.revoked.set(delegate, removeMsg);
+    return ok(undefined);
+  }
+
+  public _numSigners(): number {
+    return this.custodySigners.size;
+  }
+
+  private _addDelegateEdge(parentPubkey: string, delegate: string): boolean {
```
varunsrin, line 204: naming of params is inconsistent -- both values are pubkeys, yet one is labelled Pubkey and the other isn't. 

given the scope of this method, i would simply call it `_addEdge(parent: string, child: string)`

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+    const parentChildren = this.edges.get(parentPubkey);
+    if (parentChildren === undefined) {
+      console.error(`pubkey ${parentPubkey} exists in edges set, not in revoked but no corresponding edge`);
```
varunsrin, line 207: prefer returning Result types -- instead of logging strings and returning false. they make it very easy to debug. the method calling this can simply check and pass the result back up, allowing this error message to trickle up the chain for debugging

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+      return false;
+    }
+
+    parentChildren.add(delegate);
+    this.edges.set(parentPubkey, parentChildren);
+    const newParentChildren = this.edges.get(parentPubkey);
+    if (newParentChildren !== undefined && !newParentChildren.has(delegate)) {
```
varunsrin, line 214: what would cause this to happen? isn't newParentChildren always isomorphic to parentChildren?

src/sets/signerSet/signerSet.test.ts
drshrey: Good point, I agree, I think treating `addCustody` as idempotent makes sense here.

From the context of the caller, `addCustody` should only ever be called from the trigger of a `transfer` or `register` on-chain event (relating to the transfer or registration of a Farcaster account). In the case of the attempt to add an existing custody signer, this would be a `transfer` event e.g. custody signer A registers the account, transfer to B and B transfers back to A. 

There is no practical consequence of future actions like `addDelegate`, `removeDelegate` from making `addCustody` idempotent. However, the benefit of treating the addition as idempotent is that there is one less failure case the client has to deal with (reduces caller complexity). 

One argument for returning a failure with the appropriate error is if the caller is making the assumption a **new** custody signer has been added and is performing logic or storing state with that basis. Having the failure could help the caller distinguish that, but I don't have enough context on the caller to know whether that's a real concern.

drshrey: cc @varunsrin if you have caller context/ any other input that may help getting to a conclusion here

varunsrin: making it idempotent seems reasonable

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+    /* new code */
+    const delegate = removeMsg.message.body.childKey;
+    const parent = removeMsg.envelope.parentSignerPubkey;
+
+    // check that delegate nor parent are in revoked set
+    if (this.revoked.has(delegate) || this.revoked.has(parent)) {
+      return err(`delegate ${delegate} or ${parent} has already been revoked in this account`);
+    }
+
+    // check that delegate exists and maps to proposed parent
```
pfletcherhill, line 169: I think this is a problem with the spec, but will the sets potentially diverge if adds and removes are received in different orders? If `b` is a valid signer in the tree, and a hub receives `add(b,c)` and then `rem(b,c)`, its state will be unchanged after processing the two messages in order. If another hub with the same state receives `rem(b,c)` and then later `add(b,c)`, it will end up thinking `c` is a valid signer, because it will throw away the first `rem(b,c)`.

src/sets/signerSet/signerSet.ts
drshrey: We could; how is that hash computed? Are we guaranteed for those hashes to always be distinct to the message e.g. a function of the parent and child pubkeys values of the SignerAdd object?

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+      this.revoked.has(proposedParentPubkey) ||
+      (!this.adds.has(proposedParentPubkey) && !this.custodySigners.has(proposedParentPubkey))
+    ) {
+      return err(`unable to use ${proposedParentPubkey}`);
+    }
+
+    // Concurrent Edge Case handling for when the delegate exists under a different parent
+    // See the "Conflicting Parents" section in https://farcasterxyz.notion.site/Signer-Set-85706ad04aea4bb6a3ac4e30f07bcd09
+    if (existingSignerAdd !== undefined && existingSignerAdd.envelope.parentSignerPubkey !== proposedParentPubkey) {
+      // check if current edge has a higher lexicographical hash than the proposed edge
```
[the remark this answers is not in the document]
pfletcherhill: Oh sorry, didn't see Varun commented the same thing below

src/sets/signerSet/signerSet.test.ts
drshrey: We import secp like `const secp = require('ethereum-cryptography/secp256k1');`. This is signaling my limits with typescript. I try to instead write it as a conventional import statement i.e. `import secp from 'ethereum-cryptography/secp256k1' and get errors like 

```
TypeError: Cannot read properties of undefined (reading 'getPublicKey')Jest
```

which corresponds to 
```
const signerRootPubkey = secp.getPublicKey(custodySigner);
```

Have you seen this issue in the past?

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+  private _removeSubtree(pubkey: string): boolean {
+    if (!this.edges.has(pubkey)) {
+      return false;
+    }
+
+    return this.edges.delete(pubkey);
+  }
+
+  private _nodeWithPubkeyExists(pubkey: string): boolean {
+    return (this.adds.has(pubkey) || this.custodySigners.has(pubkey)) && !this.revoked.has(pubkey);
```
pfletcherhill, line 253: I think the `!this.revoked.has(pubkey)` check is redundant

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    return true;
+  }
+
+  private _removeSubtree(pubkey: string): boolean {
+    if (!this.edges.has(pubkey)) {
+      return false;
+    }
+
+    return this.edges.delete(pubkey);
```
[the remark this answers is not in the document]
drshrey: Ah yes, great point, taking care of that now

src/sets/signerSet/signerSet.ts
varunsrin: yes, see other comment about using `hashCompare`

src/sets/signerSet/signerSet.ts
drshrey: Yep good call; my thinking here is that maintaining it as a Map with an associated block number is better. Reason being, we do `custodySigners` lookups in `addDelegate` and in `addCustody`, where `O(1)` time complexity is useful. 

The only time we will need it in an ordered format is during custody signer revocation which seems like it will be used a lot less often than the above two methods. In that case, the extra complexity in revocation logic using a map vs. array is:

1. an additional operation of retrieving the list of keys of the custody signer map (`O(number of items in set)` time complexity) 
2. check if the custody signer's block number is less than the specified upper limit block number (latest custody signer?)

I may be wrong here, but my assumption is the common case of custody signer revocation would effectively revoke all custody signers before the latest one. In which case, when we go through each key in the map to check if the block number is less than the latest/ specified block number, almost all of them should yield true regardless.

src/sets/signerSet/signerSet.ts
```diff
@@ -0,0 +1,352 @@
...
+          this.removed.delete(child);
+          this.adds.add(child);
+          this._moveChildToNewParent(existingEdge.parentPubkey, child, edge.hash);
+          // move all descendants in subtree including child to add set from removed
+          this._unremoveDelegateSubtree(child);
+        } else if (hashVal > 0) {
+          return err(`existing edge with parent ${existingEdge.parentPubkey} preserves delegate`);
+        }
+        return ok(undefined);
+      }
```
drshrey, line 143: logic in both of these cases feel similar enough that they could be DRYed out

pfletcherhill:
Closing because all commits are included in https://github.com/farcasterxyz/hub/pull/38

PR #38, opened 2022-07-21
[feature] Add SignerSet implementation

pfletcherhill:
Updated version of SignerSet implementation. Original work done by @drshrey. The SignerSet spec has changed a lot since this work was initially done, so here's a summary of what's included.

### What's included (high-level):
* SignerSet that stores custody addresses and lets those custody addresses add and remove delegate signers
* Scheme for revoking signers and dropping messages across all sets that were signed by the revoked signer
* Replace roots and signer changes with the SignerSet in the engine and across the codebase
* Remove simulator (it will be re-added in another repo in phase 3 of our hub launch plan)
* Update readme

### About SignerSet implementation
While this implementation is much simpler than prior SignerSet specs, there are still some quirks worth mentioning.
* Custody addresses are represented in `custodyAdds` and `custodyRemoves` and moved between the sets based on whether they are added or removed
* Custody addresses can only be added via Register and Transfer events from the Farcaster ID Registry (FIR), each of which has a block number, log index, and transaction hash
* When a transfer happens in the FIR, the old custody address is not immediately removed from the SignerSet, because that would immediately drop all messages signed by delegates of that address
* Custody addresses can only be removed by merging a `CustodyRemoveAll` message that is signed by a custody address. When accepted, the `CustodyRemoveAll` message removes all custody addresses that were added before the signer of the `CustodyRemoveAll` was added (based on the FIR events).
* Custody addresses function like a LWW set, where the order of FIR events is used to resolve conflicts. This setup means that custody addresses can be re-added after being removed, which is consistent with behavior in the FIR.
* Delegate signers can only be added via SignerAdd messages signed by custody addresses, and they can only be removed via SignerRemove messages signed by custody addresses.
* Unlike past designs, there is no message ordering, so a SignerRemove message can be accepted before a SignerAdd that references the same signer. This feature means the consistency guarantees (i.e. flexibility) of the SignerSet are equivalent to our other CRDTs.
* The signer sets also operate like a LWW set, using the message signer's FIR events to resolve conflicts.
* Like custody addresses, delegate signers can be re-added after they've been removed.

### About signer revocation
When a signer (custody addresses or delegates) is removed, all messages that were signed by that signer should be dropped, so the state matches what it would have been if the signer had been removed before the messages were accepted. Here are details about how we do signer revocation in this PR:
* The SignerSet now extends EventEmitter, so other parts of the codebase (i.e. the engine) can subscribe to SignerSet events. Eventually, all our CRDTs should be event emitters so devs can stream data changes from the CRDTs.
* The SignerSet has `addCustody`, `removeCustody`, `addSigner`, and `removeSigner` events
* Internally, right before a custody address is removed, the SignerSet drops all messages signed by that custody address
* The engine subscribes to events from the SignerSet, and when a signer is removed it calls a new `revokeSigner` method in all the CRDTs except the SignerSet
* Each `revokeSigner` takes the signer that was removed and drops all messages from its state that were signed by that signer

### SignerAdd and SignerRemove types
When signers were organized in a tree, it made sense to use "parent" and "child" terminology. But now that we only have custody addresses and delegate signers, I've updated the SignerAdd and SignerRemove bodies to be consistent:
```
type SignerAddBody = {
  delegate: string;
  edgeHash: string;
  delegateSignature: string;
  delegateSignatureType: SignatureAlgorithm.Ed25519;
  schema: 'farcaster.xyz/schemas/v1/signer-add';
};

type SignerRemoveBody = {
  delegate: string;
  schema: 'farcaster.xyz/schemas/v1/signer-remove';
};
```
The SignerEdge type also changes due to this naming:
```
type SignerEdge = {
  custody: string;
  delegate: string;
};
```


src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+      if (hashCompare(existingEdgeRemovesHash, hash) < 0) {
+        this._edgeRemoves.set(edgeKey, hash);
+      }
+      return ok(undefined);
+    }
+
+    // If parent exists in vAdds or custody
+    if (this._vertexAdds.has(parentPubKey) || this._custodySigners.has(parentPubKey)) {
+      // If child does not exist in vertices
+      if (!this._vertices.has(childPubKey)) {
```
varunsrin, line 187: should we also check that the child doesn't exist in custodySigners?

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+  constructor() {
+    this._custodySigners = new Set();
+    this._vertexAdds = new Set();
+    this._vertexRemoves = new Set();
+    this._edgeAdds = new Map();
+    this._edgeRemoves = new Map();
+    this._messages = new Map();
+  }
+
+  private get _vertices() {
```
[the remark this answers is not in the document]
pfletcherhill: Fair enough. I was undecided about where to put private getters, but I can move them down.

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
+import { Result, ok, err } from 'neverthrow';
+import { SignerAdd, SignerMessage, SignerRemove } from '~/types';
+import { isSignerAdd, isSignerRemove } from '~/types/typeguards';
+import { hashCompare } from '~/utils';
+
+/*   
+  SignerSet manages the account's associated custody addresses authorized 
+  for Signing and their corresponding delegates.
```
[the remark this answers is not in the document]
pfletcherhill: Added summary at top of set

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    if (isSignerAdd(message)) {
+      return this.mergeSignerAdd(message);
+    }
+
+    return err('SignerSet.merge: invalid message format');
+  }
+
+  mergeSignerAdd(message: SignerAdd): Result<void, string> {
```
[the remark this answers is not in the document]
pfletcherhill: Made these private in favor of just using `merge`

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+      const { childPubKey } = this.getPubKeysFromEdgeKey(edgeKey);
+      const existingSet = byChildMap.get(childPubKey) || new Set();
+      existingSet.add(edgeKey);
+      byChildMap.set(childPubKey, existingSet);
+    }
+    return byChildMap;
+  }
+
+  private getEdgesByParent(edges: Map<string, string>) {
+    const byChildMap = new Map<string, Set<string>>();
```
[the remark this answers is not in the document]
pfletcherhill: Ah sloppy me

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+    for (const edgeKey of edges.keys()) {
+      const { parentPubKey } = this.getPubKeysFromEdgeKey(edgeKey);
+      const existingSet = byChildMap.get(parentPubKey) || new Set();
+      existingSet.add(edgeKey);
+      byChildMap.set(parentPubKey, existingSet);
+    }
+    return byChildMap;
+  }
+
+  private add(parentPubKey: string, childPubKey: string, hash: string): Result<void, string> {
```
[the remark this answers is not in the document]
pfletcherhill: I'd prefer to keep it `add`, because I have a `removeEdge` method below that takes an edge key (tuple string) as argument. And the `add` and `remove` methods add and remove both vertices and edges, not just edges.

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+
+    this._custodySigners.add(custodySignerPubkey);
+    return ok(undefined);
+  }
+
+  /**
+   * Private Methods
+   */
+
+  private getEdgesByChild(edges: Map<string, string>) {
```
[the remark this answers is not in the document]
pfletcherhill: Good call, adding comment

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+        // Add child to vAdds and (a,b) to eAdds
+        this._vertexAdds.add(childPubKey);
+        this._edgeAdds.set(edgeKey, hash);
+
+        return ok(undefined);
+      }
+
+      // If child exists in vAdds
+      else if (this._vertexAdds.has(childPubKey)) {
+        // Get all parents of parentPubKey by traversing edgeAdds
```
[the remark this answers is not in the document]
pfletcherhill: Added comment

src/sets/signerSet.ts
pfletcherhill: Oh this brings up an interesting edge case...what happens when the same key is added as a custody signer on one hub and as a delegate on another hub? We don't currently handle that conflict.

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  private get _vertices() {
+    return new Set([...this._vertexAdds, ...this._vertexRemoves]);
+  }
+
+  private get _edges() {
+    return new Map([...this._edgeAdds, ...this._edgeRemoves]);
+  }
+
+  private get _edgeAddsByChild() {
```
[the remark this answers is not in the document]
pfletcherhill: Hmm `_edgeAddsByChild` makes more sense to me, since it's the result of `getEdgesByChild`. And `AllChildren` implies all children in both adds and removes to me.

src/sets/signerSet.ts
pfletcherhill: Okay, upon thinking about this more...
* Custody signers will be identified using addresses (20 bytes) rather than public keys, right? Because we'll look at the `msg.sender` in the account registry contract.
* Delegate signers now are identified using their public keys (32 bytes)
* If we keep this distinction—custody signers are addresses, delegate signers are public keys—we won't have conflicts between the two

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+
+    if (isSignerAdd(message)) {
+      return this.mergeSignerAdd(message);
+    }
+
+    return err('SignerSet.merge: invalid message format');
+  }
+
+  mergeSignerAdd(message: SignerAdd): Result<void, string> {
+    const parentPubKey = message.signer;
```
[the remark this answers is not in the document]
pfletcherhill: Based on our discussion before, this will be the custody's address. So I may rename this to `parentKey` to be slightly less confusing.

src/sets/signerSet.ts
pfletcherhill: This PR addresses this https://github.com/farcasterxyz/hub/pull/50

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    test("fails when child doesn't belong to parent", async () => {
+      expect(set.merge(addA).isOk()).toBe(true);
+      expect(set.merge(addCToA).isOk()).toBe(true);
+      const remC = await Factories.SignerRemove.create(
+        { data: { body: { childKey: addCToA.data.body.childKey } } },
+        { transient: { signer: custodySigner } }
+      );
+      expect(set.merge(remC).isOk()).toBe(false);
+      expect(vRems()).toEqual(new Set());
```
varunsrin, line 267: do we expect the edge to be retained here?

src/factories.ts
```diff
@@ -12,29 +12,55 @@ import {
...
   VerificationRemoveFactoryTransientParams,
   VerificationAddFactoryTransientParams,
+  SignerAdd,
+  SignerRemove,
+  SignerEdge,
+  SignerAddFactoryTransientParams,
+  SignatureAlgorithm,
+  Message,
+  MessageFactoryTransientParams,
+  MessageSigner,
```
varunsrin, line 22: nit: message signer isn't used here

src/utils.ts
```diff
@@ -106,3 +106,14 @@ export const generateEd25519KeyPair = async (): Promise<KeyPair> => {
...
   return '0x' + ed.utils.bytesToHex(text);
 };
+
+export const generateEd25519Signer = async (): Promise<Ed25519Signer> => {
+  const { privateKey, publicKey } = await generateEd25519KeyPair();
+  const signerKey = await convertToHex(publicKey);
+  return { privateKey, signerKey, type: SignatureAlgorithm.Ed25519 };
+};
+
+export const generateEthereumSigner = async (): Promise<EthereumSigner> => {
```
varunsrin, line 116: nit: might be worth adding a comment about this being an ECDSA signature that follows EIP-191

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      expect(set.merge(remCFromA).isOk()).toBe(true);
+      expect(vRems()).toEqual(new Set([remCFromA.data.body.childKey]));
+      expect(eRems().size).toEqual(1);
+    });
+
+    test('succeeds with duplicate signer remove message', () => {
+      expect(set.merge(addA).isOk()).toBe(true);
+      expect(set.merge(remA).isOk()).toBe(true);
+      expect(set.merge(remA).isOk()).toBe(true);
+    });
```
varunsrin, line 282: another case: `a child is added by 2 different parents and then deleted by 2 different parents`

addA, addBChildOfA, addCChildofA, addDChildOfB, addDChildOfC, remDChildofB, remDChildOfC

varunsrin: another case: `a remove that is added within a sub-tree after the sub-tree has been removed` 

addA, addBChildOfA, addCChildofB, remBChildOfA, addDChildofC

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    test('fails when parent does not exist', () => {
+      expect(vAdds().size).toEqual(0);
+      expect(set.merge(addCToA).isOk()).toBe(false);
+      expect(vAdds().size).toEqual(0);
+      expect(eAdds().size).toEqual(0);
+      expect(messages().keys()).not.toContain(addCToA.hash);
+    });
+
+    test('succeeds when child has already been deleted by another parent and moves new edge to edgeRemoves', () => {
```
varunsrin, line 178: test case `a child is added into a subtree that was removed`

addA, addBChildOfA, addCChildofB, remA

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      expect(set.merge(addA).isOk()).toBe(true);
+      expect(set.merge(addCToA).isOk()).toBe(true);
+      expect(set.merge(addB).isOk()).toBe(true);
+      expect(set.merge(remCFromA).isOk()).toBe(true);
```
varunsrin, line 182: should we also test that changing the order converges to the same state? or is that implicitly covered elsewhere?

addA -> remCFromA -> addCToA -> addB

src/sets/signerSet.test.ts
varunsrin: another case: `a child is removed after its parent has already been removed` 

addA, addBChildOfA, addCChildofB, remBChildOfA, remCChildOfB

varunsrin (changes requested):
One concern I have is that the test suite isn't exhaustive because there are so many little edge cases - I'll try to do another pass on this before we merge

src/sets/signerSet.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  getDelegates(): string[] {
+    return Array.from(this._vertexAdds);
+  }
+
+  getAllDelegates(): string[] {
+    return Array.from(this._vertices);
+  }
+
+  getEdgeKey(parentPubKey: string, childPubKey: string): string {
```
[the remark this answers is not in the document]
pfletcherhill: Let's default to using lowercased versions—it's easier to convert to lowercase rather than convert to checksum case. I just made that change [here](https://github.com/farcasterxyz/hub/pull/50/commits/99eb20f4f13149a221977f1ef4a44166b17d3ed1) for verifying messages and will make the edge keys lowercased in the signer set as well.

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+  test('succeeds with new custody signer', () => {
+    expect(set.addCustody(custody1).isOk()).toBe(true);
+    expect(custodySigners()).toContain(custody1);
+    expect(vAdds().size).toEqual(0);
+  });
+
+  test('succeeds with multiple new custody signers', () => {
+    expect(set.addCustody(custody1).isOk()).toBe(true);
+    expect(set.addCustody(custody2).isOk()).toBe(true);
+    expect(custodySigners()).toContain(custody1);
```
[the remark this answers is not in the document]
pfletcherhill: Sure, changing to `toEqual(...)`

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+        beforeAll(async () => {
+          addCToADuplicate = await Factories.SignerAdd.create({}, { transient: { signer: a, childSigner: c } });
+        });
+
+        test('succeeds with a lower message hash but does not update edgeAdds', () => {
+          expect(set.merge(addA).isOk()).toBe(true);
+          expect(set.merge(addCToA).isOk()).toBe(true);
+          const addCToADuplicateFail: SignerAdd = { ...addCToADuplicate, hash: addCToA.hash.slice(0, -1) };
+          expect(set.merge(addCToADuplicateFail).isOk()).toBe(true);
```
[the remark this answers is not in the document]
pfletcherhill: Yeah, good catch

src/sets/signerSet.test.ts
pfletcherhill: The edge from `remC`? No, because it's not a valid edge (yet) and the remove message doesn't contain enough info to verify whether it is valid or invalid.

src/sets/signerSet.test.ts
pfletcherhill: addA -> remCFromA -> addCToA -> addB will not match addA -> addCToA -> addB -> remCFromA, because c has to exist before it can be removed. That failure is covered in another test (`"fails when child hasn't been added yet"`)

src/sets/signerSet.test.ts
pfletcherhill: Added test

src/sets/signerSet.test.ts
pfletcherhill: Added tests for all of these

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+        });
+
+        test('succeeds with a higher message hash', () => {
+          expect(set.merge(addA).isOk()).toBe(true);
+          expect(set.merge(addB).isOk()).toBe(true);
+          expect(set.merge(addCToA).isOk()).toBe(true);
+          expect(vAdds().size).toEqual(3);
+          const addCToBHigherHash: SignerAdd = { ...addCToB, hash: addCToA.hash + 'a' };
+          expect(set.merge(addCToBHigherHash).isOk()).toBe(true);
+          expect(vAdds().size).toEqual(3);
```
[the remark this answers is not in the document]
pfletcherhill: Updated tests to check for specific values of vAdds, vRems, eAdds, eRems, and messages for each test

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    describe('when child already exists', () => {
+      describe('with different parent', () => {
+        test('succeeds with a lower message hash but moves new edge to edgeRemoves', () => {
+          expect(set.merge(addA).isOk()).toBe(true);
+          expect(set.merge(addB).isOk()).toBe(true);
+          expect(set.merge(addCToA).isOk()).toBe(true);
+          expect(vAdds().size).toEqual(3);
+          const addCToBLowerHash: SignerAdd = { ...addCToB, hash: addCToA.hash.slice(0, -1) };
+          expect(set.merge(addCToBLowerHash).isOk()).toBe(true);
+          expect(vAdds().size).toEqual(3);
```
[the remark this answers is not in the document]
pfletcherhill: Updated

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    expect(custodySigners()).toContain(custody2);
+    expect(vAdds().size).toEqual(0);
+  });
+
+  test('succeeds with a duplicate custody signer (idempotent)', () => {
+    expect(set.addCustody(custody1).isOk()).toBe(true);
+    expect(set.addCustody(custody1).isOk()).toBe(true);
+    expect(custodySigners()).toEqual(new Set([custody1]));
+  });
+});
```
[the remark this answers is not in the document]
pfletcherhill: Yeah, added

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    test("fails when child hasn't been added yet", () => {
+      expect(vAdds()).toEqual(new Set());
+      expect(set.merge(remA).isOk()).toBe(false);
+      expect(vAdds()).toEqual(new Set());
+      expect(vRems()).toEqual(new Set());
+      expect(messages().keys()).not.toContain(remA.hash);
+    });
+
+    test('succeeds and removes subtree', () => {
```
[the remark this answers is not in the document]
pfletcherhill: Updated with additional descendants and a peer B that is not removed

pfletcherhill:
Summary of changes since last review:
* Expand signer set test suite to check full state (i.e. `vAdds`, `vRems`, `eAdds`, `eRems`, and `messages`) after every test
* Add tests for all the cases mentioned in @varunsrin's review
* Add explanatory comments to signer set file and methods
* Update cycle prevention logic to only reject adds that would beat out other edges and cause a cycle in eAdds
* Add more cycle tests
* Support removing custody addresses
* Sanitize keys coming from all public methods, so we always use lowercase versions in signer set

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,639 @@
...
+afterAll(() => {
+  set.removeAllListeners();
+});
+
+beforeEach(() => {
+  events = [];
+  set._reset();
+});
+
+describe('addCustody', () => {
```
varunsrin, line 90: a case that is currently uncovered by tests: 

- block 1: custody A is added
- block 2: custody B is added
- block 3: custody B removes custody A with a RemoveAll
- block 4: custody A is added again  

this tests line 122 in signerSet `this._custodyRemoves.delete(sanitizedAddress);`

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,639 @@
...
+    expect(custodyRems()).toEqual(new Map());
+    expect(signerAdds()).toEqual(new Map());
+    expect(signerRems()).toEqual(new Map());
+    expect(events).toEqual([
+      ['addCustody', custody1.signerKey],
+      ['addCustody', custody2.signerKey],
+    ]);
+  });
+
+  test('succeeds (no-ops) with a duplicate custody address', () => {
```
varunsrin, line 137: nit: maybe 'succeeds idempotently with a duplicate custody address' ?

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+  /**
+   * addCustody adds a custody address to custodyAdds set. The method follows this high-level logic:
+   * - If the new address has never been seen before, add it
+   * - If the new address is already in custodyAdds, keep the entry with a higher block number
+   * - If the new address is in custodyRemoves, find the block number of the address that removed it
+   *   and move the address to custodyAdds if the new entry has a higher block number
+   *
+   * @param event - event from Farcaster ID Registry
+   */
+  addCustody(event: CustodyAddEvent): Result<void, string> {
```
varunsrin, line 101: it is confusing that `addCustody` and `merge` are semantically equivalent operations with different names

why merge in one place vs add<Item> in the other?

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+
+    // If custody exists in custodyAdds
+    const existingCustodyAdd = this._custodyAdds.get(sanitizedAddress);
+    if (existingCustodyAdd) {
+      // If existing block number wins (greater or equal), no-op
+      if (existingCustodyAdd.blockNumber >= event.blockNumber) return ok(undefined);
+    }
+
+    // If custody exists in custodyRemoves
+    const existingCustodyRemove = this._custodyRemoves.get(sanitizedAddress);
```
varunsrin, line 112: nit: should we use `Rem` and `Rems` instead of `Remove` and `Removes` everywhere? brevity might make the code easier to read

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+   *   and move the address to custodyAdds if the new entry has a higher block number
+   *
+   * @param event - event from Farcaster ID Registry
+   */
+  addCustody(event: CustodyAddEvent): Result<void, string> {
+    const sanitizedAddress = sanitizeSigner(event.custodyAddress);
+
+    // If custody exists in custodyAdds
+    const existingCustodyAdd = this._custodyAdds.get(sanitizedAddress);
+    if (existingCustodyAdd) {
```
varunsrin, line 106: nit: it may be slightly clearer if we combined these clauses into a single statement and expanded on the comment. something like: 

```ts
   // If the incoming add operation is an earlier block or the same block as the prior add operation, ignore it
    if (priorCustodyAdd && priorCustodyAdd.blockNumber >= event.blockNumber) {
       return ok(undefined);
    }

```

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+      if (existingCustodyAdd.blockNumber >= event.blockNumber) return ok(undefined);
+    }
+
+    // If custody exists in custodyRemoves
+    const existingCustodyRemove = this._custodyRemoves.get(sanitizedAddress);
+    if (existingCustodyRemove) {
+      const existingCustodyAddEvent = this._custodyAdds.get(sanitizeSigner(existingCustodyRemove.signer));
+
+      if (existingCustodyAddEvent) {
+        // If existing block number wins (same or greater block), no-op
```
varunsrin, line 117: more details here would be helpful to readers

```suggestion
        // If the incoming add operation adds an address that was previously removed
        // and if the address that performed the remove operation is more recent than the incoming add operation
        // do nothing
```

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+      // If existing block number wins (greater or equal), no-op
+      if (existingCustodyAdd.blockNumber >= event.blockNumber) return ok(undefined);
+    }
+
+    // If custody exists in custodyRemoves
+    const existingCustodyRemove = this._custodyRemoves.get(sanitizedAddress);
+    if (existingCustodyRemove) {
+      const existingCustodyAddEvent = this._custodyAdds.get(sanitizeSigner(existingCustodyRemove.signer));
+
+      if (existingCustodyAddEvent) {
```
varunsrin, line 116: i assume it is impossible to be in the state `existingCustodyRemove && ! existingCustodyAddEvent == true?` because if the addevent was removed then the remove event should also be removed?

if so, should we be throwing? because it seems like we will just ignore this case right now

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+        // If existing block number wins (same or greater block), no-op
+        if (existingCustodyAddEvent.blockNumber >= event.blockNumber) return ok(undefined);
+
+        // If new block number wins, remove address from removes
```
varunsrin, line 120: expanding on this comment in the prior style might also be helpful

varunsrin (commented):
completed review on `addCustody` method

varunsrin:
@pfletcherhill the coverage report here also shows a few untested lines in other functions that are probably worth adding coverage to: https://app.codecov.io/gh/farcasterxyz/hub/compare/38/tree/src/sets/signerSet.ts

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+    this.emit('addCustody', sanitizedAddress);
+    return ok(undefined);
+  }
+
+  /**
+   * Private Methods
+   */
+
+  /**
+   * mergeCustodyRemoveAll moves all address in custodyAdds with a block number before the
```
varunsrin, line 137: nit: `addresses`

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+   * signer of the CustodyRemoveAll message to custodyRemoves.
+   *
+   * Custody addresses that were added in the same block as the signer of the CustodyRemoveAll
+   * message will not be removed.
+   *
+   * @param message - CustodyRemoveAll message
```
varunsrin, line 143: nit: we should either remove this or specify more detail like `signed message that indicates that all prior custody addresses should be removed from the valid set`

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+   */
+  private mergeCustodyRemoveAll(message: CustodyRemoveAll): Result<void, string> {
+    const sanitizedAddress = sanitizeSigner(message.signer);
```
varunsrin, line 146: nit: should we just call this `lowerCaseAddress` instead? clearer about what it does

varunsrin: similar nit about just calling `message.signer.toLowerCase()` instead of `sanitizeSigner(message.signer)` which is just as verbose but less clear about what its doing

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+
+    // If custodyAddress exists in custodyRemoves, no-op
```
varunsrin, line 148: the comment is so similar to the logic that it doesn't really add much here. 

maybe something like `If the address that requested the remove all been removed by another another remove all message, do nothing`

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+    if (this._custodyRemoves.has(sanitizedAddress)) return ok(undefined);
+
+    // If custodyAddress does not exist in custodyAdds, fail
```
varunsrin, line 151: `if the add message for this custody address was never received, fail the operation`

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+    const custodyAddEvent = this._custodyAdds.get(sanitizedAddress);
+    if (!custodyAddEvent) return err('SignerSet.mergeCustodyRemoveAll: custodyAddress does not exist');
+
+    // For each custody address, remove if block number is before signer
```
varunsrin, line 155: `Remove each custody address and their child delegate signers where the custody address if the custody address was from a block that preceeds the block in which this signer became a custody address`

varunsrin: add an explicit note about how we don't care about events in the same block

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+    for (const [address, addEvent] of this._custodyAdds) {
+      if (addEvent.blockNumber < custodyAddEvent.blockNumber) {
+        // Before we remove the custody address, we should revoke the signer by dropping all messages signed
+        // by that custody address from the set
+
+        // Drop all SignerAdd messages signed by the custody address being removed
```
varunsrin, line 161: add a note about why its a drop and not a move into the remove set

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,639 @@
...
+        expect(res.isOk()).toBe(false);
+        expect(res._unsafeUnwrapErr()).toBe('SignerSet.mergeCustodyRemoveAll: custodyAddress does not exist');
+        expect(custodyAdds()).toEqual(new Map([[custody1.signerKey, addCustody1]]));
+        expect(custodyRems()).toEqual(new Map());
+        expect(signerAdds()).toEqual(new Map());
+        expect(signerRems()).toEqual(new Map());
+        expect(events).toEqual([['addCustody', custody1.signerKey]]);
+      });
+
+      test('succeeds and removes signers added by removed custody address', () => {
```
varunsrin, line 251: nit: `succeeds and drops signers`

varunsrin, line 251: nit: `succeeds and drops signers`

src/sets/signerSet.test.ts
```diff
@@ -0,0 +1,639 @@
...
+        expect(signerRems()).toEqual(new Map());
+        expect(events).toEqual([
+          ['addCustody', custody1.signerKey],
+          ['addSigner', a.signerKey],
+          ['addCustody', custody2.signerKey],
+          ['removeSigner', a.signerKey],
+          ['removeCustody', custody1.signerKey],
+        ]);
+      });
+
```
varunsrin, line 267: test case:  add signers, remove signers and then remove the custody address and test that the removes were dropped as well

varunsrin, line 267: test case:  add custody address A, do a rem all from A, then add B, do a rem all from B and assert that the rem all from A was dropped

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+   * - If the new signer is in signerRemoves, find the block number of the address that removed it and move the address
+   *   to signerAdds if the new entry was signed by an address with a higher block number
+   *
+   * @param message - a SignerAdd message
+   */
+  private mergeSignerAdd(message: SignerAdd): Result<void, string> {
+    const custodyAddress = sanitizeSigner(message.signer);
+    const signerKey = sanitizeSigner(message.data.body.delegate);
+
+    // If custody address has been removed, no-op
```
varunsrin, line 206: `If the custody address that authorized the signer has been removed by another operation, do nothing`

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+    if (this._custodyRemoves.has(custodyAddress)) return ok(undefined);
+
+    // If custody address is missing, fail
```
varunsrin, line 209: `If the custody address that authorized the signer was never added, do nothing`

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+      const existingCustodyAddEvent = this._custodyAdds.get(sanitizeSigner(existingSignerAdd.signer));
+
+      if (existingCustodyAddEvent) {
+        // If existing block number wins, no-op
+        if (existingCustodyAddEvent.blockNumber > custodyAddEvent.blockNumber) return ok(undefined);
+
+        // If block numbers are the same
+        if (existingCustodyAddEvent.blockNumber === custodyAddEvent.blockNumber) {
+          // If the custody addresses are the same and existing signer add has the same or higher hash, no-op
+          if (
```
varunsrin, line 225: can we do this more deterministcally by requiring a transactionId and logIndex to be passed in along with the block? 

that way we can determine exactly which custodyAdd event wes later and choose accordingly

src/sets/signerSet.test.ts
pfletcherhill: Adding

src/sets/signerSet.ts
pfletcherhill: Updated

src/sets/signerSet.ts
pfletcherhill: Eh I'm on the fence about this one. I went with "remove" because the remove types (`CastRemove` `VerificationRemove`, etc) will always have it spelled out, and I thought the sets should be consistent. But I could go either way...

src/sets/signerSet.ts
pfletcherhill: fwiw I think "existing" is better than "prior" because when we define that variable we don't know whether it is prior or not, just that it exists or doesn't

src/sets/signerSet.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  lookupCustody(custodyAddress: string) {
+    return this._custodyAdds.get(custodyAddress);
+  }
+
+  lookupSigner(signer: string) {
+    return this._signerAdds.get(signer);
+  }
+
+  merge(message: SignerMessage | CustodyRemoveAll): Result<void, string> {
```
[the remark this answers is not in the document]
pfletcherhill: See comment here https://github.com/farcasterxyz/hub/pull/38/#discussion_r944584009

src/sets/signerSet.ts
pfletcherhill: Added a comment from the last PR about this: https://github.com/pfletcherhill/farcaster-hub/pull/3#discussion_r942905953

If we decide to change from lowercase to uppercase for some reason, that's trivial to do if we used `sanitizeSigner` everywhere.

src/sets/signerSet.test.ts
pfletcherhill: Ah this is a bug! Good catch! We should update the past remove messages to reference a later block number rather than dropping them.

src/sets/signerSet.ts
varunsrin: ah ok fair, thats a good argument for the abstraction

src/sets/signerSet.test.ts
pfletcherhill: Eh I don't know. I always misspell idempotent when I'm typing it out. You definitely think it's better?

src/sets/signerSet.ts
pfletcherhill: Yeah, should be impossible. I added "unexpected state" errors here and a few other places we have a similar pattern.

varunsrin (approved):

PR #50, opened 2022-07-29
Add signatureType to messages

pfletcherhill:
Implements https://github.com/farcasterxyz/hub/issues/49

Summary of changes:
* Add `SignatureAlgorithm` enum type
* Add `MessageSigner` type that encompasses either an Ed25519 signer (with private key) or Ethereum signer (with wallet)
* Use `MessageSigner` instead of `privateKey` in factories and whenever we generated a key pair in order to support multiple types of signers
* Add `generateEd25519Signer` and `generateEthereumSigner` utils
* Clean up factories to share envelope stamping/signing code
* Update `validateMessage` method in the engine to handle multiple kinds of signatures

### DIFF
--- a/src/engine/engine.cast.test.ts
+++ b/src/engine/engine.cast.test.ts
@@ -1,32 +1,32 @@
 import Engine from '~/engine';
 import { Factories } from '~/factories';
-import { Cast, Reaction, Root } from '~/types';
+import { Cast, MessageFactoryTransientParams, MessageSigner, Reaction, Root } from '~/types';
 import Faker from 'faker';
-import { generateEd25519KeyPair, convertToHex } from '~/utils';
-import { hexToBytes } from 'ethereum-cryptography/utils';
+import { generateEd25519Signer, generateEthereumSigner } from '~/utils';
 
 const engine = new Engine();
 const username = 'alice';
 
 describe('mergeCast', () => {
-  let alicePrivateKey: string;
+  // let alicePrivateKey: string;
+  let aliceSigner: MessageSigner;
   let aliceAddress: string;
   let root: Root;
   let cast: Cast;
   let reaction: Reaction;
-  let transient: { transient: { privateKey: Uint8Array } };
+  let transient: { transient: MessageFactoryTransientParams };
   const subject = () => engine._getCastAdds(username);
 
   beforeAll(async () => {
-    const keyPair = await generateEd25519KeyPair();
+    // Randomly generate either an Ed25519 or Ethereum signer
+    if (Math.random() > 0.5) {
+      aliceSigner = await generateEd25519Signer();
+    } else {
+      aliceSigner = await generateEthereumSigner();
+    }
+    aliceAddress = aliceSigner.signerKey;
 
-    const privateKeyBuffer = keyPair.privateKey;
-    alicePrivateKey = await convertToHex(privateKeyBuffer);
-
-    const addressBuffer = keyPair.publicKey;
-    aliceAddress = await convertToHex(addressBuffer);
-
-    transient = { transient: { privateKey: hexToBytes(alicePrivateKey) } };
+    transient = { transient: { signer: aliceSigner } };
 
     root = await Factories.Root.create({ data: { rootBlock: 100, username: 'alice' } }, transient);
 
--- a/src/engine/engine.cast.test.ts
+++ b/src/engine/engine.cast.test.ts
@@ -2,7 +2,7 @@ import Engine from '~/engine';
 import { Factories } from '~/factories';
 import { Cast, MessageFactoryTransientParams, MessageSigner, Reaction, Root } from '~/types';
 import Faker from 'faker';
-import { generateEd25519Signer, generateEthereumSigner } from '~/utils';
+import { generateEd25519Signer } from '~/utils';
 
 const engine = new Engine();
 const username = 'alice';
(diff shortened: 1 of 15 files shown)


src/engine/engine.cast.test.ts
```diff
@@ -1,32 +1,32 @@
...
-    const keyPair = await generateEd25519KeyPair();
-
-    const privateKeyBuffer = keyPair.privateKey;
-    alicePrivateKey = await convertToHex(privateKeyBuffer);
-
-    const addressBuffer = keyPair.publicKey;
-    aliceAddress = await convertToHex(addressBuffer);
-
-    transient = { transient: { privateKey: hexToBytes(alicePrivateKey) } };
+    // Randomly generate either an Ed25519 or Ethereum signer
```
pfletcherhill, line 21: @varunsrin because it takes longer to generate and use an Ethereum signer, adding randomness like this means our test suite timing changes +/- 5 seconds based on how many test suites are using Ethereum vs Ed25519. Do you think that's okay for now, or should I default to Ed25519 and add an extra test to each engine suite that uses an Ethereum signer?

src/engine/engine.cast.test.ts
```diff
@@ -1,32 +1,32 @@
...
-import { generateEd25519KeyPair, convertToHex } from '~/utils';
-import { hexToBytes } from 'ethereum-cryptography/utils';
+import { generateEd25519Signer, generateEthereumSigner } from '~/utils';
 
 const engine = new Engine();
 const username = 'alice';
 
 describe('mergeCast', () => {
-  let alicePrivateKey: string;
+  // let alicePrivateKey: string;
```
varunsrin, line 11: nit: can we just eliminate this?

src/engine/engine.cast.test.ts
varunsrin, line 21: why random? all casts must be signed by Ed25519 according to our specification 

the only message types that can be signed by Ethereum signers are the SignerAdds

src/engine/engine.reaction.test.ts
```diff
@@ -1,29 +1,30 @@
...
   const subject = () => engine._getActiveReactions(username);
 
   beforeAll(async () => {
-    const keyPair = await generateEd25519KeyPair();
-    const privateKeyBuffer = keyPair.privateKey;
-    alicePrivateKey = await convertToHex(privateKeyBuffer);
-    const addressBuffer = keyPair.publicKey;
-    aliceAddress = await convertToHex(addressBuffer);
-    transient = { transient: { privateKey: hexToBytes(alicePrivateKey) } };
+    // Randomly generate either an Ed25519 or Ethereum signer
```
varunsrin, line 20: same comment about types here

varunsrin (changes requested):

src/engine/engine.cast.test.ts
pfletcherhill: Oh I didn't realize that. I thought a custody address could sign messages as well.

src/engine/engine.cast.test.ts
pfletcherhill: Oops

src/engine/engine.cast.test.ts
pfletcherhill: No longer relevant

pfletcherhill: Resolved based on slack conversation. No longer generates Ethereum signers.

src/engine/index.ts
```diff
@@ -356,16 +357,27 @@ class Engine {
...
-      return err('validateMessage: invalid signature');
+    // 3. Check that the signatureType and signature are valid.
+    if (message.signatureType === SignatureAlgorithm.EthereumPersonalSign) {
+      try {
+        const recoveredSigner = ethers.utils.verifyMessage(message.hash, message.signature);
+        if (recoveredSigner.toLowerCase() !== message.signer.toLowerCase()) {
+          return err('validateMessage: invalid signature');
+        }
+      } catch (e: any) {
+        return err('validateMessage: invalid signature');
```
varunsrin, line 368: nit: should we try to stringify the error and include it here so it makes debugging easier?

src/factories.ts
```diff
@@ -12,29 +12,48 @@ import {
...
-import { convertToHex, hashMessage, signEd25519, hashFCObject } from '~/utils';
-import * as ed from '@noble/ed25519';
+import { hashMessage, signEd25519, hashFCObject, generateEd25519Signer, generateEthereumSigner } from '~/utils';
+
+/**
+ * addEnvelopeToMessage adds hash, signer, signature, and signatureType to a message
+ * object using the signer in transientParams if one is present
+ */
+const addEnvelopeToMessage = async (
+  props: Message,
```
varunsrin, line 27: nit: why props instead of message? especially since the comment says message object too

varunsrin (commented):
Good to merge, left some small nits that can be fixed in this PR or a followup

varunsrin:
Good to merge, left some small nits that can be fixed in this PR or a followup

src/factories.ts
pfletcherhill: I had kept it consistent with the factories that have props and transientParams, but I don't feel strongly either way. I'll change to `message`.

src/engine/index.ts
pfletcherhill: Eventually, yeah. There's a todo in verifications to pass through more helpful error messages from `verifyMessage` too. I'm going to punt on doing that now though.

PR #53, opened 2022-08-02
Use enums for hash algorithm and use them in all messages

zachterrell57:


src/client.ts
```diff
@@ -1,5 +1,6 @@
+import { keccak256 } from 'ethers/lib/utils';
```
pfletcherhill, line 1: Can you remove these unused imports?

src/types/index.ts
```diff
@@ -236,6 +237,12 @@ export type KeyPair = {
   publicKey: Uint8Array;
 };
 
+/** HashingAlgorithm enum */
```
pfletcherhill, line 240: `HashAlgorithm`

src/types/index.ts
```diff
@@ -236,6 +237,12 @@ export type KeyPair = {
...
+export enum HashAlgorithm {
+  Blake2b = 'blake2b',
+  Keccak256 = 'keccak256',
```
pfletcherhill, line 243: We don't support Keccak256 anymore, so this enum should just be `Blake2b = 'blake2b'`. Then in the future if we add more hashing options, they can go here.

src/utils.ts
```diff
@@ -15,13 +15,16 @@ export const hashMessage = async (item: Message): Promise<string> => {
  * The object is canonicalized before hashing, and all properties that start with an underscore are removed,
  * after which the string is passed to blake2b.
  */
-export const hashFCObject = async (object: Record<string, any>): Promise<string> => {
+export const hashFCObject = async (object: Record<string, any>, algorithm: HashAlgorithm): Promise<string> => {
```
pfletcherhill, line 18: I'd make this an optional parameter (i.e. `..., algorithm: HashAlgorithm = HashAlgorithm.Blake2b` so we don't have to add it every time we call `hashFCObject`.

src/utils.ts
```diff
@@ -15,13 +15,16 @@ export const hashMessage = async (item: Message): Promise<string> => {
...
   // Remove any keys that start with _ before hashing, as these are intended to be unhashed.
   const objectCopy = JSON.parse(JSON.stringify(object));
   removeProps(objectCopy);
 
   // Canonicalize the object according to JCS: https://datatracker.ietf.org/doc/html/rfc8785
   const canonicalizedObject = canonicalize(objectCopy) || '';
+  if (algorithm === HashAlgorithm.Keccak256) {
```
pfletcherhill, line 25: No need to support keccak256 anymore

pfletcherhill (changes requested):
I think you can make this a simpler change. You don't have to support keccak256, and you can make the hash algorithm an optional parameter in the utils functions to avoid specifying `HashAlgorithm.Blake2b` everywhere.

src/types/index.ts
```diff
@@ -4,13 +4,14 @@ import { ethers } from 'ethers';
  * Message is a generic type that represents any cryptographically signed Message on Farcaster
  *
  * @data - the data that is being signed.
- * @hash - the keccak256 hash of the message.
+ * @hash - the blake2b hash of the message.
```
pfletcherhill, line 7: Can you update this comment to include hashType and signatureType? (seems like I forgot to add that 🤦)

src/utils.test.ts
```diff
@@ -2,7 +2,7 @@ import { ethers } from 'ethers';
 import Faker from 'faker';
 import * as ed from '@noble/ed25519';
 import { hashFCObject, hashCompare, generateEthereumSigner, generateEd25519Signer, convertToHex } from '~/utils';
-import { Ed25519Signer, EthereumSigner } from '~/types';
+import { Ed25519Signer, EthereumSigner, HashAlgorithm } from '~/types';
```
pfletcherhill, line 5: Can you run `yarn lint` locally to pick up these linting warnings and fix them?

src/utils.ts
```diff
@@ -15,7 +15,10 @@ export const hashMessage = async (item: Message): Promise<string> => {
  * The object is canonicalized before hashing, and all properties that start with an underscore are removed,
  * after which the string is passed to blake2b.
  */
-export const hashFCObject = async (object: Record<string, any>): Promise<string> => {
+export const hashFCObject = async (
+  object: Record<string, any>,
+  algorithm: HashAlgorithm = HashAlgorithm.Blake2b
```
pfletcherhill, line 20: This is an annoying linting warning, because we should be including this param but aren't using it because there's only one option. How about wrapping the hashing part of this method in a `if (algorithm === HashAlgorithm.Blake2b) {...}` and then erroring out if algorithm is anything else?

pfletcherhill (changes requested):

src/utils.ts
pfletcherhill: Alternatively, we can just remove the `algorithm` param entirely until we have others to check

pfletcherhill (approved):

PR #54, opened 2022-08-12
Faster Verification Set Tests

zachterrell57:
Decreased test time from ~8.7 seconds to ~4.9 seconds

### DIFF
--- a/src/engine/engine.verification.test.ts
+++ b/src/engine/engine.verification.test.ts
@@ -1,6 +1,13 @@
 import Engine from '~/engine';
 import { Factories } from '~/factories';
-import { MessageSigner, Root, Verification, VerificationAddFactoryTransientParams, VerificationClaim } from '~/types';
+import {
+  MessageSigner,
+  Root,
+  Verification,
+  VerificationAdd,
+  VerificationAddFactoryTransientParams,
+  VerificationClaim,
+} from '~/types';
 import Faker from 'faker';
 import { ethers } from 'ethers';
 import { hashFCObject, generateEd25519Signer } from '~/utils';
@@ -17,6 +24,7 @@ describe('mergeVerification', () => {
   let aliceClaimHash: string;
   let aliceExternalSignature: string;
   let transientParams: { transient: VerificationAddFactoryTransientParams };
+  let genericVerificationAddMessage: VerificationAdd;
 
   // Generate key pair for alice and root message
   beforeAll(async () => {
@@ -26,15 +34,28 @@ describe('mergeVerification', () => {
     transientParams = { transient: { signer: aliceSigner, ethWallet: aliceEthWallet } };
     aliceRoot = await Factories.Root.create({ data: { rootBlock: 100, username: 'alice' } }, transientParams);
 
-    // Generate a claim hash for alice
     const verificationClaim: VerificationClaim = {
       username: 'alice',
       externalUri: aliceEthWallet.address,
     };
     aliceClaimHash = await hashFCObject(verificationClaim);
 
-    // Generate an external signature for alice
     aliceExternalSignature = await aliceEthWallet.signMessage(aliceClaimHash);
+
+    genericVerificationAddMessage = await Factories.VerificationAdd.create(
+      {
+        data: {
+          rootBlock: aliceRoot.data.rootBlock,
+          username: 'alice',
+          signedAt: aliceRoot.data.signedAt + 1,
+          body: {
+            claimHash: aliceClaimHash,
+            externalSignature: aliceExternalSignature,
+          },
+        },
+      },
+      transientParams
+    );
   });
 
   // Every test should start with a valid signer and root for alice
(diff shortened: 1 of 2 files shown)


src/engine/engine.verification.test.ts
```diff
@@ -230,6 +273,10 @@ describe('mergeVerification', () => {
           rootBlock: aliceRoot.data.rootBlock,
           username: 'alice',
           signedAt: aliceRoot.data.signedAt + 1,
+          body: {
```
pfletcherhill, line 276: Small thing, but in situations like this where we generate a totally vanilla verification add, could we instead generate one in `beforeAll` and then re-use it here?

pfletcherhill: The test right above this could use the same message, for instance, no reason to generate it twice.

src/engine/engine.verification.test.ts
```diff
@@ -13,14 +13,28 @@ describe('mergeVerification', () => {
...
   // Generate key pair for alice and root message
   beforeAll(async () => {
     aliceSigner = await generateEd25519Signer();
+    aliceEthWallet = ethers.Wallet.createRandom();
     aliceAddress = aliceSigner.signerKey;
-    transientParams = { transient: { signer: aliceSigner } };
+    transientParams = { transient: { signer: aliceSigner, ethWallet: aliceEthWallet } };
     aliceRoot = await Factories.Root.create({ data: { rootBlock: 100, username: 'alice' } }, transientParams);
+
+    // Generate a claim hash for alice
```
pfletcherhill, line 29: Nit: I don't think the comments here add much, the code is pretty self-explanatory

src/sets/verificationSet.test.ts
```diff
@@ -1,12 +1,43 @@
...
+  let aliceClaimHash: string;
+  let aliceExternalSignature: string;
+  let transientParams: { transient: VerificationAddFactoryTransientParams };
+
+  beforeAll(async () => {
+    aliceSigner = await generateEd25519Signer();
+    aliceEthWallet = ethers.Wallet.createRandom();
+    transientParams = { transient: { signer: aliceSigner, ethWallet: aliceEthWallet } };
+
+    // Generate a claim hash for alice
```
pfletcherhill, line 30: Drop the comments (like above)

src/sets/verificationSet.test.ts
```diff
@@ -1,12 +1,43 @@
...
 const set = new VerificationSet();
 const adds = () => set._getAdds();
 const removes = () => set._getRemoves();
 
 describe('merge', () => {
+  let aliceSigner: MessageSigner;
+  let aliceEthWallet: ethers.Wallet;
+  let aliceClaimHash: string;
+  let aliceExternalSignature: string;
+  let transientParams: { transient: VerificationAddFactoryTransientParams };
```
pfletcherhill, line 23: I think you can generate three verification adds and two verification removes at the top of this test suite and re-use them all throughout the tests. Testing the set is easier than testing the engine, because the set doesn't validate the full message format, so instead of having to re-generate a full message you can just clone it and update one of its attributes.

See the upcoming signer set test as an example: https://github.com/farcasterxyz/hub/pull/38/files#diff-921b6d5993256316ba07bf4ee1911ca31e2ca2b705bbac2d17078b12999f2911

pfletcherhill: Try to replace every instance of `add1`, `add2`, etc in the suite below with the ones you generate. Ideally, we only have async code in `beforeAll` and all the tests are synchronous.

pfletcherhill (changes requested):
Added a few comments about cutting out even more factory calls

src/sets/verificationSet.test.ts
```diff
@@ -1,12 +1,79 @@
...
+} from '~/types';
+import { ethers } from 'ethers';
+import { generateEd25519Signer, hashFCObject } from '~/utils';
 
 const set = new VerificationSet();
 const adds = () => set._getAdds();
 const removes = () => set._getRemoves();
 
 describe('merge', () => {
+  let aliceSigner: MessageSigner;
```
pfletcherhill, line 19: You can just call this `signer`—we only care about usernames in the engine—and make it an `Ed25519Signer` rather than `MessageSigner`. `MessageSigner` can be either `Ed25519Signer` or `EthereumSigner` but all verification messages are signed by Eddsa signers.

src/sets/verificationSet.test.ts
```diff
@@ -1,12 +1,79 @@
...
+  let aliceEthWallet: ethers.Wallet;
```
pfletcherhill, line 20: `ethWallet`, no need to include "alice." And remove "alice" from the other variables too.

src/sets/verificationSet.test.ts
```diff
@@ -1,12 +1,79 @@
...
+  beforeAll(async () => {
+    aliceSigner = await generateEd25519Signer();
+    aliceEthWallet = ethers.Wallet.createRandom();
+    transientParams = { transient: { signer: aliceSigner, ethWallet: aliceEthWallet } };
+
+    const verificationClaim: VerificationClaim = {
+      username: 'alice',
+      externalUri: aliceEthWallet.address,
+    };
+    aliceClaimHash = await hashFCObject(verificationClaim);
```
pfletcherhill, line 41: I know it's faster, but as I'm thinking about it more I'm a bit wary of using the same claim hash for all test messages. The verification set indexes its sets by claim hash so messages with the same claim hash should collide. How much slower is it if you organize it the way you have it but have the factory generate the claim hash for add1, add2, and add3?

pfletcherhill (changes requested):
Nice to see overall deletions! Suggested some changes below.

pfletcherhill (approved):
Lgtm

PR #99, opened 2022-09-09
feat: Implement a JSON RPC server and client

sagar-a16z:
## Motivation

https://github.com/farcasterxyz/hub/issues/98 Need a mechanism to communicate between Hubs and to be able to serve data to clients.

## Change Summary

Created a JSON RPC server and client. 
Added the ability to transfer `CastSet`s, `FID`s, and `SignerSet`s.

## Merge Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,8 @@
   },
   "dependencies": {
     "@noble/ed25519": "^1.6.1",
+    "@types/body-parser": "^1.19.2",
+    "body-parser": "^1.20.0",
     "caip": "^1.1.0",
     "canonicalize": "^1.0.8",
     "cli-table3": "^0.6.1",
@@ -56,6 +58,7 @@
     "ethereum-cryptography": "^1.1.2",
     "ethers": "^5.6.1",
     "faker": "5.5.3",
+    "jayson": "^4.0.0",
     "log-update": "5.0.1",
     "neverthrow": "^5.0.0",
     "tiny-typed-emitter": "^2.1.0",
--- a/package.json
+++ b/package.json
@@ -49,8 +49,6 @@
   },
   "dependencies": {
     "@noble/ed25519": "^1.6.1",
-    "@types/body-parser": "^1.19.2",
-    "body-parser": "^1.20.0",
     "caip": "^1.1.0",
     "canonicalize": "^1.0.8",
     "cli-table3": "^0.6.1",
--- a/src/engine/index.ts
+++ b/src/engine/index.ts
@@ -78,15 +78,8 @@ class Engine implements RPCHandler {
    */
 
   /** Get a Set of all the FIDs known */
-  async allFIDs(): Promise<Set<number>> {
-    const fids = new Set<number>();
-    const maps = [this._casts, this._reactions, this._verifications, this._signers, this._follows];
-    for (const map of maps) {
-      for (const key of map.keys()) {
-        fids.add(key);
-      }
-    }
-    return fids;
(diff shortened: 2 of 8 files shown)


src/engine/index.ts
```diff
@@ -164,6 +195,12 @@ class Engine {
     return signerSet.mergeIDRegistryEvent(event);
   }
 
+  async signersForFID(fid: number): Promise<SignerSet> {
```
pfletcherhill, line 198: `getAllSignerMessagesforFid`, return `Promise<Set<SignerMessage>>`

src/network/rpc.ts
```diff
@@ -0,0 +1,166 @@
...
+
+const serverOpts = {
+  version: VERSION,
+  replacer,
+  reviver,
+};
+
+export class RPCServer {
+  _jsonServer!: jayson.Server;
+  _handler!: RPCHandler;
```
pfletcherhill, line 54: What's the reason for having a separate `RPCHandler` interface representing the Engine instead of just typing it as `Engine` directly? I'd keep it simple and have the constructor take `engine: Engine` and store the same.

src/network/rpc.ts
```diff
@@ -0,0 +1,166 @@
...
+import * as jayson from 'jayson/promise';
+import CastSet from '~/sets/castSet';
+import SignerSet from '~/sets/signerSet';
+
+const VERSION = 0.1;
+
+export type Port = number;
+
+export enum RPCRequest {
+  allCasts = 'allCasts',
```
pfletcherhill, line 13: Nit: I think convention for enums is to capitalize the keys

src/network/rpc.ts
```diff
@@ -0,0 +1,166 @@
...
+};
+
+const serverOpts = {
+  version: VERSION,
+  replacer,
+  reviver,
+};
+
+export class RPCServer {
+  _jsonServer!: jayson.Server;
```
pfletcherhill, line 53: Add `private` if these are meant to be private attributes. I assume so based on their naming.

src/sets/castSet.ts
```diff
@@ -48,6 +48,28 @@ class CastSet {
     return ok(undefined);
   }
 
+  static replacer(key: any, value: any) {
```
pfletcherhill, line 51: Replace with `getAllMessages()` method

src/sets/signerSet.ts
```diff
@@ -137,6 +137,45 @@ class SignerSet extends TypedEmitter<SignerSetEvents> {
     return ok(undefined);
   }
 
+  static replacer(key: any, value: any) {
```
pfletcherhill, line 140: Replace with `getAllMessages()` method

pfletcherhill (changes requested):

src/network/rpc.ts
sagar-a16z: I think it's cleaner this way. 

I can think of a future where the RPCHandler is actually implemented by more than 1 object underneath and not just an Engine.

package.json
```diff
@@ -49,13 +49,16 @@
   },
   "dependencies": {
     "@noble/ed25519": "^1.6.1",
+    "@types/body-parser": "^1.19.2",
+    "body-parser": "^1.20.0",
```
pfletcherhill, line 53: Do you use body-parser anywhere?

sagar-a16z: Removed! Thanks!

pfletcherhill (approved):
Lgtm

PR #101, opened 2022-09-09
feat: Add RPC getters for all Sets

sagar-a16z:
## Change Summary

Added RPC APIs for the remaining Sets in the engine. 
Refactored API names a bit
## Merge Checklist

### DIFF
--- a/src/engine/index.ts
+++ b/src/engine/index.ts
@@ -78,7 +78,7 @@ class Engine implements RPCHandler {
    */
 
   /** Get a Set of all the FIDs known */
-  async getFids(): Promise<Set<number>> {
+  async getUsers(): Promise<Set<number>> {
     return new Set(Array.from(this._signers.keys()));
   }
 
@@ -126,7 +126,7 @@ class Engine implements RPCHandler {
   }
 
   /** Get the entire cast set for an fid */
-  async getAllCastsForFid(fid: number): Promise<Set<Cast>> {
+  async getAllCastsByUser(fid: number): Promise<Set<Cast>> {
     const casts = this._casts.get(fid);
     if (casts) {
       return casts.getAllMessages();
@@ -144,6 +144,15 @@ class Engine implements RPCHandler {
     return reactionSet ? reactionSet.get(targetURI) : undefined;
   }
 
+  /** Get te entire reaction set for an fid */
+  async getAllReactionsByUser(fid: number): Promise<Set<Reaction>> {
+    const reactions = this._reactions.get(fid);
+    if (reactions) {
+      return reactions.getAllMessages();
+    }
+    return new Set();
+  }
+
   /**
    * Follow Methods
    */
@@ -154,6 +163,14 @@ class Engine implements RPCHandler {
     return followSet ? followSet.get(targetURI) : undefined;
   }
 
+  async getAllFollowsByUser(fid: number): Promise<Set<Follow>> {
+    const follows = this._follows.get(fid);
+    if (follows) {
+      return follows.getAllMessages();
+    }
+    return new Set();
+  }
+
   /**
    * Verification methods
    */
@@ -162,6 +179,14 @@ class Engine implements RPCHandler {
   getVerification(fid: number, claimHash: string): Verification | undefined {
     const verificationSet = this._verifications.get(fid);
     return verificationSet ? verificationSet.get(claimHash) : undefined;
+  }
+
+  async getAllVerificationsByUser(fid: number): Promise<Set<Verification>> {
+    const verifications = this._verifications.get(fid);
+    if (verifications) {
+      return verifications.getAllMessages();
+    }
+    return new Set();
   }
 
   /**
@@ -188,7 +213,8 @@ class Engine implements RPCHandler {
     return signerSet.mergeIDRegistryEvent(event);
   }
 
-  async getAllSignerMessagesForFid(fid: number): Promise<Set<SignerMessage>> {
+  /** Get the entire set of signers for an Fid */
+  async getAllSignerMessagesByUser(fid: number): Promise<Set<SignerMessage>> {
     const signerSet = this._signers.get(fid);
     if (signerSet) {
       return signerSet.getAllMessages();
--- a/src/network/rpc.test.ts
+++ b/src/network/rpc.test.ts
@@ -1,10 +1,25 @@
 import { AddressInfo } from 'net';
-import { CastRemove, CastShort, EthereumSigner, IDRegistryEvent, MessageSigner, ReactionAdd, SignerAdd } from '~/types';
+import {
+  CastRemove,
+  CastShort,
+  EthereumSigner,
+  FollowAdd,
+  FollowRemove,
+  IDRegistryEvent,
+  MessageSigner,
+  ReactionAdd,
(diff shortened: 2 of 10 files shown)


src/engine/index.ts
```diff
@@ -78,7 +78,7 @@ class Engine implements RPCHandler {
    */
 
   /** Get a Set of all the FIDs known */
-  async getFids(): Promise<Set<number>> {
+  async getUsers(): Promise<Set<number>> {
```
pfletcherhill, line 81: This shouldn't be a promise, it returns data synchronously from the signers map:
```
getUsers(): Set<number> {
  return new Set([...this._signers.keys()]);
}
```

src/engine/index.ts
```diff
@@ -126,7 +126,7 @@ class Engine implements RPCHandler {
   }
 
   /** Get the entire cast set for an fid */
-  async getAllCastsForFid(fid: number): Promise<Set<Cast>> {
+  async getAllCastsByUser(fid: number): Promise<Set<Cast>> {
```
pfletcherhill, line 129: No async/promise. Looks like this comment applies to all the new methods.

src/network/rpc.test.ts
```diff
@@ -41,27 +70,72 @@ describe('rpc', () => {
...
+    castRemove = await Factories.CastRemove.create(createParams, createOptions);
+    reaction = await Factories.ReactionAdd.create(createParams, createOptions);
+    reactionRemove = await Factories.ReactionRemove.create(createParams, createOptions);
+    follow = await Factories.FollowAdd.create(createParams, createOptions);
+    followRemove = await Factories.FollowRemove.create(
+      { data: { fid: aliceFid, body: { targetUri: follow.data.body.targetUri + '1' } } },
+      createOptions
+    );
+
+    const verificationClaim: VerificationEthereumAddressClaim = {
```
pfletcherhill, line 104: Because you're including `aliceEthWallet` in verificationOptions above, you don't need to create `verificationClaim`, `aliceClaimHash`, or `aliceExternalSignature` here. The factory will do it for you. Those were manually set in the engine.verifications test so we could re-use those specifically as part of a slight performance improvement, but they're not relevant (and make the code more complicated) when you're generating a single verification.

src/network/rpc.test.ts
```diff
@@ -41,27 +70,72 @@ describe('rpc', () => {
...
+      fid: aliceFid,
+      externalUri: Factories.EthereumAddressURL.build(undefined, {
+        transient: { address: aliceEthWallet.address },
+      }).toString(),
+      blockHash: aliceBlockHash,
+    };
+    aliceClaimHash = await hashFCObject(verificationClaim);
+    aliceExternalSignature = await aliceEthWallet.signMessage(aliceClaimHash);
+
+    verification = await Factories.VerificationEthereumAddress.create(
```
pfletcherhill, line 114: If you want a verification add and remove that are unrelated, then only pass `createParams` and `createOptions` in like other message types. The way it's written now, `verificationRemove` will override `verification`, because they have the same external ethereum address and the remove happens later.

src/network/rpc.test.ts
```diff
@@ -74,17 +148,24 @@ describe('rpc', () => {
     await server.stop();
   });
 
+  test('get all signers for FID', async () => {
+    const response = await client.getAllSignerMessagesByUser(aliceFid);
```
pfletcherhill, line 152: This shouldn't have to be async

src/network/rpc.test.ts
```diff
@@ -95,10 +176,29 @@ describe('rpc', () => {
     expect(response._unsafeUnwrap()).toEqual(new Set([aliceFid]));
   });
 
-  test('get all signers for FID', async () => {
-    const response = await client.getAllSignerMessagesForFid(aliceFid);
+  test('get all reactions for Fid', async () => {
```
pfletcherhill, line 179: Nit: "...for an fid" rather than "...for Fid"

src/network/rpc.ts
```diff
@@ -2,23 +2,28 @@ import { AddressInfo } from 'net';
...
 
 const VERSION = 0.1;
 
 export type Port = number;
 
 export enum RPCRequest {
-  GetFids,
-  GetAllCastsForFid,
-  GetAllSignerMessagesForFid,
+  GetUsers,
```
pfletcherhill, line 12: If we don't include any values, the default is that they'll all convert to numbers (i.e. `GetUsers` is really 0). Is that intended?

src/network/rpc.ts
```diff
@@ -51,23 +56,40 @@ export class RPCServer {
     this._handler = rpcHandler;
     this._jsonServer = new jayson.Server(
       {
-        [RPCRequest[RPCRequest.GetFids]]: new jayson.Method({
+        [RPCRequest[RPCRequest.GetUsers]]: new jayson.Method({
```
pfletcherhill, line 59: I'm a bit confused by the syntax here. Wouldn't `{ [RPCRequest.GetUsers]: ... }` be the same?

src/network/rpc.ts
```diff
@@ -119,31 +141,53 @@ export class RPCClient {
...
+      // TODO use the family?
       // family: address.family
       replacer,
       reviver,
     });
   }
 
-  async getAllCastsForFid(fid: number): Promise<Result<CastSet, string>> {
-    const response = await this._tcpClient.request(RPCRequest[RPCRequest.GetAllCastsForFid], { fid: fid });
+  async getFids(): Promise<Result<Set<number>, string>> {
```
pfletcherhill, line 151: `getUsers`?

src/sets/followSet.ts
```diff
@@ -62,6 +62,10 @@ class FollowSet {
     return ok(undefined);
   }
 
+  getAllMessages(): Set<Follow> {
+    return new Set([Array.from(this._adds.values()), Array.from(this._removes.values())].flat());
```
pfletcherhill, line 66: Simpler syntax: `return new Set([...this._adds.values(), ...this._removes.values()])`

pfletcherhill (changes requested):

src/network/rpc.ts
sagar-a16z: `[RPCRequest.GetUsers]` evaluates to the enum value, where I want the enum key here.

src/engine/index.ts
sagar-a16z: I'm just planning for an async storage sometime down the line.

src/network/rpc.test.ts
sagar-a16z: Makes sense. Thank you. I was running into some errors with the Factory default so just copied what was in the Verification tests.

src/network/rpc.test.ts
sagar-a16z: This is a RPC call it needs to be async or if we use Jayson without promises it will be a callback.

src/network/rpc.ts
sagar-a16z: Yeah that's intended. It's only being used for its key strings. 

It shouldn'tbe exported though. Just used by the client and server classes internally.

pfletcherhill (approved):

PR #105, opened 2022-09-12
feat(utils): Generate a large number of mock events to populate an Engine

sagar-a16z:
## Motivation

Fixes https://github.com/farcasterxyz/hub/issues/103

## Change Summary

Added a mock event generator that creates fake users and generates events for them.
Also fixed a bug in the Factory for `FollowRemove`

## Merge Checklist


src/engine/engine.mock.test.ts
```diff
@@ -0,0 +1,58 @@
+import Engine from '~/engine';
+import { populateEngine, EventConfig, createUser, mockEvents, MockFCEvent } from '~/engine/engine.mock';
```
varunsrin, line 2: 'EventConfig' is defined but never used

src/engine/engine.mock.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  test('Generates a mock event of each type', async () => {
+    const fid = Faker.datatype.number();
+    const user = await createUser(engine, fid);
+    const users = await engine.getUsers();
+    expect(users.size).toEqual(1);
+
+    await mockEvents(engine, [user], 1, MockFCEvent.Cast);
+    const castadds = engine._getCastAdds(fid);
```
varunsrin, line 20: 'castadds' is assigned a value but never used

src/engine/engine.mock.ts
```diff
@@ -0,0 +1,138 @@
+import Engine from '~/engine';
+import { Factories } from '~/factories';
+import { Cast, EthereumSigner, MessageSigner } from '~/types';
```
varunsrin, line 3: 'Cast' is defined but never used

src/engine/engine.mock.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Message } from '~/types';
+
+const engine = new Engine();
+
+describe('PopulateEngine', () => {
+  beforeEach(async () => {
+    engine._reset();
+  });
+
+  test('Generates a mock event of each type', async () => {
```
varunsrin, line 13: nit: test descriptions should start with lowercase to be consistent

src/engine/engine.mock.ts
```diff
@@ -0,0 +1,138 @@
...
+  await mockEvents(engine, userInfos, config.Verifications, MockFCEvent.Verification);
+  // create casts and removes
+  await mockEvents(engine, userInfos, config.Casts, MockFCEvent.Cast);
+  // create follows
+  await mockEvents(engine, userInfos, config.Follows, MockFCEvent.Follow);
+  // create reactions
+  await mockEvents(engine, userInfos, config.Reactions, MockFCEvent.Reaction);
+};
+
+/** creates a random user and registers thir information with the engine */
```
varunsrin, line 55: nit: typo in comments `thir` and slightly unclear what 'register with engine' means. might say something lile 

`Simulate a new user being added to the engine by creating mock on-chain event and off-chain delegate message and merging both with the engine`

varunsrin: also might call the method something like `mockFid`?

src/engine/engine.mock.ts
```diff
@@ -0,0 +1,138 @@
...
+/** creates and merges a number of add/remove pair Events for the given event type */
+export const mockEvents = async (engine: Engine, userInfos: UserInfo[], count: number, event: MockFCEvent) => {
+  for (const userInfo of userInfos) {
+    const createParams = { data: { fid: userInfo.fid } };
+    const createOptions = { transient: { signer: userInfo.delegateSigner } };
+    let result;
+    switch (event) {
+      case MockFCEvent.Verification: {
+        await Promise.all(
+          [...Array(count)].map(async () => {
```
varunsrin, line 89: does the parallelization here improve run times? my intuition is that since this is really single-threaded at the CPU level and since the create call is probably CPU bound and not I/O bound, this may have no improvement or be slightly worse.

src/engine/engine.mock.ts
```diff
@@ -0,0 +1,138 @@
...
+          })
+        );
+        break;
+      }
+      case MockFCEvent.Cast: {
+        await Promise.all(
+          [...Array(count)].map(async () => {
+            result = await engine.mergeMessage(await Factories.CastShort.create(createParams, createOptions));
+            expect(result.isOk).toBeTruthy();
+            result = await engine.mergeMessage(await Factories.CastRemove.create(createParams, createOptions));
```
varunsrin, line 105: the removes aren't going to be "true" inversions of the adds since they don't have a common reference. so what we're doing here is generating an arbitrary N adds and arbitrary N removes that are for different items. 

that maybe OK for what you need, but calling it out in case it isnt.

src/engine/engine.mock.ts
```diff
@@ -0,0 +1,138 @@
+import Engine from '~/engine';
```
varunsrin, line 1: the naming of `engine.mock.test` and `engine.mock` might be a little confusing for readers. it implies that the former is the test suite and the latter is logic, based on conventions in our codebase. what i think you are doing is using `engine.mock` to create utils for testing and `engine.mock.test` to execute the tests.  

there may not be an obvious way to separate these two files withotu causing confusion. perhaps it is better to leave them in a mega-file until the logical separation becomes more obvious?

varunsrin (changes requested):

src/engine/engine.mock.ts
sagar-a16z: Makes sense, moving them both to engine.test

src/engine/engine.mock.ts
sagar-a16z: Perfect, will do!

src/engine/engine.mock.ts
sagar-a16z: It is intentional, since both types of messages are stored differently on each Set. 

I'll add a comment to make sure that's clear.

src/engine/engine.mock.ts
sagar-a16z: Yes it did improve when I did this. I think I can revisit the perf of this utility function later.

varunsrin (approved):

PR #106, opened 2022-09-13
feat(sync): Implement a simple Engine sync over RPC

sagar-a16z:
## Motivation

Fixes https://github.com/farcasterxyz/hub/issues/102

## Change Summary

- Wrote a simple test that copies data from one Engine (server) to another (client) over RPC.
- Added RPC APIs for querying IDRegistry events as well

## Merge Checklist

## Additional Context

Depends on https://github.com/farcasterxyz/hub/pull/105. 
I will rebase this once that's landed.

### DIFF
--- a/src/engine/index.ts
+++ b/src/engine/index.ts
@@ -235,6 +235,14 @@ class Engine implements RPCHandler {
     return new Set();
   }
 
+  async getCustodyEventByUser(fid: number): Promise<IDRegistryEvent | undefined> {
+    const signerSet = this._signers.get(fid);
+    if (signerSet) {
+      return signerSet.getCustodyAddressEvent();
+    }
+    return undefined;
+  }
+
   /**
    * Private Methods
    */
--- a/src/engine/index.ts
+++ b/src/engine/index.ts
@@ -237,10 +237,7 @@ class Engine implements RPCHandler {
 
   async getCustodyEventByUser(fid: number): Promise<IDRegistryEvent | undefined> {
     const signerSet = this._signers.get(fid);
-    if (signerSet) {
-      return signerSet.getCustodyAddressEvent();
-    }
-    return undefined;
+    return signerSet ? signerSet.getCustodyAddressEvent() : undefined;
   }
 
   /**
--- a/src/network/rpc.sync.test.ts
+++ b/src/network/rpc.sync.test.ts
@@ -51,7 +51,7 @@ describe('rpcSync', () => {
 
       for (const user of userIds) {
         // get the signer messages first so we can prepare to ingest the remaining messages later
-        let result: Result<any, string> = await client.getCustodyEventByuser(user);
+        let result: Result<any, string> = await client.getCustodyEventByUser(user);
         expect(result.isOk()).toBeTruthy();
         const custodyResult = await clientEngine.mergeIDRegistryEvent(result._unsafeUnwrap());
         expect(custodyResult.isOk()).toBeTruthy();
--- a/src/network/rpc.sync.test.ts
+++ b/src/network/rpc.sync.test.ts
@@ -1,0 +1,109 @@
+import { AddressInfo } from 'net';
+import { populateEngine } from '~/engine/engine.mock.test';
+import { Result } from 'neverthrow';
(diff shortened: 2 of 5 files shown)


src/network/rpc.ts
```diff
@@ -193,4 +201,15 @@ export class RPCClient {
     }
     return new Ok(response.result);
   }
+
+  async getCustodyEventByuser(fid: number): Promise<Result<IDRegistryEvent, string>> {
```
sds, line 205: Should this be capitalized as `...ByUser`?

sagar-a16z: yes it should! Thanks!

src/engine/index.ts
```diff
@@ -235,6 +235,14 @@ class Engine implements RPCHandler {
     return new Set();
   }
 
+  async getCustodyEventByUser(fid: number): Promise<IDRegistryEvent | undefined> {
+    const signerSet = this._signers.get(fid);
+    if (signerSet) {
```
varunsrin, line 240: teeny-tiny nit: can use ternary to make the code easier to read 

```suggestion
    return signerSet ? signerSet.getCustodyAddressEvent() : undefined;
```

src/network/rpc.sync.test.ts
```diff
@@ -0,0 +1,109 @@
+import { AddressInfo } from 'net';
```
varunsrin, line 1: i would call this file `rpcSync.test.ts` instead of `rpc.sync.test.ts` because the latter implies that sync is a namespace that we should treat differently (but I don't believe that's the case)

src/network/rpc.sync.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  test(
+    'sync data from one engine to another ',
+    async () => {
+      const clientEngine = new Engine();
+
+      // sanity test that the server has something in it
+      const users = await serverEngine.getUsers();
+      expect(users.size).toEqual(NUM_USERS);
+
+      //
```
varunsrin, line 39: nit: prefer `/** **/` for multiline comments

src/network/rpc.sync.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      // Merge all the messages
+      //
+
+      const result = await client.getUsers();
+      expect(result.isOk()).toBeTruthy();
+      const userIds = result._unsafeUnwrap();
+
+      for (const user of userIds) {
+        // get the signer messages first so we can prepare to ingest the remaining messages later
+        let result: Result<any, string> = await client.getCustodyEventByUser(user);
```
varunsrin, line 54: avoid use of any wherever possible. 

even in this case where it is trivial and won't improve anything meaningful, it just sets a bad example that might encourage use of any in other cases where it is important. 

image

src/network/rpc.sync.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        expect(result.isOk()).toBeTruthy();
+        const custodyResult = await clientEngine.mergeIDRegistryEvent(result._unsafeUnwrap());
+        expect(custodyResult.isOk()).toBeTruthy();
+        // next, get all the signer messages
```
varunsrin, line 58: nit: i would recommend being consistent with comment style. the general approaches i like is: 

1. Document the why, when something is not clear
2. Document the what/how, if you are using advanced features or doing fancy stuff
3. Otherwise don't document

I think this is a case of 3 where its pretty obvious what's happening

src/network/rpc.sync.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+        // now collect all the messages from each set
+        const responses = await Promise.all([
+          client.getAllCastsByUser(user),
+          client.getAllFollowsByUser(user),
+          client.getAllReactionsByUser(user),
+          client.getAllVerificationsByUser(user),
+        ]);
+
+        const messages = responses.flatMap((response) => {
```
varunsrin, line 74: the "what" might be useful to document for future readers since there is some fancier than usual stuff happening

varunsrin (approved):

PR #107, opened 2022-09-13
feat: rocksdb persistence and make CastSet a LWW CRDT

pfletcherhill:
## Motivation

Implements
* https://github.com/farcasterxyz/hub/issues/61
* https://github.com/farcasterxyz/hub/issues/112

## Change Summary

* Add rocksdb persistence layer (using [rocksdb](https://github.com/Level/rocksdb/) and [abstract-leveldown](https://github.com/Level/abstract-leveldown/) with custom types declaration files)
* Refactor sets to use rocksdb for state instead of memory
* Refactor engine to instantiate one set for each CRDT type (rather than one per type per fid)
* Refactor CastSet CRDT to be a LWW (i.e. CastRemove with timestamp _before_ associated cast is not valid)
* Isolate usage of neverthrow Result pattern to engine and network (i.e. sets/ and db/ files will throw errors)
* Add `FarcasterError` class to improve error handling

### Engine vs sets vs db
Previously, the engine stored an instance of each set (i.e. SignerSet or CastSet) for each fid and those sets stored messages in memory. Now that we're persisting messages to RocksDB, the engine stores a single instance of each set for all fids. This PR also splits sets into two classes—a set (i.e. CastSet) that resolves conflicts between messages and a db (i.e. CastDB) that fetches and mutates set state. I chose to separate these, because the db class includes the RocksDB schema for that set.

Each db class includes public methods for fetching and mutating data, and all mutations are RocksDB transactions. You can see the schema for each set by looking at the relevant db file—there is a comment at the top with the format of keys/values, or you can look at the `...Prefix` or `...Key` methods.

The db classes all extend a common `MessageDB` class that includes generic methods for fetching, putting, and deleting messages. This class stores serialized messages by their hash and it also indexes messages by their signer (to make revoking signers more efficient later on).

### Example: merge CastShort message
1. Call `mergeMessage` on Engine with message
2. Engine validates the format of the message. If it's valid, the engine calls `merge` on CastSet
3. CastSet checks for conflicts between the new message and existing ones. If the message should be merged, the set calls `putCastAdd` on CastDB.
4. CastDB creates and commits a new RocksDB transaction that (1) adds the message (`messages!<message hash>: CastShort`), (2) indexes the message by signer (`fid!<fid>!signer!<signer key>!<message type>: <message hash>`), (3) adds the message to castAdds (`fid!<fid>!castAdds!<message hash>: <message hash>`), (4) removes a conflicting CastRemove message if it exists (from `fid!<fid>!castRemoves!<message hash>`), and (5) indexes the message by the cast target (`castShortsByTarget!<cast target>!<message hash>: <message hash>`).

## Merge Checklist


sds (commented):
I understand this is a rough draft, so I don't have anything concrete about the code at this time.

I also realize that the team probably already discussed this, but I want to raise a minor concern about using LevelDB at this stage of our implementation. The three main issues are:

* Ability to deploy new versions of hub code without downtime.
* Potential performance limitations of LevelDB (I might be misreading their docs)
* 

### Ability to deploy new versions of hub code without downtime
All of our deployment pipelines assume a stateless service—that is, the service itself does not store data on-disk. It instead entrusts that data to a managed service (Redis, Postgresql, etc.) which is run by AWS.

If we go ahead with using LevelDB, we have to either:
* Implement a deployment model where we want for the new Fargate tasks have synced from existing Fargate tasks (and hope there are no bugs in that process).
* Implement a custom disk snapshotting/restoration mechanism (since ECS/Fargate doesn't support this directly)
* Run our hubs on long-lived EC2 instances and deploy using a more traditional/old-school approach.

I think it's worth chatting about whether a another persistent store (i.e. one that runs separately from the hub process itself, for now) is worth exploring.

### Potential performance limitations of LevelDB

While I have no doubt LevelDB has significant performance benefits, there was one limitation mentioned in its docs which concerned me:
>Only a single process (possibly multi-threaded) can access a particular database at a time.

— https://github.com/google/leveldb#limitations

If I am understanding this statement, given that we are using NodeJS for the foreseeable future, we are necessarily limiting the amount of parallelism we can have when serving requests from a hub (NodeJS recently added support for multi-threading via [worker threads](https://nodejs.org/api/worker_threads.html), but I'm not sure we want to go down that route just yet). Since hubs perform additional CPU-based computation when receiving requests, it's important that we allow hubs to be parallelizable. If we are limited to a single process per on-disk DB, that might not be ideal.

With that said, Geth and other major implementations use LevelDB, so I feel like I'm missing something crucial about this constraint which makes it not worth worrying about in practice.

---

Zooming out, I just wanted to understand if there were any other options we considered. Supporting zero-downtime node deployments is hard if we need to manage the on-disk data ourselves vs entrusting to another service. I know it's not conventional, but could we use a Postgres DB as a backend _for now_ so that we could use a well-established deployment model and have trustworthy backup/recovery processes?

rohitkulshreshtha:
A separate persistence store could also provide necessary building blocks to do https://github.com/farcasterxyz/hub/issues/71 without burdening the hubs.

PR #130, opened 2022-09-28
feat: Hubs

sagar-a16z:
## Motivation

We need a class definition for the `Hub` object. 
A Hub is made up of a collection of services (RPC, GossipSub, Engine...) that together implement all of its behaviors.

fixes https://github.com/farcasterxyz/hub/issues/64

## Change Summary

Created a Hub class.

- Pulled in PubSub peer discovery to run alongside gossip to keep the libp2p peerlist up to date
- Added new RPC API to Publish messages to Hubs (Hubs will then publish those to the rest of the network via Gossip)
- Added support for bootstrapping a Hub off of a list of other Hub addresses. https://github.com/farcasterxyz/hub/issues/64
- Wrote a test that bootstraps a Hub off of an existing Hub on a network. 

Note - we need a lot more tests here as we flesh out more features. Filed https://github.com/farcasterxyz/hub/issues/129

## Merge Checklist

### DIFF
--- a/src/hub.test.ts
+++ b/src/hub.test.ts
@@ -5,8 +5,10 @@ import { Factories } from '~/factories';
 import { Hub, HubOpts } from '~/hub';
 import { RPCClient } from '~/network/rpc';
 import { sleep } from '~/utils';
-import { Content, GossipMessage, NETWORK_TOPIC_PRIMARY } from '~/network/protocol';
+import { ContactInfoContent, Content, GossipMessage, NETWORK_TOPIC_PRIMARY } from '~/network/protocol';
 import { Message } from '~/types';
+import { ServerError } from '~/errors';
+import { jest } from '@jest/globals';
 
 const TEST_TIMEOUT = 2 * 60 * 1000;
 const opts: HubOpts = { simpleSync: false };
@@ -34,7 +36,7 @@ const tearDownHub = async (hub: Hub) => {
   await hub.destroyDB();
 };
 
-describe('Hub tests', () => {
+describe('Hub running tests', () => {
   beforeEach(async () => {
     hub = new Hub(opts);
     await hub.start();
@@ -150,5 +152,48 @@ describe('Hub tests', () => {
     };
     const result = await hub.handleGossipMessage(badMessage as any as GossipMessage);
     expect(result.isErr());
+  });
+
+  test('Fail to submit invalid messages', async () => {
+    // try to send it a message
+    const submitResult = hub.submitMessage(await Factories.CastShort.create());
+    expect((await submitResult).isErr()).toBeTruthy();
+  });
+});
+
+describe('Hub negative tests', () => {
+  test('Poke a hub before starting it', async () => {
+    hub = new Hub(opts);
+
+    expect(hub.rpcAddress).toBeUndefined();
+    expect(hub.gossipAddresses).toEqual([]);
+    expect(() => {
+      hub.identity;
+    }).toThrow(ServerError);
+  });
+
+  test('Fail to sync from invalid peers', async () => {
+    hub = new Hub({});
+    await hub.start();
+
+    const syncHandler = jest.fn((success: boolean) => {
+      expect(success).toBeFalsy();
+    });
+
+    hub.addListener('syncComplete', syncHandler);
+
+    const badPeerInfo: ContactInfoContent = {
+      peerId: '',
+    };
+    // fails because this peerinfo has no RPC to sync from
+    hub.simpleSyncFromPeer(badPeerInfo);
+    badPeerInfo.rpcAddress = { address: '', port: 0, family: 'ip4' };
+    // fails because hub has no peers (despite this peerInfo having an rpc address)
+    hub.simpleSyncFromPeer(badPeerInfo);
+
+    expect(syncHandler).toBeCalledTimes(2);
+
+    await hub.stop();
+    await hub.destroyDB();
   });
 });
--- a/src/hub.test.ts
+++ b/src/hub.test.ts
@@ -1,10 +1,12 @@
 import Faker from 'faker';
 import { AddressInfo } from 'net';
-import { mockFid, populateEngine } from '~/engine/mock';
+import { generateUserInfo, getIDRegistryEvent, getSignerAdd, mockFid, populateEngine } from '~/engine/mock';
 import { Factories } from '~/factories';
 import { Hub, HubOpts } from '~/hub';
 import { RPCClient } from '~/network/rpc';
 import { sleep } from '~/utils';
+import { Content, GossipMessage, NETWORK_TOPIC_PRIMARY } from '~/network/protocol';
+import { Message } from '~/types';
 
 const TEST_TIMEOUT = 2 * 60 * 1000;
 const opts: HubOpts = { simpleSync: false };
@@ -40,7 +42,7 @@ describe('Hub tests', () => {
   });
 
   afterEach(async () => {
-    tearDownHub(hub);
+    await tearDownHub(hub);
   });
 
   test('Run a Hub and send it a message', async () => {
@@ -80,7 +82,8 @@ describe('Hub tests', () => {
         await secondHub.start();
         // wait until sync completes
         await new Promise((resolve) => {
-          secondHub.addListener('sync_complete', () => {
+          secondHub.addListener('sync_complete', (success) => {
+            expect(success).toBeTruthy();
             resolve(undefined);
           });
         });
@@ -93,4 +96,59 @@ describe('Hub tests', () => {
     },
     TEST_TIMEOUT
   );
+
+  const testMessage = async (message: Message) => {
+    const gossipMessage: GossipMessage<Content> = {
+      content: {
+        message,
+        root: '',
+        count: 0,
+      },
+      topics: [NETWORK_TOPIC_PRIMARY],
+    };
+    const result = await hub.handleGossipMessage(gossipMessage);
+    expect(result.isOk()).toBeTruthy();
+  };
+
+  test('Hub handles various valid gossip messages', async () => {
+    const aliceFid = Faker.datatype.number();
+    const aliceInfo = await generateUserInfo(aliceFid);
+    const IDRegistryEvent: GossipMessage<Content> = {
+      content: {
+        message: await getIDRegistryEvent(aliceInfo),
+        root: '',
+        count: 0,
+      },
+      topics: [NETWORK_TOPIC_PRIMARY],
+    };
+    const idRegistryResult = await hub.handleGossipMessage(IDRegistryEvent);
+    expect(idRegistryResult.isOk());
+
+    await testMessage(await getSignerAdd(aliceInfo));
+    await testMessage(await getSignerAdd(aliceInfo));
+    await testMessage(
+      await Factories.CastShort.create({ data: { fid: aliceFid } }, { transient: { signer: aliceInfo.delegateSigner } })
+    );
+    await testMessage(
+      await Factories.FollowAdd.create({ data: { fid: aliceFid } }, { transient: { signer: aliceInfo.delegateSigner } })
+    );
+    await testMessage(
+      await Factories.ReactionAdd.create(
+        { data: { fid: aliceFid } },
+        { transient: { signer: aliceInfo.delegateSigner } }
+      )
+    );
+  });
+
+  test('Invalid messages fail', async () => {
+    const badMessage = {
+      content: {
+        not: '',
+        a: 0,
+        message: {},
+      },
+    };
+    const result = await hub.handleGossipMessage(badMessage as any as GossipMessage);
+    expect(result.isErr());
+  });
 });
--- a/src/hub.test.ts
+++ b/src/hub.test.ts
@@ -82,7 +82,7 @@ describe('Hub tests', () => {
         await secondHub.start();
         // wait until sync completes
         await new Promise((resolve) => {
-          secondHub.addListener('sync_complete', (success) => {
+          secondHub.addListener('syncComplete', (success) => {
             expect(success).toBeTruthy();
             resolve(undefined);
           });
--- a/src/hub.test.ts
+++ b/src/hub.test.ts
@@ -1,0 +1,96 @@
+import Faker from 'faker';
+import { AddressInfo } from 'net';
+import { mockFid, populateEngine } from '~/engine/mock';
+import { Factories } from '~/factories';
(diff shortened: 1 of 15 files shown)


src/hub.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        await new Promise((resolve) => {
+          secondHub.addListener('sync_complete', () => {
+            resolve(undefined);
+          });
+        });
+        await compareHubs(hub, secondHub);
+        await secondHub.stop();
+        shouldStop = false;
+      } finally {
+        if (shouldStop) await secondHub.stop();
```
pfletcherhill, line 86: Couldn't you use `catch` here instead of the flag + finally? Or am I missing something?

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+  private options: Partial<HubOpts>;
+  private gossipNode: Node;
+  private rpcServer: RPCServer;
+  private syncState: SimpleSyncState;
+  private contactTimer?: NodeJS.Timer;
+  private rocksDB: RocksDB;
+
+  engine: Engine;
+
+  constructor(options: Partial<HubOpts>) {
```
pfletcherhill, line 71: Nit: rather than `Partial<HubOpts>` here, I'd prefer to be explicit in the type about which members are optional

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+  sync_start: () => void;
+  /**
+   * Triggered when a simple sync completes
+   */
+  sync_complete: () => void;
+}
+
+export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
+  private options: Partial<HubOpts>;
+  private gossipNode: Node;
```
pfletcherhill, line 63: I think I have a pretty good understanding about the different purposes and functionality, but would you mind writing up a short doc at some point summarizing the difference between all these classes (Hub, Node, RPCServer, Engine, etc)? It's a good thing to have documented somewhere, and I just want to confirm we have the right separation of concerns.

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+    await this.rpcServer.stop();
+    if (!this.options.rocksDBName) {
+      console.warn(
+        this.identity,
+        'Destroying temporary database at',
+        this.rocksDB.location,
+        '\n Specify a name to persist DB'
+      );
+      await this.rocksDB.destroy();
+      this.rocksDB = undefined as any;
```
pfletcherhill, line 139: I don't love tricking typescript in this way. Could we make `rocksDB` an optional attribute or make use of the rocksdb `status` attribute instead?

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+
+  get gossipAddresses() {
+    return this.gossipNode.multiaddrs ?? [];
+  }
+
+  async stop() {
+    clearInterval(this.contactTimer);
+    await this.gossipNode.stop();
+    await this.rpcServer.stop();
+    if (!this.options.rocksDBName) {
```
pfletcherhill, line 131: Is there a production case for doing this? If not, and the only usecase we see is for testing, could we put this logic in a test instead?

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+      await this.rocksDB.destroy();
+      this.rocksDB = undefined as any;
+    } else {
+      await this.rocksDB.close();
+    }
+  }
+
+  // Returns the Gossip peerId string of this Hub
+  get identity(): string {
+    return this.gossipNode.peerId ? this.gossipNode.peerId.toString() + ':' : 'No Identity';
```
pfletcherhill, line 147: What would be the case where a hub would have no identity?

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  private registerEventHandlers() {
+    // subscribe to all interesting topics
+    this.gossipNode.gossip?.subscribe(NETWORK_TOPIC_PRIMARY);
+    this.gossipNode.gossip?.subscribe(NETWORK_TOPIC_CONTACT);
+
+    this.gossipNode.addListener('message', async (_topic, message) => {
+      console.debug(message);
+      await message.match(
```
pfletcherhill, line 162: Could we put this logic in another method so we can test it more easily?

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+        async (gossipMessage) => {
+          let result;
+          if (isUserContent(gossipMessage.content as UserContent)) {
+            const message = (gossipMessage.content as UserContent).message;
+            result = await this.engine.mergeMessage(message as Message);
+          } else if (isIDRegistryContent(gossipMessage.content as IDRegistryContent)) {
+            const message = (gossipMessage.content as IDRegistryContent).message;
+            result = await this.engine.mergeIDRegistryEvent(message as IDRegistryEvent);
+          } else if (isContactInfo(gossipMessage.content as ContactInfo)) {
+            // TODO Maybe we need a ContactInfo CRDT?
```
pfletcherhill, line 172: What would the ContactInfo CRDT look like? A list of addresses by peer? I think we could do that, but we'd want to add a timestamp to each contact info message and potentially have a peer sign each contact info. Would that be possible?

pfletcherhill (commented):
Initial batch of comments/questions

src/hub.ts
```diff
@@ -0,0 +1,344 @@
...
+  private contactTimer?: NodeJS.Timer;
+  private rocksDB: RocksDB;
+
+  engine: Engine;
+
+  constructor(options: Partial<HubOpts>) {
+    super();
+    this.options = options;
+    this.rocksDB = new RocksDB(options.rocksDBName ? options.rocksDBName : randomDbName());
+    this.engine = new Engine(this.rocksDB, options.networkUrl, options.IDRegistryAddress);
```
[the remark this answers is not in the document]
sagar-a16z: I think it's just much cleaner this way. 

I view engine as our storage layer. I would argue that we move the Database down into the engine as well (instead of managing its lifecycle outside). 

It's much cleaner to have the CRDTs and DB management live inside the Engine class and the networking and rpc layers elsewhere. 

Hub on the other hand is a collection of services where the storage backend is one of those services. The Hub bridges information between these services. 
This lets us handle events that span across the storage layer and network layers in a separate space (Hub). 

On a technical level we could totally have Engine and Hub be the same thing but this is much better from a code organization and maintenance perspective as well.

src/hub.test.ts
sagar-a16z: Ah, yeah so `catch` ends up catching the errors thrown by `except` as well. Which makes the test always pass. 
This way `expect` can still throw, but we always clean up once everything's done.

src/hub.ts
sagar-a16z: Can do.

src/hub.ts
sagar-a16z: Yeah that's a good idea. I do have a writeup about the Anatomy of a Hub here https://github.com/farcasterxyz/hub/discussions/109

src/hub.ts
sagar-a16z: I was thinking it's when testing a prod setup. The Hub will clean up after itself but perhaps there's a more explicit way of doing this. 

I'll move this logic into tests for now.

src/hub.ts
sagar-a16z: If it has not been started.

src/hub.ts
sagar-a16z: Yeah I was thinking a list of peers and their latest ContactInfo object. 

Hubs can sign stuff. Right now they just use ephemeral keys for everything but that can change. We would change the schema of the ContactInfo to support signed data but it's something we should explore.

src/hub.ts
sagar-a16z: Would be nice to have a "destroyed" status on RocksDB. Using an optional for now.

sagar-a16z: actually, ended up removing this logic altogether since it was only used by tests

src/network/node.ts
```diff
@@ -60,20 +65,48 @@ export class Node extends TypedEmitter<NodeEvents> {
...
-  async start() {
+  async start(bootstrapAddrs: Multiaddr[] | undefined = undefined) {
     this._node = await createNode();
     this.registerListeners();
 
     await this._node.start();
-    console.log('LibP2P started...' + this._node.getMultiaddrs());
+    console.log(this.identity, 'LibP2P started...');
+    console.log('listening on addresses:');
+    this._node?.getMultiaddrs().forEach((addr) => {
```
pfletcherhill, line 75: In what case would `this._node` be undefined here? Can `createNode()` fail?

src/network/node.ts
```diff
@@ -115,6 +148,21 @@ export class Node extends TypedEmitter<NodeEvents> {
     return err('Connection failure: Unable to connect to any peer address');
   }
 
+  /**
+   * Connect with a peer's NodeAddress
+   *
+   * @param address The NodeAddress to attempt a connection with
+   */
+  async connectAddress(address: Multiaddr): Promise<Result<void, string>> {
```
pfletcherhill, line 156: Can we use `FarcasterError` as the error type? I tried to move away from string errors as much as possible in my persistence PR.

src/network/protocol.ts
```diff
@@ -1,6 +1,16 @@
+import { AddressInfo } from 'net';
 import { err, ok, Result } from 'neverthrow';
-import { IDRegistryEvent, Message } from '~/types';
 import { isGossipMessage } from '~/network/typeguards';
+import { IDRegistryEvent, Message } from '~/types';
+
+// Network topic for all FC protocol messages
+export const NETWORK_TOPIC_PRIMARY = 'f_network_topic_primary';
```
pfletcherhill, line 7: Are there standards for the format of topics?

src/network/protocol.ts
```diff
@@ -44,6 +54,18 @@ export type IDRegistryContent = {
...
 };
 
+/**
+ * ContactInfo allows gossip nodes to share additional information about each other
+ * over the gossip network.
+ *
+ * @rpcAddress - The address at which this node is serving RPC requests
+ * @peerId - The peerId of the node
+ */
+export type ContactInfo = {
```
pfletcherhill, line 64: Should this be named `ContactInfoContent` so we know it's interchangeable with `IDRegistryContent` and `UserContent`?

src/hub.ts
```diff
@@ -0,0 +1,348 @@
...
+  Pending,
+  InProgress,
+  Complete,
+}
+
+interface HubEvents {
+  /**
+   * Triggered when a simple sync starts
+   */
+  sync_start: () => void;
```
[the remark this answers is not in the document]
sagar-a16z: No reason, didn't see that there was precedent.

src/hub.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  engine: Engine;
+
+  constructor(options: HubOpts) {
+    super();
+    this.options = options;
+    this.rocksDB = new RocksDB(options.rocksDBName ? options.rocksDBName : randomDbName());
+    this.engine = new Engine(this.rocksDB, options.networkUrl, options.IDRegistryAddress);
+    this.gossipNode = new Node();
+    this.rpcServer = new RPCServer(this);
```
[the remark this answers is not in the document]
sagar-a16z: After we introduced `submitMessage` to the RPC API, it made sense to move the interface to the Hub class rather than the engine. Our RPCHandler APIs right now are closely tied with Engine, but `submitMessage` is the first example of wanting to do things beyond just the engine. 

Specifically, the Hub receives a request to `submitMessage`, merges it to the engine and then sends it to all the other Hubs over gossip. 

I suspect we'll always want a way to "do some other stuff" instead of just letting the engine handle it.

src/network/protocol.ts
sagar-a16z: I think it's upto us. There's a default topic that's being used by the pubsub discovery mechanism 
`_peer-discovery._p2p._pubsub` but I went with `f_network_topic_*` for now.

src/network/protocol.ts
sagar-a16z: Yeah fair point

src/network/node.ts
sagar-a16z: Yeah I missed this after I rebased.

src/hub.ts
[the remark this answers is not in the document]
pfletcherhill: I see. Maybe it's necessary, but I don't love repeating all the query methods in the Hub. How would you compare the current approach to an alternate approach where we make the Engine an event emitter and emit `mergeMessage` events for the Hub to consume and gossip? It's also worth considering how we'll implement the [mutations API](https://github.com/farcasterxyz/hub/issues/71), where we'll want to be able to publish changes to clients that have subscribed.

src/hub.ts
pfletcherhill: Cool, do you mind making an issue in the repo to explore this concept further?

src/hub.ts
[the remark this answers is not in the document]
pfletcherhill: Sounds good, thanks for the explanation

src/hub.ts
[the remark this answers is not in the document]
sagar-a16z: Yeah we can definitely do this with an EventEmitter but I suspect we'll eventually run into a case where RPC is not directly bolted to Engine. 

I'm happy to change it either way though. It's easy enough to do at this point.

I also imagine the same might emerge for the Mutations API. Right now we could do it via events on an Engine but I think it would be cleaner to have another component manage subscriptions for us and the Hub can mediate between that component and the Engine.

src/hub.ts
sagar-a16z: https://github.com/farcasterxyz/hub/issues/135

src/hub.ts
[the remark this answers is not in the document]
pfletcherhill: All right, let's keep it as you have it, and then re-consider the organization when we implement the mutations API.

pfletcherhill (approved):

src/network/node.test.ts
```diff
@@ -48,6 +54,31 @@ const trackMessages = () => {
...
+  });
+
+  test('fails to connect with a node that has not started', async () => {
+    const node = new Node();
+    await node.start([]);
+
+    let result = await node.connectAddress(new Multiaddr());
+    expect(result.isErr()).toBeTruthy();
+
+    const offlineNode = new Node();
```
pfletcherhill, line 74: Should fix this linter warning in a follow-on PR

PR #140, opened 2022-10-03
feat: hub executable

sagar-a16z:
## Motivation

Need a Hub executable

## Change Summary

Add CLI parsing to `server.ts` and integrate Hub. 

## Merge Checklist

### DIFF
--- a/src/hub.test.ts
+++ b/src/hub.test.ts
@@ -196,4 +196,15 @@ describe('Hub negative tests', () => {
     await hub.stop();
     await hub.destroyDB();
   });
+
+  test('Starts with clear RocksDB set', async () => {
+    hub = new Hub({ ...opts, resetDB: true });
+    await hub.start();
+    expect(hub.rpcAddress).toBeDefined();
+    expect(() => {
+      hub.identity;
+    }).not.toThrow();
+    await hub.stop();
+    hub.destroyDB();
+  });
 });
--- a/src/hub.test.ts
+++ b/src/hub.test.ts
@@ -10,7 +10,8 @@ import { Message } from '~/types';
 import { ServerError } from '~/errors';
 import { jest } from '@jest/globals';
 
-const TEST_TIMEOUT = 2 * 60 * 1000;
+const TEST_TIMEOUT_SHORT = 10 * 1000;
+const TEST_TIMEOUT_LONG = 2 * 60 * 1000;
 const opts: HubOpts = { simpleSync: false };
 
 let hub: Hub;
@@ -47,7 +48,7 @@ describe('Hub running tests', () => {
     await tearDownHub(hub);
   });
 
-  test('Run a Hub and send it a message', async () => {
+  test('run a Hub and send it a message', async () => {
     expect(hub.rpcAddress).toBeTruthy();
 
     const rpcClient = new RPCClient(hub.rpcAddress as AddressInfo);
@@ -68,7 +69,7 @@ describe('Hub running tests', () => {
   });
 
   test(
-    'Bootstrap one Hub off of another',
+    'bootstrap one Hub off of another',
     async () => {
       // populate Hub1's engine
       await populateEngine(hub.engine, 5);
@@ -96,7 +97,7 @@ describe('Hub running tests', () => {
         if (shouldStop) await secondHub.stop();
       }
     },
-    TEST_TIMEOUT
+    TEST_TIMEOUT_LONG
   );
 
   const testMessage = async (message: Message) => {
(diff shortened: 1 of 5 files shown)


src/hub.test.ts
```diff
@@ -196,4 +196,15 @@ describe('Hub negative tests', () => {
     await hub.stop();
     await hub.destroyDB();
   });
+
+  test('Starts with clear RocksDB set', async () => {
```
varunsrin, line 200: nit: test descriptions are uusally lowercased by convention: 'starts with clear rocksdb set'

src/hub.test.ts
```diff
@@ -196,4 +196,15 @@ describe('Hub negative tests', () => {
...
+    hub = new Hub({ ...opts, resetDB: true });
+    await hub.start();
+    expect(hub.rpcAddress).toBeDefined();
+    expect(() => {
+      hub.identity;
+    }).not.toThrow();
+    await hub.stop();
+    hub.destroyDB();
```
varunsrin, line 208: is this duplicated because the `afterEach` hook on line 46 calls this again?

src/server.ts
```diff
@@ -1,4 +1,36 @@
...
+  port: cliOptions.rpcPort,
+  simpleSync: cliOptions.simpleSync,
+  rocksDBName: cliOptions.dbName,
+  resetDB: cliOptions.dbReset,
+};
+
+const hub = new Hub(options);
+hub.start();
+process.stdin.resume();
+process.on('SIGINT', async () => {
```
varunsrin, line 33: should we handle 'SIGTERM' and 'SIGQUIT' for thoroughness?

src/server.ts
```diff
@@ -1,4 +1,36 @@
...
+app.name('farcaster-hub').description('Farcaster Hub').version('0.1.0');
+
+app
+  .option('-n, --network-url <url>', 'ID Registry network URL')
+  .option('-a, --id-registry-address <address>', 'ID Registry address')
+  .option('-b, --bootstrap-addresses <addresses...>', 'A list of MultiAddrs to use for bootstrapping')
+  .option('--rpc-port <port>', 'The RPC port to use. (default: selects one at random')
+  .option('--simple-sync', 'Enable/Disable simple sync', true)
+  .option('--db-reset', 'Clear the database before starting', false)
+  .option('--db-name <name>', 'The name of the rocks db instance', 'rocks.hub._default');
```
varunsrin, line 15: nit: inconsistent casing across command descriptions (rocks db vs MultiAddrs)

src/server.ts
```diff
@@ -1,4 +1,36 @@
...
-console.log(hashCompare('Hello, world!', '1'));
+const app = new Command();
+
+app.name('farcaster-hub').description('Farcaster Hub').version('0.1.0');
+
+app
+  .option('-n, --network-url <url>', 'ID Registry network URL')
+  .option('-a, --id-registry-address <address>', 'ID Registry address')
+  .option('-b, --bootstrap-addresses <addresses...>', 'A list of MultiAddrs to use for bootstrapping')
+  .option('--rpc-port <port>', 'The RPC port to use. (default: selects one at random')
```
varunsrin, line 12: maybe just port? unless we think there are non-rpc ports that need to be configured

src/server.ts
```diff
@@ -1,4 +1,36 @@
-import { hashCompare } from '~/utils';
+import { Command } from 'commander';
+import { Hub, HubOpts } from '~/hub';
 
-// Import something external to make sure it works!
-console.log(hashCompare('Hello, world!', '1'));
+const app = new Command();
+
+app.name('farcaster-hub').description('Farcaster Hub').version('0.1.0');
```
varunsrin, line 6: version mismatch with package.json

src/hub.test.ts
```diff
@@ -196,4 +196,15 @@ describe('Hub negative tests', () => {
     await hub.stop();
     await hub.destroyDB();
   });
+
+  test('Starts with clear RocksDB set', async () => {
+    hub = new Hub({ ...opts, resetDB: true });
+    await hub.start();
+    expect(hub.rpcAddress).toBeDefined();
+    expect(() => {
+      hub.identity;
```
varunsrin, line 205: why does this prove that the db was reset? if i'm missing something obvious please ignore and merge, but i didn't quite follow this

varunsrin (approved):

src/hub.test.ts
sagar-a16z: Actually this is not a good test for anything. I added it because something was throwing earlier but let me see if I can rewrite it so it actually tests that the DB is reset

src/server.ts
sagar-a16z: I went back and forth on this. I prefer keeping it explicit RPC. There is a Listen port in use for libp2p, and I'm guessing we will need to make that configurable in the near future, so keeping that separate makes sense imo

src/hub.test.ts
sagar-a16z: There's two test suites in this file, the suite that this test is in doesn't have a `beforeEach`.

PR #145, opened 2022-10-05
feat(sync): Merkle trie to keep track messages merged by the engine

sanjayprabhu:
## Motivation

See https://github.com/farcasterxyz/hub/issues/66. First in a number of PRs that enable hubs to differentially sync with each other.

## Change Summary

Add a MerkleTrie implementation that keeps track of messages merged by the engine

## Merge Checklist

## Additional Context

This PR adds a merkle trie implementation that just keeps track of messages merged by the engine on gossip message received. A lot of other changes are explicitly out-of-scope to keep changesets small and will be handled in future PRs. In particular, this PR does not handle existing messages already merged into the engine. See https://github.com/farcasterxyz/hub/issues/66 for the implementation plan.

### DIFF
--- a/src/sync/merklesync.ts
+++ b/src/sync/merklesync.ts
@@ -1,0 +1,14 @@
+import { Message } from '~/types';
+import { MerkleTrie, SyncId } from '~/sync/merkletrie';
+
+export class MerkleSync {
+  private _trie: MerkleTrie;
+
+  constructor() {
+    this._trie = new MerkleTrie();
+  }
+
+  public addMessage(message: Message): void {
+    this._trie.insert(new SyncId(message));
+  }
+}
--- a/src/sync/merklesync.ts
+++ b/src/sync/merklesync.ts
@@ -2,7 +2,12 @@ import { Message } from '~/types';
 import { MerkleTrie, SyncId } from '~/sync/merkletrie';
 import Engine from '~/engine';
 
-export class MerkleSync {
+/**
+ * MerkleSync handles the logic required to determine where and how two hubs differ
+ * from each other and bring them into sync efficiently. See https://github.com/farcasterxyz/hub/issues/66
+ * for more details on design of the sync algorithm.
+ */
+class MerkleSync {
   private _trie: MerkleTrie;
   private engine: Engine;
 
@@ -23,3 +28,5 @@ export class MerkleSync {
     return this._trie;
   }
 }
+
+export { MerkleSync };
--- a/src/sync/merklesync.ts
+++ b/src/sync/merklesync.ts
@@ -1,14 +1,25 @@
 import { Message } from '~/types';
 import { MerkleTrie, SyncId } from '~/sync/merkletrie';
+import Engine from '~/engine';
 
 export class MerkleSync {
   private _trie: MerkleTrie;
+  private engine: Engine;
 
-  constructor() {
+  constructor(engine: Engine) {
     this._trie = new MerkleTrie();
+    this.engine = engine;
+
+    this.engine.on('messageMerged', async (fid, type, message) => {
+      this.addMessage(message);
+    });
   }
 
   public addMessage(message: Message): void {
     this._trie.insert(new SyncId(message));
+  }
+
+  public get trie(): MerkleTrie {
+    return this._trie;
   }
 }
--- a/src/sync/merkletrie.test.ts
+++ b/src/sync/merkletrie.test.ts
@@ -33,12 +33,12 @@ describe('merkle trie', () => {
     const syncId = new SyncId(message);
 
     expect(trie.items).toEqual(0);
-    expect(trie.hash).toEqual('');
+    expect(trie.rootHash).toEqual('');
 
     trie.insert(syncId);
 
     expect(trie.items).toEqual(1);
-    expect(trie.hash).toBeTruthy();
+    expect(trie.rootHash).toBeTruthy();
   });
 
   test('idempotency', async () => {
@@ -56,15 +56,17 @@ describe('merkle trie', () => {
     secondTrie.insert(syncId1);
 
     // Order does not matter
-    expect(firstTrie.hash).toEqual(secondTrie.hash);
+    expect(firstTrie.rootHash).toEqual(secondTrie.rootHash);
     expect(firstTrie.items).toEqual(secondTrie.items);
-    expect(firstTrie.hash).toBeTruthy();
+    expect(firstTrie.rootHash).toBeTruthy();
 
     firstTrie.insert(syncId2);
     secondTrie.insert(syncId1);
 
     // Re-adding same item does not change the hash
-    expect(firstTrie.hash).toEqual(secondTrie.hash);
+    expect(firstTrie.rootHash).toEqual(secondTrie.rootHash);
+    expect(firstTrie.items).toEqual(secondTrie.items);
+    expect(firstTrie.items).toEqual(2);
   });
 
   test('insert multiple items', async () => {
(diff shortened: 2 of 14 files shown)


src/sync/merklesync.ts
```diff
@@ -0,0 +1,32 @@
+import { Message } from '~/types';
+import { MerkleTrie, SyncId } from '~/sync/merkletrie';
+import Engine from '~/engine';
+
+/**
+ * MerkleSync handles the logic required to determine where and how two hubs differ
+ * from each other and bring them into sync efficiently. See https://github.com/farcasterxyz/hub/issues/66
```
varunsrin, line 7: nit: this link points to a ticket that points to a notion doc, which may not be a good permanent reference.

src/sync/merkletrie.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  test('insert multiple items', async () => {
+    const messages = await Factories.CastShort.createList(25);
+    const syncIds = messages.map((message) => new SyncId(message));
+
+    const firstTrie = new MerkleTrie();
+    const secondTrie = new MerkleTrie();
+
+    syncIds.forEach((syncId) => firstTrie.insert(syncId));
+    const shuffledIds = syncIds.sort((a, b) => 0.5 - Math.random());
```
varunsrin, line 94: nit: use _a and _b to make the linter stop complaining

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
+import { Message } from '~/types';
+import { createHash } from 'crypto';
+
```
varunsrin, line 3: nit: should filename be `merkleTrie` to stay consistent with capitalization?

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+const TIMESTAMP_LENGTH = 10;
+const HASH_LENGTH = 128; // We're using 64 byte blake2b hashes
+const ID_LENGTH = TIMESTAMP_LENGTH + HASH_LENGTH;
+
+/**
+ * SyncId represent the unique id of a message in the sync protocol.
```
varunsrin, line 9: this makes it sound like the reason we are combining the hash and timestamp is to get a unique id, which is a bit misleading since we know hash is unique.

i believe the reason we are doing it is to get a key that is both unique and lexicographically causally ordered?

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+    this._timestamp = message.data.signedAt;
+    this._hash = message.hash;
+  }
+
+  public toString(): string {
+    return `${this.timestampString}${this.hashString}`;
+  }
+
+  public get timestampString(): string {
+    return Math.floor(this._timestamp / 1000)
```
varunsrin, line 26: worth explaining the "why" with a comment since the purpose here is not immediately obvious

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+    this._items = 0;
+    this._children = new Map();
+    this._value = value;
+
+    if (value !== undefined) {
+      this._updateHash();
+    }
+  }
+
+  public insert(key: string, value: string, current_index = 0) {
```
varunsrin, line 57: is this guaranteed to return a boolean? i think i see some code paths where it might return undefined which might be confusing to interpret vs false

can test this by defining a return signature here and seeing if typechecker fails

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+    const char = key[current_index];
+
+    // TODO: Optimize by using MPT extension nodes for leaves
+    if (current_index === ID_LENGTH - 1) {
+      if (this._children.has(char) && this._children.get(char)?.value === value) {
```
varunsrin, line 62: might be worth adding a comment saying "this is a duplicate key-value"

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+      this._updateHash();
+      return true;
+    }
+
+    if (!this._children.has(char)) {
+      this._children.set(char, new TrieNode());
+      this._children = new Map([...this._children.entries()].sort());
+    }
+
+    const success = this._children.get(char)?.insert(key, value, current_index + 1);
```
varunsrin, line 77: add comment explaining the case - there is a child node with the character specified but it does not contain the value so we insert it (or something like that)

src/sync/merklesync.ts
```diff
@@ -0,0 +1,32 @@
...
+ * for more details on design of the sync algorithm.
+ */
+class MerkleSync {
```
varunsrin, line 10: Is the intended class relationship `Hub` owns `MerkleSync` owns `Engine` ? 

MerkleSync feels like a different naming convention, perhaps we should call it the SyncEngine vs CRDT Engine?

varunsrin (commented):
Left a first pass of comments looking at the implementation, will review with tests tomorrow

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * SyncId represent the unique id of a message in the sync protocol.
+ * It is a combination of the message's timestamp and hash
+ */
+class SyncId {
+  private readonly _timestamp: number;
+  private readonly _hash: string;
+
+  constructor(message: Message) {
```
varunsrin, line 16: if i were to somehow pass a message that had arbitrary data (by accidentally casting to any and recasting to message) is it possible to permanently bork sync? 

it could create an "unfetachable entry" and the two nodes would never coalesce. i'm not sure if we can 100% defend against such a case, but sanitizing inputs (even when typed) might prevent some errors

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  public get(key: string): string | undefined {
+    if (this.isLeaf) {
+      return this._value;
+    }
+
+    const char = key[0];
+    if (!this._children.has(char)) {
+      return undefined;
```
varunsrin, line 92: this branch isn't covered by a test case, which may be worth doing

src/sync/merkletrie.ts
```diff
@@ -0,0 +1,165 @@
...
+  public get isLeaf(): boolean {
+    return this._children.size === 0;
+  }
+
+  // Only available on leaf nodes
+  public get value(): string | undefined {
+    if (this.isLeaf) {
+      return this._value;
+    }
+    return undefined;
```
varunsrin, line 130: this branch isn't covered by a test case, which may be worth doing

varunsrin (changes requested):

src/sync/merklesync.ts
sanjayprabhu: Discussed in person, will be recorded in a doc you're creating.

src/sync/merkletrie.ts
sanjayprabhu: Good catch

src/sync/merklesync.ts
sanjayprabhu: What's a good place to move the notion doc to? The protocol spec? A docs directory inside this repo?

src/sync/merkletrie.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  describe('syncid', () => {
+    let message: Message;
+    let syncId: SyncId;
+
+    beforeEach(async () => {
+      message = await Factories.CastShort.create();
+      syncId = new SyncId(message);
+    });
+
+    test('timestamp is correct', () => {
```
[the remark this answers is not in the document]
sanjayprabhu: Testing agains the factory felt more robust long term, but I did add one test with specific strings for clarity.

src/sync/merkletrie.ts
sanjayprabhu: Yes, that can happen. The real fix for that is to be able to repair/recreate the trie from the db which will be implemented down the line. 

For now, I did add validation in the constructor to make sure the hashes conform to what we expect. The timestamp is harder to validate since any number is theoretically valid, but added a conservative length check. Together, they should be sufficient protection (but could cause a potential hard hub crash in case we get bad data).

varunsrin (approved):

PR #146, opened 2022-10-05
feat: Script to benchmark a network of Hubs

sagar-a16z:
## Motivation

We want to be able to test a "real" network of hubs with dummy data

## Change Summary

Added a benchmark script that can generate data for `N` users and broadcast 1 `castShort` for each to the network. 

## Merge Checklist

## Additional Context

This is just the first pass at the bench client. We'll presumably need more features and changes over time.

### DIFF
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -66,7 +66,10 @@ const submitInBatches = async (messages: Message[] | IDRegistryEvent[]) => {
 
 const app = new Command();
 
-app.name('farcaster-benchmark-client').description('Farcaster Benchmark').version('0.1.0');
+app
+  .name('farcaster-benchmark-client')
+  .description('Farcaster Benchmark')
+  .version(process.env.npm_package_version ?? '1.0.0');
 
 app
   .requiredOption('-a, --ip-address <address>', 'The IP address of a Hub to submit messages to')
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -1,0 +1,157 @@
+import { Command } from 'commander';
+import { RPCClient } from '~/network/rpc';
+import { AddressInfo, isIP } from 'net';
+import { generateUserInfo, getIDRegistryEvent, getSignerAdd, UserInfo } from '~/engine/mock';
+import Faker from 'faker';
+import { IDRegistryEvent, Message, SignerAdd } from '~/types';
+import { Factories } from '~/factories';
+import { Result } from 'neverthrow';
+import { sleep } from '~/utils';
+import { JSONRPCError } from 'jayson/promise';
+import { isIDRegistryEvent, isMessage } from '~/types/typeguards';
+
+const post = (msg: string, start: number, stop: number) => {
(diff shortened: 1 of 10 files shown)


src/bench.ts
```diff
@@ -0,0 +1,160 @@
...
+import { AddressInfo, isIP } from 'net';
+import { generateUserInfo, getIDRegistryEvent, getSignerAdd, UserInfo } from '~/engine/mock';
+import Faker from 'faker';
+import { IDRegistryEvent, Message, SignerAdd } from '~/types';
+import { Factories } from '~/factories';
+import { Result } from 'neverthrow';
+import { sleep } from '~/utils';
+import { JSONRPCError } from 'jayson/promise';
+import { isIDRegistryEvent, isMessage } from '~/types/typeguards';
+
```
varunsrin, line 12: a high level comment describing what this file is doing would be very helpful

tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "forceConsistentCasingInFileNames": true,
 
     /* Use ESM by default instead of CJS*/
-    "module": "ES2020",
+    "module": "ES2022",
```
varunsrin, line 15: out of curiosity, what new feature from 2022 was needed?

src/types/typeguards.ts
```diff
@@ -116,7 +116,7 @@ export const isData = (data: any): data is FC.Data => {
   );
 };
 
-export const isIDRegistryEvent = (msg: FC.IDRegistryEvent): msg is FC.IDRegistryEvent => {
+export const isIDRegistryEvent = (msg: any): msg is FC.IDRegistryEvent => {
```
varunsrin, line 119: could we type this more expressively instead of reverting to any?

src/bench.ts
```diff
@@ -0,0 +1,160 @@
+import { Command } from 'commander';
```
varunsrin, line 1: how long lived do we think this file will be? asking so that i can guide my review 

right now, it's not the easiest to understand but that's ok if we're likely to throw it away and rewrite soon. if we think this is going to be a long-lived benchmarking framework, i'll go through and leave more thorough comments

varunsrin (approved):

tsconfig.json
sagar-a16z: So, I'm using "top-level await" in the script and support for it was added in `ES2022`. 
This is the original proposal I think https://github.com/tc39/proposal-top-level-await

src/types/typeguards.ts
sagar-a16z: This is a typeguard, and is intentionally allows the comparision against `any `. (similar to other typeguards in the file)

src/bench.ts
sagar-a16z: Just put something very simple in. LMK if that's insufficient.

src/types/typeguards.ts
varunsrin: ah derp

PR #149, opened 2022-10-09
refactor: code organization

varunsrin:
## Motivation

The Hub has become more complicated and harder to reason about as the size of the codebase as grown. A high level design that we all agree on will make it much easier to read and write code as we are getting ready for the first release.

## Change Summary

The PR is very large, touches many files and squeezed in all the changes I could over a single weekend to reduce merge conflicts during the workweek. A series of smaller followup changes will be made over the next week to bring the rest in line. Comments and suggestions are welcome for other things that should be updated, or patterns that people disagree with. 

The major changes are: 

- Added new README.md that describes the proposed architecture
- Reorganized code under the high level folders `network`, `storage`, `types`, `utils` and `urls`
- Refactored `rpc` into a module, which is a pattern we should replicate over the codebase to keep files small
- Moved non-unit tests files and helpers into `test` folder
- Added documentation to `hubs.ts`
- Enforced stricter type checking rules to minimize type errors

## Merge Checklist

## Additional Context


.eslintrc.cjs
```diff
@@ -12,7 +12,7 @@ module.exports = {
     '@typescript-eslint/ban-ts-ignore': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
-    'no-restricted-imports': ['error', { patterns: ['.*'] }],
+    'no-restricted-imports': ['error', { patterns: ['"*/lib/*"'] }],
```
varunsrin, line 15: note to self: should be `build` not lib

README.md
```diff
@@ -1,18 +1,125 @@
...
+Clients[FC Clients] & Clients & Clients -.-> |RPC| Hub
+Hub <-.-> |RPC + Gossip|Hubs[FC Hubs] & Hubs & Hubs
+Storage <--> Networking
+```
+
+### Storage Engine
+
+The Storage Engine validates the signature and schema of every received message. It computes the set of valid signers for a user by tracking Farcaster contract events and uses this to verify each message.
+
+Verified messages are merged into a Set, which is a data structure represents the current state of the user's data on the network. A Set uses CRDT's to ensure that merging a message is commutative, associative and idempotent. This means that two Hubs that receive a set of messages in any order will always end up in the same state.
```
sagar-a16z, line 55: nit: which is a data structure `that` represents the

src/hub.ts
```diff
@@ -11,39 +11,44 @@ import {
...
+  /** Address of the IdRegistry contract  */
+  IDRegistryAddress?: string;
+
   /*
-   * Whether simple sync should be used once the network is established
-   * This would typically only be disabled on the first node to join a new network
+   * Enable SimpleSync once network is established.
+   *
+   * @remarks
+   * Usually disabled after the first node joins the network
```
sagar-a16z, line 40: Correction: Usually disabled _for_ the first node in a network.

src/hub.ts
```diff
@@ -85,102 +87,81 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
...
+  get rpcAddress() {
+    // Safety: RPC is always configured on an IP socket, so it can be cast safely here
+    return this.rpcServer.address ? (this.rpcServer.address as AddressInfo) : undefined;
+  }
+
+  get gossipAddresses() {
+    return this.gossipNode.multiaddrs ?? [];
+  }
+
+  // Returns the Gossip peerId string of this Hub
```
sagar-a16z, line 105: Update to this for consistency?
```
/** Returns the Gossip peerId string of this Hub */ 
```

src/network/rpc/utils.ts
```diff
@@ -0,0 +1,44 @@
+import { Result } from 'neverthrow';
```
sagar-a16z, line 1: nit: I don't think `utils` is the right name for this. It's the core definition of the RPC API implemented by RPC Servers. 

Can just call it `api`, `interfaces` or something along those lines. The `replacer` and `reviver` here are utils though but the interface declarations here are more important.

sagar-a16z (approved):

src/hub.ts
varunsrin: good catch

src/network/rpc/utils.ts
varunsrin: yeah, agreed

PR #158, opened 2022-10-11
chore: add Dockerfile to build container

varunsrin:
## Motivation

Prepare the application to be deployed on AWS

## Change Summary

Add a Dockerfile that can be use to start a container with a running node and expose relevant ports.

## Merge Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,27 @@
-FROM node:16
+FROM node:18
 
-# Uncomment and set this to the relevant Farcaster node port
-# EXPOSE 8000
+# Set non-root user and expose port 8080
+USER node
+EXPOSE 8080
 
-COPY . .
-RUN yarn install
+WORKDIR /home/node/app
+
+# Copy dependency information and install dependencies
+COPY --chown=node:node tsconfig.json package.json yarn.lock ./
+
+# Use the frozen lockfile to speed up the install and prevent package drift
+RUN yarn install --frozen-lockfile 
+
+# Copy source code (and all other relevant files)
+COPY --chown=node:node src ./src
+
+# Build code
 RUN yarn build
 
-CMD ["yarn", "server"]
+CMD [ "yarn", "start" ]
+
+# TODO: Use a multi-stage build as describe below to improve security and reduce build sizes
+# https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/docker/multi_stage_builds.md
+#
+# This is hard to implement today because of our dependency on the tsx runner which 
+# cannot be invoked on a compiled js file and instead requires the entire src folder to be available.
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+**/node_modules/
+**/.git
+**/README.md
+**/LICENSE
+**/.vscode
+**/npm-debug.log
+**/coverage
+**/.env
+**/.editorconfig
+**/.aws
+**/.github
+**/.rocks
+**/build
+**/.husky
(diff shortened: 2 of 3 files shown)


varunsrin:
@sagar-a16z do any specific ports need to be exposed for libp2p to work correctly?

sagar-a16z:
Well...yes. 

I guess we need to pick a couple ports. 

One for RPC and the other for libp2p to listen on.

The latter is not presently configurable or exposed. I'll push a PR up for that first thing tomorrow.

varunsrin:
@sds tagging you for a review in case you know any obvious optimizations / improvements that I am missing

Dockerfile
```diff
@@ -1,10 +1,50 @@
...
+RUN yarn install --frozen-lockfile --production
+
+# Copy results from previous stage
+COPY --chown=node:node --from=build /home/node/app/build ./build
+
+# TODO: determine if this can be removed while using tsx (or find alternative) 
+# since we should be able to run with just the compiled javascript in build/
+COPY --chown=node:node --from=build /home/node/app/src ./src
+
+CMD [ "yarn", "start" ]
```
sds, line 50: I would suggest running the `node` executable directly rather than via `yarn`, since `yarn` adds to the process tree and in my experience that can lead to signals sometimes getting mysteriously lost (and thus container shutdown isn't as graceful).

To be clear: this would likely work fine, but out of an abundance of caution I would write:
```
CMD ["npx", "tsx", ...]
```

Taking a step back, I don't think you need to invoke `tsx` at all since you've built the code into raw JS files. You should be able to just run your entrypoint via `node`:

```
CMD ["node", "path/to/main.js"]
```

This would also allow you to omit the original TypeScript files.

In any case, this should work fine as-is! Not a blocker.

Dockerfile
```diff
@@ -1,10 +1,50 @@
...
+# Set non-root user and expose port 8080
+USER node
+EXPOSE 8080
+
+WORKDIR /home/node/app
+
+# Copy dependency information and install dependencies
+COPY --chown=node:node tsconfig.json package.json yarn.lock ./
+
+RUN yarn install --frozen-lockfile --production
```
sds, line 41: This is likely fine for now, but there are ways to reduce the image size even further (though I don't think we should block on them). See:

https://github.com/merkle-manufactory/backend/blob/85cb0581603970c92c01a446243d635fe20a43eb/Dockerfile#L32

sds (commented):
Overall looks great.

Dockerfile
varunsrin: that doesn't work unfortunately due to messiness with ESM 

tsx does some magic node configuration that imports all our packages well

Dockerfile
varunsrin: what are they 😄  ?

Dockerfile
varunsrin: but i like your suggestion of directly calling tsx instead of yarn, that's probably a good first step to optimize here

varunsrin: actually, im just going to open a ticket to look into the "right fix" of not having to call tsx on the typescript files to get things to work

Dockerfile
sds: [`node-prune`](https://github.com/tj/node-prune) has been useful for reducing size in other contexts. Also, depending on what your `postinstall` scripts do, you can add the `--ignore-scripts` to the `yarn install` command to avoid running those scripts, but that is highly dependent on whether those scripts do something important in a production context.

PR #159, opened 2022-10-11
fix: allow configuration of libp2p listen port

sagar-a16z:
## Motivation

Hub operators need to know exactly which ports to make externally reachable. 
Just like RPC, the libp2p listen port also needs to be configurable.

## Change Summary

Add an option to configure which port libp2p binds to

## Merge Checklist

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -16,6 +16,7 @@ app
   .option('-N, --network-url <url>', 'ID Registry network URL')
   .option('-A, --id-registry-address <address>', 'ID Registry address')
   .option('-B, --bootstrap-addresses <addresses...>', 'A list of MultiAddrs to use for bootstrapping')
+  .option('--port <port>', 'The port libp2p should listen on. (default: selects one at random')
   .option('--rpc-port <port>', 'The RPC port to use. (default: selects one at random')
   .option('--simple-sync <enabled>', 'Enable/Disable simple sync', true)
   .option('--db-reset', 'Clear the database before starting', false)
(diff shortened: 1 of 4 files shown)


src/cli.ts
```diff
@@ -16,6 +16,7 @@ app
   .option('-N, --network-url <url>', 'ID Registry network URL')
   .option('-A, --id-registry-address <address>', 'ID Registry address')
   .option('-B, --bootstrap-addresses <addresses...>', 'A list of MultiAddrs to use for bootstrapping')
+  .option('--port <port>', 'The port libp2p should listen on. (default: selects one at random')
```
varunsrin, line 19: we should add a default port to the `yarn start` command in package.json so that we have a consistent port on every bootup in the cloud

src/storage/engine/engine.mock.test.ts
```diff
@@ -6,35 +6,40 @@ import { mockEvents, MockFCEvent, mockFid, populateEngine } from '~/storage/engi
...
 
-    await mockEvents(engine, [user], 1, MockFCEvent.Reaction);
-    const reactions = await engine.getAllReactionsByUser(fid);
-    expect(reactions.size).toEqual(2);
-  });
+      await mockEvents(engine, [user], 1, MockFCEvent.Reaction);
+      const reactions = await engine.getAllReactionsByUser(fid);
+      expect(reactions.size).toEqual(2);
+    },
+    TEST_TIMEOUT_SHORT
```
varunsrin, line 41: this doesn't seem to be doing anything fancy, is there a reason we want to enforce timeouts here instead of just using the default jest configs?

varunsrin (changes requested):

src/storage/engine/engine.mock.test.ts
sagar-a16z: This test seems to time out randomly in CI. Seems like the default 5s jest timeout is right on the limit of what this ends up needing in CI. 

(see previous CI run on the first commit here that failed)

src/cli.ts
sagar-a16z: I think the script that calls `yarn start` in the cloud should be supplying those defaults.

varunsrin (approved):

PR #170, opened 2022-10-13
chore: add more detail to hub, node, and rpc logs

sagar-a16z:
## Change Summary

Updated logs in hub, node, and rpc to carry component and context information where applicable. 

## Merge Checklist

### DIFF
--- a/src/hub.ts
+++ b/src/hub.ts
@@ -79,6 +79,10 @@ interface HubEvents {
   syncComplete: (success: boolean) => void;
 }
 
+const log = logger.child({
+  component: 'Hub',
+});
+
 export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
   private options: HubOpts;
   private gossipNode: Node;
@@ -125,7 +129,7 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
   async start() {
     await this.rocksDB.open();
     if (this.options.resetDB === true) {
-      logger.info('Clearing RocksDB...');
+      log.info('Clearing RocksDB...');
       await this.rocksDB.clear();
     }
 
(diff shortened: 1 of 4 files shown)


src/hub.ts
```diff
@@ -197,7 +201,11 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
     if (this.syncState != SimpleSyncState.Pending) return;
 
     this.emit('syncStart');
-    logger.info(this.identity, 'Attempting to sync from Peer', peer);
+    log.info(
+      { context: { function: 'simpleSyncFromPeer', identity: this.identity, peer: peer } },
```
varunsrin, line 205: one thing that might be useful is to have a single top level object with all the props we care about (component, function etc) 

From reading docs I believe you can just do: 

```
const log = logger.child({
  component: 'Hub',
});


log.info({ function: 'simpleSyncFromPeer', identity: this.identity, peer: peer } , hello world);
```

varunsrin: this *might* save you a lot of future typing where you can just query the top level object instead of querying sub objects

src/hub.ts
```diff
@@ -240,9 +257,9 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
           // get the signer messages first so we can prepare to ingest the remaining messages later
           const custodyEventResult = await rpcClient.getCustodyEventByUser(user);
           if (custodyEventResult.isErr()) {
-            logger.info(
-              this.identity,
-              custodyEventResult,
+            log.error(
+              custodyEventResult.error.message,
```
varunsrin, line 261: the pino api only supports one object and one string for logging purposes. any additional params must be explicitly interpolated into the string or they will be ignored. the one object will be merged with any previous objects held by parents. 

see more here: https://github.com/pinojs/pino/blob/master/docs/api.md#logger

you can verify this with the following; 

```
logger.info({ component: 'CLI' }, 'Starting CLI', { version: '0.0.1' });
```

which outputs only: 

```
{"level":30,"time":1665702586129,"pid":43668,"hostname":"Varuns-MBP.home","component":"CLI","msg":"Starting CLI"}
``` 

and doesn't track the version number.

src/network/p2p/node.ts
```diff
@@ -123,15 +128,15 @@ export class Node extends TypedEmitter<NodeEvents> {
...
     const encodedMessage = encodeMessage(message);
     encodedMessage.match(
       async (msg) => {
         const results = await Promise.all(topics.map((topic) => this.gossip?.publish(topic, msg)));
-        logger.debug(this.identity, ': Published to ' + results.length + ' peers');
+        log.debug({ context: { identity: this.identity } }, 'Published to ' + results.length + ' peers');
       },
       async (err) => {
-        logger.error(this.identity, err, '. Failed to publish message.');
+        log.error(err, { context: { identity: this.identity } }, 'Failed to publish message.');
```
varunsrin, line 139: see earlier message about params, this isn't going to output what you want it to output

varunsrin (changes requested):

varunsrin (approved):

PR #172, opened 2022-10-14
feat: Option to specify Hub IP address

sagar-a16z:
## Motivation

Hub's are identified by their [libp2p MultiAddrs](https://docs.libp2p.io/concepts/addressing/). 

Hub's don't know their public IPs and by default, they will bind to `localhost` and advertise that address in their logs.
For example:
```
/ip4/127.0.0.1/tcp/61597/p2p/12D3KooWSjZrKzRXJWo3ykD1Qa7tvpaztjcumc31GWpEZnyGJw6m
```  

It's useful to share the Hub's MultiAddr with other Hub operators (for bootstrapping) or third parties. 
To make the log a little bit more helpful, we want to be able to specify which address the Hub tries to bind to. 

## Change Summary

Add an option to the cli to specify the Hub's IP address as a MultiAddr format string. 

Example: 
```
$ yarn start --ip /ip6/2600:1700:6cf0:990:2052:a166:fb35:830a
[10:58:23.401] INFO (83011): Starting libp2p
    component: "Node"
    identity: "12D3KooWSjZrKzRXJWo3ykD1Qa7tvpaztjcumc31GWpEZnyGJw6m"
    addresses: [
"/ip6/2600:1700:6cf0:990:2052:a166:fb35:830a/tcp/61597/p2p/12D3KooWSjZrKzRXJWo3ykD1Qa7tvpaztjcumc31GWpEZnyGJw6m"
    ]
```

## Merge Checklist

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 import { Command } from 'commander';
-import { Hub, HubOpts } from '~/hub';
+import { Hub, HubOptions } from '~/hub';
 import { createEd25519PeerId, createFromProtobuf, exportToProtobuf } from '@libp2p/peer-id-factory';
 import { writeFile, mkdir } from 'fs/promises';
 import { existsSync } from 'fs';
@@ -31,8 +31,9 @@ app
   .option('-A, --id-registry-address <address>', 'ID Registry address')
   .option('-B, --bootstrap-addresses <addresses...>', 'A list of MultiAddrs to use for bootstrapping')
   .option('--allowed-peers <peerIds...>', 'An "allow list" of Peer Ids. Blocks all other connections')
-  .option('--port <port>', 'The port libp2p should listen on. (default: selects one at random')
-  .option('--rpc-port <port>', 'The RPC port to use. (default: selects one at random')
+  .option('--ip <ip-multiaddr>', 'The IP Multi Address libp2p should listen on. (default: "/ip4/127.0.0.1/")')
+  .option('--port <port>', 'The TCP port libp2p should listen on. (default: selects one at random)')
+  .option('--rpc-port <port>', 'The RPC port to use. (default: selects one at random)')
   .option('--simple-sync <enabled>', 'Enable/Disable simple sync', true)
   .option('--db-reset', 'Clear the database before starting', false)
   .option('--db-name <name>', 'The name of the RocksDB instance', 'rocks.hub._default')
@@ -43,12 +44,13 @@ app
       process.exit();
     };
 
-    const options: HubOpts = {
+    const options: HubOptions = {
       peerId: await readPeerId(cliOptions.id),
       networkUrl: cliOptions.networkUrl,
       IDRegistryAddress: cliOptions.idRegistryAddress,
       bootstrapAddrs: cliOptions.bootstrapAddresses,
       allowedPeers: cliOptions.allowedPeers,
+      IPMultiAddr: cliOptions.ip,
       port: cliOptions.port,
       rpcPort: cliOptions.rpcPort,
       simpleSync: cliOptions.simpleSync,
--- a/src/network/p2p/node.test.ts
+++ b/src/network/p2p/node.test.ts
@@ -90,7 +90,7 @@ describe('node unit tests', () => {
       // node 3 has node 1 in its allow list, but not node 2
       const node3 = new Node();
       if (node1.peerId) {
-        await node3.start([], [node1.peerId.toString()]);
+        await node3.start([], { allowedPeerIdStrs: [node1.peerId.toString()] });
       } else {
         throw Error('Node1 not started, no peerId found');
       }
@@ -109,6 +109,25 @@ describe('node unit tests', () => {
         await node2.stop();
         await node3.stop();
       }
+    },
+    TEST_TIMEOUT_SHORT
+  );
+
+  test(
+    'handles invalid start options',
+    async () => {
+      const node = new Node();
+      // port and transport addrs in the IP MultiAddr is not allowed
+      let options = { IPMultiAddr: '/ip4/127.0.0.1/tcp/8080' };
+      await expect(node.start([], options)).rejects.toThrow();
+      expect(node.isStarted()).toBeFalsy();
+      await node.stop();
+
+      // an IPv6 being supplied as an IPv4
+      options = { IPMultiAddr: '/ip4/2600:1700:6cf0:990:2052:a166:fb35:830a' };
+      await expect(node.start([], options)).rejects.toThrow();
+      expect(node.isStarted()).toBeFalsy();
+      await node.stop();
     },
     TEST_TIMEOUT_SHORT
   );
--- a/src/network/p2p/node.ts
+++ b/src/network/p2p/node.ts
@@ -5,7 +5,7 @@ import { PeerId } from '@libp2p/interface-peer-id';
 import { Mplex } from '@libp2p/mplex';
 import { PubSubPeerDiscovery } from '@libp2p/pubsub-peer-discovery';
 import { TCP } from '@libp2p/tcp';
-import { Multiaddr } from '@multiformats/multiaddr';
+import { multiaddr, Multiaddr } from '@multiformats/multiaddr';
 import { createLibp2p, Libp2p } from 'libp2p';
 import { err, ok, Result } from 'neverthrow';
 import { TypedEmitter } from 'tiny-typed-emitter';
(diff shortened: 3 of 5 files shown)


sds (commented):
Is the prefixing with `/ip6` a standard used elsewhere? I naïvely thought we could infer the IP version from the string passed, but if there are edge cases perhaps we should use explicit flags?

Also, I could see wanting the ability to bind to a private network IPV4 address (for internal traffic to the hub) while also bound to a public network IPV6 address. Is that achievable with the libraries we're using?

sagar-a16z:
> Is the prefixing with `/ip6` a standard used elsewhere? I naïvely thought we could infer the IP version from the string passed, but if there are edge cases perhaps we should use explicit flags?

As I was talking about on earlier, I didn't want to build/import a module to parse IP strings when we already have libp2p. 
The libp2p addressing model uses these Multi Address strings and so I think it's best that we use these strings to specify addresses. https://docs.libp2p.io/concepts/addressing/

 > Also, I could see wanting the ability to bind to a private network IPV4 address (for internal traffic to the hub) while also bound to a public network IPV6 address. Is that achievable with the libraries we're using?

Yup, libp2p can be bound to multiple addresses and protocols (quic/tcp/websocket etc) simultaneously. 
The Hubs currently don't allow for this since the address being configured here is _only_ for the Gossip mesh. 
It's simple enough to add however. It didn't seem like we have an immediate need.

varunsrin:
@sagar-a16z the option ` The PR has been tagged with the appropriate change type label(s) (i.e. documentation, feature, bugfix, or chore)` is checked but no labels were applied

I've added the label for now, but just FYI for the future

sagar-a16z:
> @sagar-a16z the option ` The PR has been tagged with the appropriate change type label(s) (i.e. documentation, feature, bugfix, or chore)` is checked but no labels were applied
> 
> 
> 
> I've added the label for now, but just FYI for the future

I've been forgetting to do that. My bad!

src/network/p2p/node.test.ts
```diff
@@ -112,6 +112,25 @@ describe('node unit tests', () => {
     },
     TEST_TIMEOUT_SHORT
   );
+
+  test(
```
varunsrin, line 116: nit: i would write these as two separate tests and describe them in the test title instead of using a generic title and describing the cases with comments (unless there is a big perf hit to doing so).

src/network/p2p/node.ts
```diff
@@ -30,6 +30,18 @@ interface NodeEvents {
   peerDisconnect: (connection: Connection) => void;
 }
 
+/** Node create options */
+interface StartOptions {
+  /** PeerId to use as the Node's Identity. Generates a new ephemeral PeerId if not specified*/
+  peerId?: PeerId | undefined;
+  /** IP address in MultiAddr format to bind to */
+  IPMultiAddr?: string | undefined;
```
varunsrin, line 38: nit: camelCase

src/network/p2p/node.ts
```diff
@@ -225,31 +235,81 @@ export class Node extends TypedEmitter<NodeEvents> {
...
       globalSignaturePolicy: 'StrictSign',
     });
 
     let connectionGater: ConnectionFilter | undefined;
-    if (allowedPeerIdStrs) {
+    if (options.allowedPeerIdStrs) {
       log.info(
-        { identity: this.identity, function: 'createNode', allowedPeerIdStrs },
-        `!!! PEER-ID RESTRICTIONS ENABLED !!!\nAllowed Peers: ${allowedPeerIdStrs}`
+        { identity: this.identity, function: 'createNode', allowedPeerIds: options.allowedPeerIdStrs },
```
varunsrin, line 301: 1. multiline logging with \n does not work well in server-side logging cases
2. if you're passing  `options.allowedPeerIdStrs` as an object, no need to pass it in again in the message
3. you may want to include the peerId that is being blocked in the object (i don't see it)
4. The message can just be a statement since all the context is in the object (e.g. "PeerID not in allow-list")

src/network/p2p/node.ts
```diff
@@ -225,31 +235,81 @@ export class Node extends TypedEmitter<NodeEvents> {
...
+    try {
+      return ok(multiaddr(multiaddrStr));
+    } catch (error: any) {
+      return err(new ServerError('Invalid MultiAddr: ' + error.message));
+    }
+  }
+
+  /** Checks that the IP address to bind to is valid and that the combined IP, transport, and port multiaddr is valid  */
+  private checkAddrs(listenIPAddr: string, listenCombinedAddr: string) {
+    // check that the provided IP addr is a valid multiaddr format
```
varunsrin, line 249: nit: extraneous comment, the previous comment in 247 says this is happening and the code is descriptive enough

src/network/p2p/node.ts
```diff
@@ -225,31 +235,81 @@ export class Node extends TypedEmitter<NodeEvents> {
...
+    let parsedAddr = this.parseAddress(listenIPAddr);
+    parsedAddr.match(
+      (addr) => {
+        if (!addr) throw new Error('Invalid IP MultiAddr. Unable to create p2p Node');
```
varunsrin, line 253: nit: `Hub bound to Invalid IP Multiaddr` is more helpful when the error is viewed out of context

src/network/p2p/node.ts
```diff
@@ -225,31 +235,81 @@ export class Node extends TypedEmitter<NodeEvents> {
...
+    // check that the combined address is actually valid
+    parsedAddr = this.parseAddress(listenCombinedAddr);
+    parsedAddr.match(
+      (addr) => {
+        if (!addr) throw new Error('Invalid Node MultiAddr. Unable to create p2p Node');
+        const options = addr.toOptions();
+        if (options.transport != 'tcp') throw new ServerError('Invalid Node MultiAddr: transport must be tcp');
+      },
+      (error) => {
+        throw error;
```
varunsrin, line 278: this case isn't covered by tests. if its not important for coverage, i'd add a line explaining why for future audits

src/network/p2p/node.ts
```diff
@@ -30,6 +30,18 @@ interface NodeEvents {
   peerDisconnect: (connection: Connection) => void;
 }
 
+/** Node create options */
+interface StartOptions {
```
varunsrin, line 34: what about `NodeOptions` or just `Options` instead of `StartOptions` ?

varunsrin (approved):

src/cli.ts
```diff
@@ -43,12 +44,13 @@ app
...
     };
 
-    const options: HubOpts = {
+    const options: HubOptions = {
       peerId: await readPeerId(cliOptions.id),
       networkUrl: cliOptions.networkUrl,
       IDRegistryAddress: cliOptions.idRegistryAddress,
       bootstrapAddrs: cliOptions.bootstrapAddresses,
       allowedPeers: cliOptions.allowedPeers,
+      IPMultiAddr: cliOptions.ip,
```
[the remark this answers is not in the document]
sagar-a16z: I think the rest of the codebase still uses `IDRegistry` so this would be a general refactor of all that.

src/network/p2p/node.ts
sagar-a16z: > 3\. you may want to include the peerId that is being blocked in the object (i don't see it)

You didn't see it here because this is just the init call. The actual filter takes place in the "ConnectionFilter" class. I'll add some debug logs to it. 

> 4\. The message can just be a statement since all the context is in the object (e.g. "PeerID not in allow-list")

This peer allow-list is a "temporary" solution and so I want it to be really visually loud when it's enabled.

src/network/p2p/node.ts
sagar-a16z: Missed cleanup. I tend to just write comments and some tests before I start writing the code. Sometimes this slips through 😇

src/network/p2p/node.ts
sagar-a16z: Since this is inside the CheckAddrs funciton, I wanted the error cases to be immediately clear about what happened. (`Invalid IP MultiAddr` vs `Invalid Node MultiAddr`). I'll update this log to `Invalid IP MultiAddr: Unable to bind Hub`.

src/network/p2p/node.ts
sagar-a16z: refactoring this out into utils so that I can unit test it. Current code path cannot reach that line in `node.ts`.

src/network/p2p/node.ts
sagar-a16z: Ended up refactoring this out of here so it looks a little bit different overall.

varunsrin (approved):

PR #173, opened 2022-10-14
Hub config file

sagar-a16z:
## Motivation

Too many knobs to control via cli. A config file lets us define defaults. 

## Change Summary

Add the option to specify a config file to use to start a Hub.

Also, removed `Commander` option defaults as there's no way to differentiate between user specified options and `Commander` defaults in an `action` handler.

## Merge Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -9,6 +9,7 @@ module.exports = {
     'plugin:security/recommended',
   ],
   plugins: ['prefer-arrow-functions'],
+  ignorePatterns: ['start.config.js'],
   rules: {
     '@typescript-eslint/ban-ts-ignore': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off',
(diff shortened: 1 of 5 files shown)


.eslintrc.cjs
```diff
@@ -9,6 +9,7 @@ module.exports = {
     'plugin:security/recommended',
   ],
   plugins: ['prefer-arrow-functions'],
+  ignorePatterns: ['start.config.js'],
```
varunsrin, line 12: what about just ignoring `**/.hub/` instead?

.gitignore
```diff
@@ -25,4 +25,7 @@ coverage/
 # Databases
 .level/
 .rocks/
-.hub/
+
+# Config
+.hub/*
+!.hub/start.config.ts
```
varunsrin, line 31: should we keep this somewhere else? i generally don't like mixing ignore and includes in the same folder, leads to footguns down the line

.hub/start.config.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
varunsrin, line 1: consider making this a .ts config file outside of .hub so that: 

1. the type definitions are available and invalid params cannot be passed
2. documenting params here is unnecessary since the use can just view the types

varunsrin (approved):

PR #179, opened 2022-10-20
feat: Implement diff sync

sanjayprabhu:
## Motivation
See https://github.com/farcasterxyz/hub/issues/66. Allows diff sync between two hubs. 

TODO:
- [x] A lot more tests

## Change Summary

Implements the RPC calls and SyncEngine logic to determine which messages are missing and sync only those messages between two hubs.

## Merge Checklist

## Additional Context

Some differences between the original design and the implementation:
 - Instead of sending the root hash with every message, we only send it in the ContactInfo message. This was originally because I couldn't find a good way to determine the peer address we received a gossip message from. But it turned out to be a better approach (no need to send redundant data on every message).
 - We have a "sync threshold", and send the excluded hashes for the most recent timestamp that was a multiple of the sync threshold. This helps prevent "trashing" during high traffic. 
 - Also removed simple sync entirely, because diff sync can completely replace it's behavior right now and there's no reason to keep both modes of sync around


src/hub.ts
```diff
@@ -244,67 +254,9 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
...
-      },
-      async (error) => {
-        log.error(error, 'sync failure: failed to get users');
-      }
-    );
-    log.info({ function: 'simpleSyncFromPeer', identity: this.identity, peer: peer }, 'sync complete');
-    this.emit('syncComplete', true);
-    this.syncState = SimpleSyncState.Complete;
+    // Validate connection
+    await rpcClient.getUsers();
```
sanjayprabhu, line 258: Whoops, this was just for testing. Will remove.

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
     this._trie.insert(new SyncId(message));
   }
 
+  public shouldSync(excludedHashes: string[], numMessages: number): boolean {
+    if (this._isSyncing) {
+      log.debug('shouldSync: already syncing');
```
varunsrin, line 44: debug logs are not enabled by default on production, if you want to visualize these when we launch the hubs you'd need to set a special env var or change them to log.info

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+      return false;
+    }
+    const ourSnapshot = this.snapshot;
+    const excludedHashesMatch =
```
varunsrin, line 48: not: this function and the next one are hard to parse because of the density of code. 

would recommend adding some spaces to create visual separation between logical blocks of code within method

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+      // Excluded hashes match exactly, so we don't need to sync
+      log.debug('shouldSync: excluded hashes match');
+      return false;
+    }
+    if (ourSnapshot.numMessages > numMessages) {
+      log.debug('shouldSync: we have more messages');
+      // We have more messages than the other hub, we don't need to sync
+      return false;
+    } else if (ourSnapshot.numMessages === numMessages) {
+      // We have the same number of messages as the other hub, randomly return true or false to determine if we should sync
```
varunsrin, line 61: i don't understand why this is random here?

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+      const result = Math.random() < 0.5;
+      log.debug(`shouldSync: random result: ${result}`);
+      return result;
+    } else if (ourSnapshot.numMessages < numMessages) {
```
varunsrin, line 65: if you do this as an `else` block instead of an `else if` then codecov will stop complaining about line 70 being uncovered and you can also safely remove line 70

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+    const theirNodeResult = await rpcClient.getSyncMetadataByPrefix(prefix);
+
+    const missingHashes: string[] = [];
+    await theirNodeResult.match(
+      async (theirNode) => {
+        if (theirNode.numMessages <= HASHES_PER_FETCH) {
+          const result = await rpcClient.getSyncIdsByPrefix(prefix);
+          result.match(
+            (ids) => {
+              missingHashes.push(...ids);
```
varunsrin, line 103: the nested match inside conditional inside match is a bit hard to follow as a reader. is there an easier way this can be expressed for future readers/editors to follow more easily?

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+            },
+            (err) => {
+              log.warn({ err }, `Error fetching ids for prefix ${prefix}`);
+            }
+          );
+        } else if (theirNode.children) {
```
varunsrin, line 109: it took me some time to parse out why we were checking for children here after the previous conditional

it would be easier to follow if we handled the (!children) case first, and then did a conditional based on number o of children

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+      async (err) => {
+        log.warn({ err }, `Error fetching metadata for prefix ${prefix}`);
+      }
+    );
+    return missingHashes;
+  }
+
+  public async fetchAndMergeMessages(hashes: string[], rpcClient: RPCClient): Promise<boolean> {
+    let result = true;
+    if (hashes.length === 0) {
```
varunsrin, line 127: nit: prefer the style `return false if (hashes.length === 0)`

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+      return false;
+    }
+    const messages = await rpcClient.getMessagesByHashes(hashes.map((hash) => '0x' + hash));
```
varunsrin, line 130: is this safe to be an unbounded number of hashes?

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+    await messages.match(
+      async (msgs) => {
+        const mergeResults = [];
+        // Sync messages sequentially, so we can handle missing users. We could optimize by collecting all
+        // failures and retrying them in a batch, but this is simpler for now.
```
varunsrin, line 135: 1. should this be merge, not sync?
2. prefer comments of the active, present-tense doing form per the [style guide](https://github.com/farcasterxyz/hub/blob/main/CONTRIBUTING.md#32-writing-docs) 

suggestion based on my understanding:  

```
Merges messages into the engine sequentially, performing a user sync if an unknown user is encountered.

TODO: optimize by collecting failures and handling them in a batch. 
```

see style guide here:

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+            mergeResults.push(result);
+          }
+        }
+        log.info(
+          { messages: mergeResults.length, success: mergeResults.filter((r) => r.isOk()).length },
+          'Merged messages'
+        );
+      },
+      async () => {
+        result = false;
```
varunsrin, line 152: this line isn't covered by tests. 

if unimportant, id recommend adding a comment here

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+    // second resolution is too fine grained, and fall outside sync threshold anyway
+    return this._trie.getSnapshot((this.snapshotTimestamp / 10).toString());
+  }
+
+  public get isSyncing(): boolean {
+    return this._isSyncing;
+  }
+
+  // Returns the most recent timestamp in seconds that's within the sync threshold
+  // (i.e. highest timestamp that's < current time and timestamp % sync_thershold == 0)
```
varunsrin, line 181: nit: typo in `sync_thershold`

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+  public get snapshotTimestamp(): number {
+    const currentTimeInSeconds = Math.floor(Date.now() / 1000);
+    return Math.floor(currentTimeInSeconds / SYNC_THRESHOLD_IN_SECONDS) * SYNC_THRESHOLD_IN_SECONDS;
+  }
+
+  private async syncUserAndRetryMessage(message: Message, rpcClient: RPCClient): Promise<Result<void, FarcasterError>> {
+    const user = message.data.fid;
+    const custodyEventResult = await rpcClient.getCustodyEventByUser(user);
+    if (custodyEventResult.isErr()) {
```
varunsrin, line 190: why not just pass the Err object up the chain?

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+
+  // Returns the most recent timestamp in seconds that's within the sync threshold
+  // (i.e. highest timestamp that's < current time and timestamp % sync_thershold == 0)
+  public get snapshotTimestamp(): number {
+    const currentTimeInSeconds = Math.floor(Date.now() / 1000);
+    return Math.floor(currentTimeInSeconds / SYNC_THRESHOLD_IN_SECONDS) * SYNC_THRESHOLD_IN_SECONDS;
+  }
+
+  private async syncUserAndRetryMessage(message: Message, rpcClient: RPCClient): Promise<Result<void, FarcasterError>> {
+    const user = message.data.fid;
```
varunsrin, line 188: should this be fid not user? also applies to the method name `getCustodyEventByUser` which is better understood if it is called `getCustodyEventByFid`

varunsrin (commented):
I did a first pass over the main sync methods, found no issues except minor style feedback

Will do another pass over the test methods tomorrow

varunsrin:
@sagar-a16z i suspect this is probably worth a code review from you as well, given that it touches on some of the p2p sync stuff

src/hub.ts
```diff
@@ -152,8 +141,13 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
           ? { peerId: this.gossipNode.peerId.toString(), rpcAddress: this.rpcAddress }
           : { peerId: this.gossipNode.peerId.toString() };
 
+        const currentSnapshot = this.syncEngine.snapshot;
         const gossipMessage: GossipMessage<ContactInfoContent> = {
-          content,
+          content: {
+            ...content,
```
sagar-a16z, line 147: nit: rename variable `content` to `contact` or `address` for clarity

src/network/sync/merkleTrie.ts
```diff
@@ -1,6 +1,19 @@
 import { createHash } from 'crypto';
 import { ID_LENGTH, SyncId } from '~/network/sync/syncId';
 
+export type TrieSnapshot = {
+  prefix: string;
+  excludedHashes: string[];
+  numMessages: number;
+};
+
+export type NodeMetadata = {
```
sagar-a16z, line 10: nit: comments as to what these types represent would be nice

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+    }
+    return false;
+  }
+
+  async performSync(excludedHashes: string[], rpcClient: RPCClient) {
+    try {
+      this._isSyncing = true;
+      const ourSnapshot = this.snapshot;
+      const divergencePrefix = this._trie.getDivergencePrefix(ourSnapshot.prefix, excludedHashes);
+      log.debug({ divergencePrefix, prefix: ourSnapshot.prefix }, 'Divergence prefix');
```
sagar-a16z, line 78: Would be useful to have the logs here be `info` for now

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+            },
+            (err) => {
+              log.warn({ err }, `Error fetching ids for prefix ${prefix}`);
+            }
+          );
+        } else if (theirNode.children) {
+          for (const [theirChildChar, theirChild] of theirNode.children.entries()) {
+            // recursively fetch hashes for every node where the hashes don't match
+            if (ourNode?.children?.get(theirChildChar)?.hash !== theirChild.hash) {
+              missingHashes.push(...(await this.fetchMissingHashes(theirChild.prefix, rpcClient)));
```
sagar-a16z, line 113: nit: split lines for readability

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+    }
+    const messages = await rpcClient.getMessagesByHashes(hashes.map((hash) => '0x' + hash));
+    await messages.match(
+      async (msgs) => {
+        const mergeResults = [];
+        // Sync messages sequentially, so we can handle missing users. We could optimize by collecting all
+        // failures and retrying them in a batch, but this is simpler for now.
+        for (const msg of msgs) {
+          const result = await this.engine.mergeMessage(msg);
+          if (result.isErr() && result.error.message.includes('unknown user')) {
```
sagar-a16z, line 138: Looks like we should add new error types for these sync errors. So that this can just look at the error code instead of a string. 

@pfletcherhill thoughts?

src/network/sync/syncEngine.ts
```diff
@@ -25,9 +39,172 @@ class SyncEngine {
...
+      return err(new ServerError('Failed to fetch custody event'));
+    }
+    await this.engine.mergeIdRegistryEvent(custodyEventResult._unsafeUnwrap());
+    // Probably not required to fetch the signer messages, but doing it here means
+    //  sync will complete in one round (prevents messages failing to merge due to missed or out of order signer message)
+    const signerMessagesResult = await rpcClient.getAllSignerMessagesByUser(user);
+    if (signerMessagesResult.isErr()) {
+      return err(new ServerError('Failed to fetch signer messages'));
+    }
+    const results = await Promise.all(this.engine.mergeMessages([...signerMessagesResult._unsafeUnwrap()]));
```
sagar-a16z, line 200: avoid `_unsafeUnwrap` where possible. In this case you should be able to `signerMessagesResult.error` since it's already been determined that `signerMessagesResult` is not `isErr()` (line 197)

src/network/sync/syncEngine.ts
sanjayprabhu: Yes, this was intentional. This will be useful to help debug sync issues if they surface, but we don't need them during normal operation.

src/network/sync/syncEngine.ts
sanjayprabhu: If the hashes diverge, and two hubs have the exact same number of messages, then it's possible two hubs have seen two different messages. The random is a probabilistic tie breaker to determine which hub should sync with the other instead of always syncing both which could be more network/cpu intensive. 

Now that I think about it, I wrote this before switching to timer based sync instead of after every message. In the timer based version, it's less important to avoid this. Will change this to always return true.

src/network/sync/syncEngine.ts
sanjayprabhu: No, will paginate in an upcoming optimization focused PR.

src/network/p2p/protocol.ts
```diff
@@ -31,39 +31,34 @@ export type Content = IdRegistryContent | UserContent | ContactInfoContent;
...
 };
 
 /**
  * ContactInfoContent allows gossip nodes to share additional information about each other
  * over the gossip network.
  *
- * @peerId - The peerId of the node
  * @rpcAddress - The address at which this node is serving RPC requests. Unset if RPC is not offered.
+ * @excludedHashes - The excluded hashes of the sender's current trie snapshot
+ * @count - The number of messages under the root
```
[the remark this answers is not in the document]
sanjayprabhu: Yes, exactly. In addition to that, it would've added a lot of overhead to the normal gossip message.

src/network/sync/syncEngine.ts
sanjayprabhu: It's a typing issue, the RPCClient returns a JSONRPCError type, but we're expecting a FarcasterError type. Even if we returned a generic Error type, we'd still have to cast/construct a new object because the Result<> types don't match.

src/network/sync/syncEngine.ts
sanjayprabhu: Is that legal JS grammar? Doesn't seem like I can do `return if (<condition>);`. Am I missing something?

src/hub.ts
```diff
@@ -199,18 +190,39 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
...
+    if (!this.syncEngine.shouldSync(message.excludedHashes, message.count)) {
+      log.debug(`Upto date with peer, skipping sync`);
+      this.emit('syncComplete', false);
+      return;
+    }
+    await this.syncEngine.performSync(message.excludedHashes, rpcClient);
+    this.emit('syncComplete', true);
+  }
+
+  private async getRPCClientForPeer(peer: ContactInfoContent): Promise<RPCClient | undefined> {
```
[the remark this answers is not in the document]
sanjayprabhu: Just personally, I prefer to use `Result` only across component boundaries. Within the same class, it's a lot of additional boiler plate to check for the error/unwrap etc. But happy to do it if you feel strongly.

src/hub.ts
```diff
@@ -350,6 +299,21 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
   getCustodyEventByUser(fid: number): Promise<Result<IdRegistryEvent, FarcasterError>> {
     return this.engine.getCustodyEventByUser(fid);
   }
+  getSyncMetadataByPrefix(prefix: string): Promise<Result<NodeMetadata, FarcasterError>> {
```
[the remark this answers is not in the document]
sanjayprabhu: They are all async to keep it consistent with the other RPCHandler method signatures. I think it would be weird to have some of them be promises and some not. Will add a comment

src/hub.ts
```diff
@@ -199,18 +190,39 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
...
+      // Don't fire syncComplete
+      log.debug(`Already syncing, skipping sync`);
+      return;
+    }
+    if (!this.syncEngine.shouldSync(message.excludedHashes, message.count)) {
+      log.debug(`Upto date with peer, skipping sync`);
+      this.emit('syncComplete', false);
+      return;
+    }
+    await this.syncEngine.performSync(message.excludedHashes, rpcClient);
```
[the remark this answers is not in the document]
sanjayprabhu: Good point, fixed.

src/network/rpc/interfaces.ts
```diff
@@ -24,6 +28,9 @@ export interface RPCHandler {
   getAllFollowsByUser(fid: number): Promise<Set<Follow>>;
   getAllVerificationsByUser(fid: number): Promise<Set<Verification>>;
   getCustodyEventByUser(fid: number): Promise<Result<IdRegistryEvent, FarcasterError>>;
+  getSyncMetadataByPrefix(prefix: string): Promise<Result<NodeMetadata, FarcasterError>>;
+  getSyncIdsByPrefix(prefix: string): Promise<Result<string[], FarcasterError>>;
+  getMessagesByHashes(hashes: string[]): Promise<Message[]>;
```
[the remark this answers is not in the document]
sanjayprabhu: I did consider that, but there's no failure mode here. In case the hashes are not found, we'll just return an empty set. So, I think it's ok to just return a Message array. There is an argument to be made that we should keep it all consistent, but we've already broken that by returning non-result responses for the first few methods.  I could change everything to a Result, but would prefer to do it in a separate PR. Don't think it's a big deal for now.

src/network/sync/syncEngine.ts
sagar-a16z: I've also been unable to use this pattern @varunsrin.

src/hub.ts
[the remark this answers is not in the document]
sagar-a16z: I would `match` the error and since the `err` branch of the map is a noop anyway, it wouldn't be terrible. But no worries, don't feel too strongly. 

I was just trying to get `diffSynIfRequired` to not deal with the undefined arg.

src/hub.ts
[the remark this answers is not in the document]
sagar-a16z: It's totally fine for the RPCHandler APIs to be sync, that's just the server side Implementation and they can be sync. The actual RPC call is always going to be async anyway.

src/network/sync/syncEngine.ts
varunsrin: derp, its not. 

for some reason i got it into my head that it was, been doing too much solidity

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+            minDate: new Date(now - 1000 * 100), // Between 100-50 seconds ago
+            maxDate: new Date(now - 1000 * 50),
+          },
+        }
+      );
+      const res = await Promise.all(serverEngine.mergeMessages(messages));
+      expect(res.every((r) => r.isOk())).toBeTruthy();
+
+      const serverSnapshot = serverSyncEngine.snapshot;
+      expect(serverSnapshot.numMessages).toBeGreaterThanOrEqual(clientSyncEngine.snapshot.numMessages);
```
varunsrin, line 157: shouldn't this always be greater than by exactly 10?

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      expect(clientSyncEngine.shouldSync(serverSnapshot.excludedHashes, serverSnapshot.numMessages)).toBeTruthy();
+
+      const divergencePrefix = clientSyncEngine.trie.getDivergencePrefix(
+        clientSyncEngine.snapshot.prefix,
+        serverSnapshot.excludedHashes
+      );
+      // The point of divergence should be some distance from the root, and must have few messages than the root
```
varunsrin, line 164: nit: `fewer` instead of `few`

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      expect(divergencePrefix.length).toBeGreaterThanOrEqual(3);
+      const serverDivergedNode = serverSyncEngine.trie.getNodeMetadata(divergencePrefix);
```
varunsrin, line 166: nit: should this be `serverDivergedNodeMetadata` or just `nodeMetadata` for clarity?

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+}
+
+let server: RPCServer;
+let client: RPCClient;
+
+const NUM_USERS = 5;
+const TEST_TIMEOUT = 2 * 60 * 1000; // 2 min timeout
+
+describe('differentialSync', () => {
```
varunsrin, line 68: should we also add a test helper that walks every branch and leaf an asserts that they are exactly equal? this would be useful to assert in every case here

src/hub.ts
```diff
@@ -350,6 +299,21 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
   getCustodyEventByUser(fid: number): Promise<Result<IdRegistryEvent, FarcasterError>> {
     return this.engine.getCustodyEventByUser(fid);
   }
+  getSyncMetadataByPrefix(prefix: string): Promise<Result<NodeMetadata, FarcasterError>> {
+    const nodeMetadata = this.syncEngine.getNodeMetadata(prefix);
+    if (nodeMetadata) {
+      return Promise.resolve(ok(nodeMetadata));
+    } else {
+      return Promise.resolve(err(new FarcasterError('no metadata found')));
```
varunsrin, line 307: this line isn't covered by tests, if tests are unimportant here please add a comment noting why

src/network/sync/merkleTrie.ts
```diff
@@ -1,6 +1,22 @@
 import { createHash } from 'crypto';
 import { ID_LENGTH, SyncId } from '~/network/sync/syncId';
 
+// Represent a snapshot of the trie at a particular timestamp. Used to compare two hubs and determine
```
varunsrin, line 4: please use TSDoc comments for exported methods and types, since they make generating docs easier: https://github.com/farcasterxyz/hub/blob/main/CONTRIBUTING.md#32-writing-docs

e.g.

`/* A snapshot of the trie at a timestamp which can be used to determine if hubs are in sync */`

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  let userInfos: UserInfo[];
+
+  beforeEach(async () => {
+    await serverEngine._reset();
+    await clientEngine._reset();
+    serverSyncEngine = new SyncEngine(serverEngine);
```
varunsrin, line 74: server and client is a bit confusing nomenclature here - i assume this means the receiving hub vs calling hub? 

i think it would be easier to just call them hub a and hub b

varunsrin (commented):
left another pass of comments, will do a final one this afternoon

src/network/sync/merkleTrie.ts
```diff
@@ -22,6 +38,37 @@ class MerkleTrie {
     return this._root.get(id.toString());
   }
 
+  public getSnapshot(prefix: string): TrieSnapshot {
```
varunsrin, line 41: would add a comment here explaining what the snapshot is

src/network/sync/merkleTrie.ts
```diff
@@ -1,6 +1,22 @@
 import { createHash } from 'crypto';
 import { ID_LENGTH, SyncId } from '~/network/sync/syncId';
 
+// Represent a snapshot of the trie at a particular timestamp. Used to compare two hubs and determine
+// if they are in sync.
+export type TrieSnapshot = {
+  prefix: string;
```
varunsrin, line 7: it's very hard to figure out what a prefix is from reading this file. there are a few problems: 

1. it's just a regular string, so the type encodes no information
2. it's the parameter name in a bunch of methods where it is simply passed to other methods

i would consider either renaming it to something that makes it more apparent or if that is not easy then adding method docs that explain what the prefix parameter is and how it is being used

src/network/sync/merkleTrie.ts
```diff
@@ -22,6 +38,37 @@ class MerkleTrie {
     return this._root.get(id.toString());
   }
 
+  public getSnapshot(prefix: string): TrieSnapshot {
+    return this._root.getSnapshot(prefix);
+  }
+
+  public getDivergencePrefix(prefix: string, excludedHashes: string[]): string {
```
varunsrin, line 45: add comments explaining why a divergence prefix might be needed and a quick description of how it should be calculated

src/network/sync/merkleTrie.ts
```diff
@@ -100,6 +147,39 @@ class TrieNode {
...
+    const innerSnapshot = this._children.get(char)?.getSnapshot(prefix, current_index + 1);
+    const excludedHash = this._excludedHash(char);
+    return {
+      prefix: innerSnapshot?.prefix || prefix.slice(0, current_index + 1),
+      excludedHashes: [excludedHash.hash, ...(innerSnapshot?.excludedHashes || [])],
+      numMessages: excludedHash.items + (innerSnapshot?.numMessages || 0),
+    };
+  }
+
+  private _excludedHash(char: string): { items: number; hash: string } {
```
varunsrin, line 170: mixing private and public methods makes it take longer to grok the class's interface and its purpose. 

i would recommend moving all private methods below all public methods

src/network/sync/merkleTrie.ts
```diff
@@ -100,6 +147,39 @@ class TrieNode {
     return this._children.get(char)?.get(key.slice(1));
   }
 
+  public getSnapshot(prefix: string, current_index = 0): TrieSnapshot {
```
varunsrin, line 150: this could use some documentation for the method and parameters

src/network/sync/merkleTrie.ts
```diff
@@ -22,6 +38,37 @@ class MerkleTrie {
...
+    const ourExcludedHashes = this.getSnapshot(prefix).excludedHashes;
+    for (let i = 0; i < prefix.length; i++) {
+      if (ourExcludedHashes[i] !== excludedHashes[i]) {
+        return prefix.slice(0, i);
+      }
+    }
+    return prefix;
+  }
+
+  public getNodeMetadata(prefix: string): NodeMetadata | undefined {
```
varunsrin, line 55: nit: we call them TrieNode's in some places and we specify them as Node's here. would recommend being consistent otherwise it may not be obvious that NodeMetadata is the Metadata of a TrieNode

src/network/sync/merkleTrie.ts
```diff
@@ -100,6 +147,39 @@ class TrieNode {
...
+    if (current_index === prefix.length - 1) {
+      const excludedHash = this._excludedHash(char);
+      return {
+        prefix: prefix,
+        excludedHashes: [excludedHash.hash],
+        numMessages: excludedHash.items,
+      };
+    }
+
+    const innerSnapshot = this._children.get(char)?.getSnapshot(prefix, current_index + 1);
```
varunsrin, line 161: can we break the TrieNode and MerkleTrie class into separate files?

It gets messy when reviewing large PR's in Github cause its not obvious that we've gone from one class's methods to the next.

varunsrin (commented):
added more comments on the merkle trie class

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+        expect(mergedHashes).toContain(message.hash);
+      }
+      // In the worst case, the maximum number of messages merged should be the number of items under the divergence prefix
+      expect(mergedHashes.length).toBeLessThanOrEqual(serverDivergedNode?.numMessages || -1);
+      // Ideally the merged hashes length should be exactly equal to the new messages,
+      // but, depending on the location of divergence, it's possible that some existing messages are syncd
+      // So just ensure it's below some reasonable threshold
+      expect(mergedHashes.length).toBeLessThanOrEqual(100);
+    },
+    TEST_TIMEOUT
```
varunsrin, line 183: test case:  assume two Hubs start with identical messages and then one merges a SignerRemove which also cascades a bunch of deletes across other sets.

varunsrin: test case: assume two Hubs start with identical messages and one merges any kind of Remove message which also deletes an add message elsewhere in the set. what if we discover the exclusion hash delta where the add message is missing before we discover the delta caused by the added remove message?

varunsrin (changes requested):

src/network/sync/syncEngine.ts
sanjayprabhu: No, happy to add it. It seemed like we were using HTTP like error codes, so wasn't sure what the convention should be. Currently we're doing this `new BadRequestError('validateMessage: unknown user')` in `validateMessage`. What error code should I use for a `UserNotFoundError`?

I'll do it in the followup PR though, would like to merge this soon.

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      Follows: 50,
+      Reactions: 0,
+    });
+  }, TEST_TIMEOUT);
+
+  afterEach(async () => {
+    await server.stop();
+  });
+
+  const ensureEnginesEqual = async () => {
```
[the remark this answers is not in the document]
sanjayprabhu: I tried this, but felt it was more unwieldily to call since we have to pass in 4 different engines each time. So, prefer to leave it as is.

src/test/e2e/differentialSync.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  test(
+    'syncs data to a completely empty client',
+    async () => {
+      // sanity test that the server has something in it and the client, nothing
+      const users = await serverEngine.getUsers();
+      expect(users.size).toEqual(NUM_USERS);
+      expect(await clientEngine.getUsers()).toEqual(new Set());
+
+      const snapshot = serverSyncEngine.snapshot;
+      await clientSyncEngine.performSync(snapshot.excludedHashes, client);
```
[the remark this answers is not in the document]
sanjayprabhu: It's covered by SyncEngine unit tests, but added it here as well.

src/test/e2e/differentialSync.test.ts
sanjayprabhu: In `ensureEngineEqual`, we compare the merkle roots which should be sufficient. Added a comment for clarity.

src/network/sync/merkleTrie.ts
sanjayprabhu: Added an explanation in the documentation for the class

src/network/sync/merkleTrie.ts
sanjayprabhu: Will fix in followup PR.

src/network/sync/merkleTrie.test.ts
```diff
@@ -1,8 +1,22 @@
 import { Factories } from '~/test/factories';
 import { MerkleTrie } from '~/network/sync/merkleTrie';
 import { SyncId } from '~/network/sync/syncId';
+import { createHash } from 'crypto';
```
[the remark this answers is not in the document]
sanjayprabhu: I think of the MerkleTrie as a thin wrapper around the root TrieNode with some helper methods (it's one of the reasons I put them in the same file). No one other than the MerkleTrie should even be interacting with a TrieNode directly. 

That's why I kept the same behaviour for tests, all these methods have tests at the merkle trie level (which directly delegates to the TrieNode). I could move them to the TrieNode, but we'd have to test them again here to make sure the delegation is correct. So prefer to keep it as is, unless there's additional logic in the MerkleTrie.

src/network/sync/merkleTrie.ts
```diff
@@ -22,6 +38,37 @@ class MerkleTrie {
...
   }
 
+  public getSnapshot(prefix: string): TrieSnapshot {
+    return this._root.getSnapshot(prefix);
+  }
+
+  public getDivergencePrefix(prefix: string, excludedHashes: string[]): string {
+    const ourExcludedHashes = this.getSnapshot(prefix).excludedHashes;
+    for (let i = 0; i < prefix.length; i++) {
+      if (ourExcludedHashes[i] !== excludedHashes[i]) {
```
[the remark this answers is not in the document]
sanjayprabhu: The length mismatch case is tested when we pass in an empty array, but also added a test case for the when the provided hashes are longer.

Accessing an array index that's out of bounds returns undefined, which is why it doesn't crash.

src/test/e2e/differentialSync.test.ts
sanjayprabhu: Good point about the removes, the test will fail because the sync engine currently doesn't remove message hashes from the trie. I'll address in a followup PR.

sanjayprabhu:
@varunsrin Addressed most of your comments. The rest, in particular, the message remove tests and custom error class for user not found, I'll address in a followup. Would like to get this merged (unless there are other blockers?) and not have such a large branch hanging around too long.

varunsrin (approved):

src/hub.ts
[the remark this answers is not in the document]
sanjayprabhu: Fixed in https://github.com/farcasterxyz/hub/pull/191

src/test/e2e/differentialSync.test.ts
sanjayprabhu: Fixed in https://github.com/farcasterxyz/hub/pull/189

PR #186, opened 2022-10-28
chore: replace faker with faker-js

varunsrin:
## Motivation

`faker` is a deprecated npm module that may be also responsible for other issues with deploys, so we're replacing with `faker-js` which is a more maintained implementation

## Change Summary

- `faker` was swapped in place for `faker-js` which mostly just worked with some API changes. 
- `signedAt` for some messages had to be explicitly set since `faker.date.recent().getTime()` works different from `faker.time.recent()`

## Merge Checklist

### DIFF
--- a/src/storage/engine/engine.follow.test.ts
+++ b/src/storage/engine/engine.follow.test.ts
@@ -1,5 +1,5 @@
 import Engine from '~/storage/engine';
-import Faker from 'faker';
+import { faker } from '@faker-js/faker';
 import { Factories } from '~/test/factories';
 import {
   Ed25519Signer,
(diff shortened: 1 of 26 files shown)


src/storage/engine/engine.follow.test.ts
```diff
@@ -49,7 +49,7 @@ describe('mergeFollow', () => {
     transientParams = { transient: { signer: aliceSigner } };
     follow = await Factories.FollowAdd.create({ data: { fid: aliceFid } }, transientParams);
     unfollow = await Factories.FollowRemove.create(
-      { data: { fid: aliceFid, body: { targetUri: follow.data.body.targetUri } } },
+      { data: { fid: aliceFid, body: { targetUri: follow.data.body.targetUri }, signedAt: follow.data.signedAt + 1 } },
```
varunsrin, line 52: @pfletcherhill  FYI / NAR 

some of our tests relied on a property of the old faker library that guaranteed that a second calls to `faker.time.recent()` would always return a timestamp >= the first call. 

with the new `faker.date.recent().getTime()` the calls to these factories need to explicitly cases where the timestamp must be greater than.

PR #189, opened 2022-10-31
feat: Handle message deletion in diff sync

sanjayprabhu:
## Motivation

Diff sync must be able to handle deleted messages (e.g. from a signer remove) and converge to the same state. 

## Change Summary

Sync engine listens for message deletions from the storage engine an deletes the item from the trie. Added tests to make sure hubs will converge to same state when they have deleted messages.

Also includes a other follow ups from review comments in https://github.com/farcasterxyz/hub/pull/179

## Merge Checklist

### DIFF
--- a/src/network/sync/merkleTrie.test.ts
+++ b/src/network/sync/merkleTrie.test.ts
@@ -97,6 +97,22 @@ describe('MerkleTrie', () => {
       expect(trie.get(syncId)).toBeFalsy();
     });
 
+    test('deleting an item that does not exist does not change the trie', async () => {
+      const message = await Factories.CastShort.create();
+      const syncId = new SyncId(message);
+
+      const trie = new MerkleTrie();
+      trie.insert(syncId);
+
+      const rootHashBeforeDelete = trie.rootHash;
+      const message2 = await Factories.CastShort.create();
+      const syncId2 = new SyncId(message2);
+      trie.delete(syncId2);
+
+      expect(trie.rootHash).toEqual(rootHashBeforeDelete);
+      expect(trie.items).toEqual(1);
+    });
+
     test('delete is an exact inverse of insert', async () => {
       const message1 = await Factories.CastShort.create();
       const syncId1 = new SyncId(message1);
--- a/src/network/sync/merkleTrie.test.ts
+++ b/src/network/sync/merkleTrie.test.ts
@@ -167,7 +167,7 @@ describe('MerkleTrie', () => {
       const trie = new MerkleTrie();
       trie.insert(syncId);
 
-      expect(trie.getNodeMetadata('166518234')).toBeUndefined();
+      expect(trie.getTrieNodeMetadata('166518234')).toBeUndefined();
     });
 
     test('returns the root metadata if the prefix is empty', async () => {
@@ -175,7 +175,7 @@ describe('MerkleTrie', () => {
       const trie = new MerkleTrie();
       trie.insert(syncId);
 
-      const nodeMetadata = trie.getNodeMetadata('');
+      const nodeMetadata = trie.getTrieNodeMetadata('');
       expect(nodeMetadata).toBeDefined();
       expect(nodeMetadata?.numMessages).toEqual(1);
       expect(nodeMetadata?.prefix).toEqual('');
@@ -185,7 +185,7 @@ describe('MerkleTrie', () => {
 
     test('returns the correct metadata if prefix is present', async () => {
       const trie = await trieWithMessages([1665182332, 1665182343]);
-      const nodeMetadata = trie.getNodeMetadata('16651823');
+      const nodeMetadata = trie.getTrieNodeMetadata('16651823');
 
       expect(nodeMetadata).toBeDefined();
       expect(nodeMetadata?.numMessages).toEqual(2);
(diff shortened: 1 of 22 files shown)


src/network/sync/trieNode.ts
```diff
@@ -67,6 +66,31 @@ class TrieNode {
     return false;
   }
 
+  public delete(key: string, current_index = 0): boolean {
```
varunsrin, line 69: nit: consider adding comments to explain the way the recursive delete and cleanup works

src/network/sync/trieNode.ts
```diff
@@ -67,6 +66,31 @@ class TrieNode {
...
+    if (this.isLeaf) {
+      return true;
+    }
+
+    const char = key[current_index];
+    if (!this._children.has(char)) {
+      return false;
+    }
+
+    const success = this._children.get(char)?.delete(key, current_index + 1);
```
varunsrin, line 79: as an aside, how does the trie handle concurrent inserts and deletions if they happen on different logical threads?

src/storage/engine/index.ts
```diff
@@ -265,6 +274,8 @@ class Engine extends TypedEmitter<EngineEvents> {
 
   private async revokeSigner(fid: number, signer: string): Promise<Result<void, FarcasterError>> {
     try {
+      // TODO: Pause syncing while revoking because it can lead to a lot of thrashing
```
varunsrin, line 277: when a signer is removed, does sync need guarantees from storage about the ordering of the removals? and in an ideal world, does it apply them to the trie in series or in parallel?

src/storage/engine/engine.revoke.test.ts
```diff
@@ -100,5 +101,18 @@ describe('revokeSigner', () => {
...
     });
+
+    test('fires events for all the dropped messages', async () => {
+      const removedMessages: Message[] = [];
+      engine.onDBEvent('messageDeleted', (message) => removedMessages.push(message));
+
+      const res = await engine._revokeSigner(aliceFid, aliceSigner.signerKey);
+      expect(res.isOk()).toBeTruthy();
+
+      expect(removedMessages).toContainEqual(aliceCast);
```
varunsrin, line 112: see my question below about ordering, if we do not care about it this approach is fine otherwise we may want to be more explicit about the returned array

src/storage/db/message.ts
```diff
@@ -1,6 +1,15 @@
 import RocksDB, { Transaction } from '~/storage/db/rocksdb';
 import { Message, MessageType } from '~/types';
 import { isMessage } from '~/types/typeguards';
+import { TypedEmitter } from 'tiny-typed-emitter';
+
+export type MessageDBEvents = {
+  /**
+   * messageDeleted is emitted whenever the DB attempts to delete a message.
+   * It is possible for a transaction to fail after this event is fired.
```
varunsrin, line 9: what happens if the tx fails? does the trie get out of sync with the stores?

src/network/sync/trieNode.ts
sanjayprabhu: There are no async methods on MerkleTrie or TrieNode classes, so with the JS concurrency model, it's not possible for them to be interrupted during an insert/delete. Therefore the trie operations are always "single-threaded".

I'll add a comment about the concurrency assumptions to the class.

src/storage/engine/index.ts
sanjayprabhu: No, the order does not matter because the trie will always end up in the same state regardless of the order of operations. Ideally, we collate all the removals per transaction and immediately apply them as soon as the transaction commits (in sequnce because the trie doesn't support parallel operations).

src/storage/engine/engine.revoke.test.ts
sanjayprabhu: Order does not matter. I'll add a comment

src/storage/db/message.ts
sanjayprabhu: Yes, but this is fine because the sync engine will self repair (by triggering a sync). I chose this approach for two reasons: a) much simpler to implement, b) With the more complex approach of buffering the deletes and only emitting on tx success, if there's a mistake/problem and we miss a delete, the trie will never get back in sync until the hub is restarted (because the sync engine only ever add messages, it never deletes messages directly).


There's a comment about this here https://github.com/sanjayprabhu/hub/blob/sanjayr/merkle-sync-followups/src/network/sync/syncEngine.ts#L37-L41

src/storage/engine/index.ts
```diff
@@ -234,6 +234,15 @@ class Engine extends TypedEmitter<EngineEvents> {
     return ResultAsync.fromPromise(this._signerSet.getCustodyEvent(fid), (e) => e as FarcasterError);
   }
 
+  onDBEvent<E extends keyof MessageDBEvents>(event: E, callback: MessageDBEvents[E]) {
```
sanjayprabhu, line 237: @pfletcherhill This is the main engine/set change in this PR. We will need same functionality with the new stores.

src/network/sync/merkleTrie.test.ts
```diff
@@ -78,6 +80,76 @@ describe('MerkleTrie', () => {
...
+
+      trie.delete(syncId2);
+      expect(trie.rootHash).toEqual(rootHashBeforeDelete);
+    });
+
+    test('trie with a deleted item is the same as a trie with the item never added', async () => {
+      const message1 = await Factories.CastShort.create();
+      const syncId1 = new SyncId(message1);
+      const message2 = await Factories.CastShort.create();
+      const syncId2 = new SyncId(message2);
```
varunsrin, line 135: nit: you could make a syncid factory that generates these values, which will make tests faster since they don't need to sign casts

sanjayprabhu: Good idea. Will fix in followup.

PR #190, opened 2022-10-31
feat: add reaction set

varunsrin:
## Motivation

Adding a ReactionSet to store reaction messages as part of the migration to using RocksDB and flatbuffers 

## Change Summary

See above

## Merge Checklist


src/storage/sets/reactionSet.ts
```diff
@@ -28,7 +28,7 @@ class ReactionSet {
   }
 
   async getReactionsByUser(fid: number): Promise<Set<ReactionAdd>> {
-    const reactions = await this._db.getReactionAddsByUser(fid);
+    const reactions = await this._db.getReactionAddsByFid(fid);
```
varunsrin, line 31: TODO: accidentally merged a find + replace, need to undo

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+  private _db: RocksDB;
+
+  constructor(db: RocksDB) {
+    this._db = db;
+  }
+
+  // DISCUSS: the terms UserPrefix and RootPrefix.User were confusing for quite a while. What we
+  // are really saying is this is a prefix that ensures that it is a "Message Storing" key.
+
+  // DISCUSS: the name Set is still slightly confusing because it has 3 overloaded meanings
```
pfletcherhill, line 41: I think I still prefer it over collection but open to other ideas

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+  static reactionRemovesKey(fid: Uint8Array, type?: ReactionType, targetId?: Uint8Array): Buffer {
+    return Buffer.concat([
+      MessageModel.userKey(fid), // --------------------------- fid prefix, 33 bytes
+      Buffer.from([UserPrefix.ReactionRemoves]), // ----------- reaction_adds key, 1 byte
+      targetId ? Buffer.from(targetId) : new Uint8Array(), //-- target id, variable bytes
+      type ? Buffer.from([type]) : new Uint8Array(), //-------- type, 1 byte
+    ]);
+  }
+
+  // DISCUSS: should reactions index by target key and type
```
pfletcherhill, line 80: Yeah, I think you should be able to return all reactions for a target and also all reactions of a certain type for a target. So the parameters should be split out.

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+   * @param type type of reaction created
+   * @param targetId id of the object being reacted to
+   *
+   * @returns RocksDB key of the form user_prefix:reaction_type:target_id which is at least 35 bytes
+   */
+  static reactionAddsKey(fid: Uint8Array, type?: ReactionType, targetId?: Uint8Array): Buffer {
+    return Buffer.concat([
+      MessageModel.userKey(fid), // --------------------------- fid prefix, 33 bytes
+      Buffer.from([UserPrefix.ReactionAdds]), // -------------- reaction_adds key, 1 byte
+      targetId ? Buffer.from(targetId) : new Uint8Array(), //-- target id, variable bytes
```
pfletcherhill, line 57: I think targetId and type should be reversed here, so reaction type comes first. If reaction type comes first, then I can return all reactions of a particular type per user easily.

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+      bytes.set(hash, 1 + FID_BYTES + FID_BYTES);
+    }
+    return Buffer.from(bytes);
+  }
+
+  /* -------------------------------------------------------------------------- */
+  /*                              Instance Methods                              */
+  /* -------------------------------------------------------------------------- */
+
+  // DISCUSS: ErrorHandling for public methods when making calls to the database. Should we treat
```
pfletcherhill, line 99: Meaning is it okay to throw NotFoundError here when the record is missing vs returning `Result<ReactionAddModel, NotFoundError>`? I personally prefer this pattern vs having to destructure Result values everywhere but open to other arguments.

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+   * @returns the ReactionRemove message if it exists, undefined otherwise
+   */
+  async getReactionRemove(fid: Uint8Array, type: ReactionType, castId: CastID): Promise<ReactionRemoveModel> {
+    const reactionRemovesKey = ReactionSet.reactionRemovesKey(fid, type, this.targetKeyForCastId(castId));
+    const reactionMessageKey = await this._db.get(reactionRemovesKey);
+
+    return MessageModel.get<ReactionRemoveModel>(this._db, fid, UserPrefix.ReactionMessage, reactionMessageKey);
+  }
+
+  /** Finds all ReactionAdd Messages by iterating through the prefixes */
```
pfletcherhill, line 133: I'm not sure this comment is quite right. We're iterating through all keys with a common prefix, not multiple prefixes.

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+  /*                               Private Methods                              */
+  /* -------------------------------------------------------------------------- */
+
+  private async mergeAdd(message: ReactionAddModel): Promise<void> {
+    const castId = message.body().cast();
+
+    if (castId) {
+      const targetKey = this.targetKeyForCastId(castId);
+
+      // DISCUSS: tsx is a hard to search for term and transactions are often abbreviated to
```
pfletcherhill, line 189: Sure, happy to change to `txn` or just `transaction`

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+    aTimestampHash: Uint8Array,
+    bType: MessageType,
+    bTimestampHash: Uint8Array
+  ): number {
+    const timestampHashOrder = bytesCompare(aTimestampHash, bTimestampHash);
+    if (timestampHashOrder !== 0) {
+      return timestampHashOrder;
+    }
+
+    // TODO: Changing these types to FollowRemove and FollowAdd did not break tests
```
pfletcherhill, line 235: I don't think other sets have sufficient tests covering these lines yet (i.e. where the timestampHash is the same but the types are different)

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+    message: ReactionAddModel | ReactionRemoveModel
+  ): Promise<Transaction | undefined> {
+    // Checks if there is a remove timestamp hash for this reaction
+    const reactionRemoveTimestampHash = await ResultAsync.fromPromise(
+      this._db.get(ReactionSet.reactionRemovesKey(message.fid(), message.body().type(), targetKey)),
+      () => undefined
+    );
+
+    // DISCUSS: we have changed the ordering of the three rules from LWW-RW-HW to LWW-HW-RW
+    // DISCUSS: is LWW-HW robust enough to be used independently? Does RW even add anything?
```
pfletcherhill, line 257: Was wondering about this too. I don't have an answer yet.

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+ * reactionTargetPrefix:targetCastTimestampHash:reactionType -> fid:timestampHash (Target Index)
+ */
+class ReactionSet {
+  private _db: RocksDB;
+
+  constructor(db: RocksDB) {
+    this._db = db;
+  }
+
+  // DISCUSS: the terms UserPrefix and RootPrefix.User were confusing for quite a while. What we
```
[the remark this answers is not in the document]
varunsrin: What do you think about calling them: 

`StoragePrefix` - differentiates between keys that store messages and keys that index message keys (i.e. what is being stored) and can be extended in the future

`MessagePrefix` - differentiates between the types of messages inside of a UserStore

src/storage/sets/flatbuffers/reactionSet.ts
varunsrin: good catch, should be `with a prefix` not `through the prefixes`

src/storage/sets/flatbuffers/reactionSet.ts
varunsrin: txn works for me, i'll change this here and we can refactor elsewhere

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  /* -------------------------------------------------------------------------- */
+  /*                               Private Methods                              */
+  /* -------------------------------------------------------------------------- */
+
+  private async mergeAdd(message: ReactionAddModel): Promise<void> {
+    const castId = message.body().cast();
+
+    if (castId) {
```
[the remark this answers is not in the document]
varunsrin: sure, ill throw

src/storage/sets/flatbuffers/reactionSet.ts
varunsrin: done

src/storage/sets/flatbuffers/reactionSet.ts
varunsrin: done

src/storage/sets/flatbuffers/reactionSet.ts
varunsrin: discussed offline, need more investigation

src/storage/sets/flatbuffers/reactionSet.ts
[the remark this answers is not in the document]
pfletcherhill: That's not really the distinction. All messages are stored at `MessageModel.primaryKey` (<RootPrefix.User, fid, MessagePrefix.CastMessage | MessagePrefix.FollowMessage | ..., timestampHash>). Everything else is an index (i.e. the value is either true or the timestampHash).

The difference between the RootPrefix and UserPrefix is really a matter of depth—RootPrefix values are always the first byte of the key, and UserPrefix values are always a byte following the fid. So we could use PrimaryPrefixByte and SecondaryPrefixByte maybe?

src/storage/sets/flatbuffers/reactionSet.ts
varunsrin: ok, we should discuss whether remove wins still makes sense in the context of these changes

src/storage/flatbuffers/model.ts
```diff
@@ -17,6 +18,10 @@ import { VerificationRemoveBody } from '~/utils/generated/farcaster/verification
 /** Used when index keys are sufficiently descriptive */
 export const TRUE_VALUE = Buffer.from([1]);
 export const FID_BYTES = 32;
+export const TARGET_KEY_BYTES = 40; // 32-byte fid, 4-byte timestamp, 4-byte hash
+
+// TODO: Add docs explaining the purpose of this class
+// TODO: naming of the class + file name does not match TS conventions
```
varunsrin, line 24: self: address this TODO

src/storage/flatbuffers/typeguards.ts
```diff
@@ -1,11 +1,14 @@
 import MessageModel from '~/storage/flatbuffers/model';
+// TODO: Import all and apply the types model here to avoid bloated imports
```
varunsrin, line 2: self: address this TODO

src/storage/flatbuffers/types.ts
```diff
@@ -1,21 +1,26 @@
...
   CastRemoveBody,
   FarcasterNetwork,
   FollowBody,
   SignerBody,
-  VerificationAddEthAddressBody,
-  VerificationRemoveBody,
+  ReactionBody,
 } from '~/utils/generated/message_generated';
 
+// TODO: Document the purpose of root prefix
```
varunsrin, line 13: self: address this TODO and the one below

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,437 @@
...
+    // Checks if there is an add timestamp hash for this reaction
+    const reactionAddTimestampHash = await ResultAsync.fromPromise(
+      this._db.get(ReactionSet.reactionAddsKey(message.fid(), message.body().type(), targetKey)),
+      () => undefined
+    );
+
+    if (reactionAddTimestampHash.isOk()) {
+      if (
+        this.reactionMessageCompare(
+          // TODO: this was set to FollowAdd and did not break tests
```
varunsrin, line 336: self: address this TODO

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,437 @@
...
+    );
+
+    return txn;
+  }
+
+  private deleteReactionRemoveTransaction(txn: Transaction, message: ReactionRemoveModel): Transaction {
+    // Delete the reactionRemove key from reactionRemoves Set
+    const targetKey = this.targetKeyForMessage(message);
+
+    // TODO: Handling the type this way is unwieldy
```
varunsrin, line 416: self: address this TODO

src/storage/sets/flatbuffers/reactionSet.ts
```diff
@@ -0,0 +1,437 @@
...
+
+    return Buffer.concat([
+      MessageModel.userKey(fid), // --------------------------- fid prefix, 33 bytes
+      Buffer.from([UserPrefix.ReactionRemoves]), // ----------- reaction_adds key, 1 byte
+      type ? Buffer.from([type]) : new Uint8Array(), //-------- type, 1 byte
+      targetKey ? Buffer.from(targetKey) : new Uint8Array(), //-- target id, variable bytes
+    ]);
+  }
+
+  // DISCUSS: Timestamp Hash vs. TsHash vs TimestampId vs TID
```
varunsrin, line 88: @pfletcherhill thoughts on the term `tsHash` to represent the timestampHash construct in a more terse way?

e.g. `reactionTsHash` instead of `reactionTimestampHash`

varunsrin: decision based on offline discussion: use timestamp-hash in text and tsHash or *TsHash in code

src/storage/sets/flatbuffers/reactionSet.ts
varunsrin: discussed offline, decision: use Store instead of Set for the higher level construct

varunsrin:
Changes that will be handled in followon PR's

- Changes to the protocol spec will be handled in a batch update with other v2 changes
- Reaction Merge logic is currently LWW(ts)-LWW(hash)-RW, should be changed to LWW(ts)-RW-LWW(hash)
- Reaction merge is still missing some test coverage for merge logic

PR #216, opened 2022-11-05
fix(VerificationStore): increase test coverage, change conflict resolution, add docs

varunsrin:
## Change Summary

- fix: sorting of messages should be performed by timestamp, then type, then hash similar to other stores
- added documentation for classes and methods
- increase test coverage for branches in merge

## Merge Checklist

### DIFF
--- a/src/storage/sets/flatbuffers/verificationStore.ts
+++ b/src/storage/sets/flatbuffers/verificationStore.ts
@@ -14,16 +14,17 @@ class VerificationStore {
     this._db = db;
   }
 
-  /** RocksDB key of the form <user prefix (1 byte), fid (32 bytes), verification removes key (1 byte), address (variable bytes)> */
-  static verificationRemovesKey(fid: Uint8Array, address?: Uint8Array): Buffer {
-    return Buffer.concat([
-      MessageModel.userKey(fid),
-      Buffer.from([UserPostfix.VerificationRemoves]),
-      address ? Buffer.from(address) : new Uint8Array(),
-    ]);
-  }
+  // RESEARCH + DOCUMENT: how can the adds set be extended to other verification types?
 
-  /** RocksDB key of the form <user prefix (1 byte), fid (32 bytes), verification adds key (1 byte), address (variable bytes)> */
+  /**
+   * Generates a unique key used to store a VerificationAdds message key in the VerificationsAdds
+   * set index
+   *
+   * @param fid farcaster id of the user who created the verification
+   * @param address Ethereum address being verified
+   *
+   * @returns RocksDB key of the form <RootPrefix>:<fid>:<UserPostfix>:<address?>
+   */
   static verificationAddsKey(fid: Uint8Array, address?: Uint8Array): Buffer {
     return Buffer.concat([
       MessageModel.userKey(fid),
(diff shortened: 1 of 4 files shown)


src/storage/sets/flatbuffers/verificationStore.ts
```diff
@@ -7,32 +7,83 @@ import { isVerificationAddEthAddress, isVerificationRemove } from '~/storage/fla
...
 class VerificationStore {
   private _db: RocksDB;
 
   constructor(db: RocksDB) {
     this._db = db;
   }
 
-  /** RocksDB key of the form <user prefix (1 byte), fid (32 bytes), verification removes key (1 byte), address (variable bytes)> */
-  static verificationRemovesKey(fid: Uint8Array, address?: Uint8Array): Buffer {
+  // DISCUSS: A verification type needs a unique key to resolve conflicts correctly. The
```
varunsrin, line 33: cc @pfletcherhill @Raynos for discussion

pfletcherhill: I think this is an implementation detail we can address when we support multiple types of verification add messages. I'd be happy with the eventual approach of both kinds of verification add models having a method of the same name that returns the address or address equivalent.

src/storage/sets/flatbuffers/verificationStore.ts
```diff
@@ -7,32 +7,83 @@ import { isVerificationAddEthAddress, isVerificationRemove } from '~/storage/fla
...
+  // VerificationAddEthereumAddress uses the Ethereum address as this identifier. But it is quite
+  // possible that we may add a verification type that has a unique property other than address.
+  // One idea to make this easier to extend is to have a generic "identifier" property on the model
+  // which is always set directly to the key and all methods calling any subtype of Verification
+  // can safely assume it exists. Another option is building a method on top of the model that
+  // can return the field, which allows us to combine por manipulate fields. Either approach offers
+  // more flexibility than the current approach which has methods strictly tied to "Address".
+
+  // DISCUSS: a similar problem to the one above exists for ordinality, where we depend on
```
pfletcherhill, line 42: We don't depend on block hash to resolve conflicts. That's only there as extra info for clients (i.e. the block the address signed the claim), but the store only cares about message timestamp, type, and hash.

src/storage/sets/flatbuffers/verificationStore.ts
pfletcherhill: Also, regardless of how we implement it down the line, it shouldn't change the VerificationAddEthAddress message format

src/storage/sets/flatbuffers/verificationStore.ts
```diff
@@ -7,32 +7,83 @@ import { isVerificationAddEthAddress, isVerificationRemove } from '~/storage/fla
...
+  /**
+   * Generates a unique key used to store a VerificationAdds message key in the VerificationsAdds
+   * set index
+   *
+   * @param fid farcaster id of the user who created the verification
+   * @param address Ethereum address being verified
+   *
+   * @returns RocksDB key of the form <RootPrefix>:<fid>:<UserPostfix>:<address?>
+   */
+  static verificationAddsKey(fid: Uint8Array, address?: Uint8Array): Buffer {
```
Raynos, line 54: Show the address be optional here ? I dont understand what you can verify if the address is `undefined` / `null` ?

src/storage/sets/flatbuffers/verificationStore.ts
Raynos: For extensability you can set the kind of operation being verified `<RootPrefix>:<fid>:<UserPostfix>:<VerificationType>:<address?>`

The `VerificationType` can be hardcoded to `Buffer.from('ethaddress')` or to save bytes it can be an enum like `0` which has a label `EthAddress`.

Since the comment / pull request is in the context of building database keys I'm not sure what other interface changes would make sense higher up in the call stack.

src/storage/sets/flatbuffers/verificationStore.ts
varunsrin: think of `verificationAddsKey` as a prefix generator. Sometimes you want to generate a prefix to get a specific address (in which case you pass in the address) and in other cases you want to generate a prefix to get all addresses, in which case you leave it blank

src/storage/sets/flatbuffers/verificationStore.ts
varunsrin: @pfletcherhill i'd imagine we can implement this on MethodModel (or a subclassed VerificationModel). if that seems straightforward, i'm on board with punting it for now

PR #218, opened 2022-11-07
add a failing test case

Raynos:
## Motivation

EIP712 encoding is picky with uint8 array size.

## Change Summary

fid need to be treated as 32byte uint8arrays

## Merge Checklist

## Additional Context


Raynos:
@pfletcherhill asked me in slack if this edge case is a real edgecase and it turns out it is because the test fails, I think paul has a better idea of how to fix this.

Raynos:
The fact that onlytests changed in this PR means that we've not written the code to build or verify EIP712 in the `hub` grpc endpoints etc ? 

I expected some other code to flip the big/little endian of flatbuffer fids in the `hub` codebase

PR #220, opened 2022-11-08
feat(p2p): advertise public IP over gossip

sagar-a16z:
## Motivation

libp2p does not automatically resolve Peer's addresses. We needed a way to figure out how to reach peers discovered via Gossip. 

## Change Summary

- Fetch public IP's using the [ipify.org](https://www.ipify.org/) API.
- Publish public IP's over gossip

## Merge Checklist

### DIFF
--- a/src/utils/p2p.test.ts
+++ b/src/utils/p2p.test.ts
@@ -7,7 +7,6 @@ import {
   addressInfoFromParts,
   ipMultiAddrStrFromAddressInfo,
   addressInfoFromNodeAddress,
-  getPublicIp,
   p2pMultiAddrStr,
 } from '~/utils/p2p';
 
@@ -39,12 +38,6 @@ describe('p2p utils tests', () => {
       '/ip4/2600:1700:6cf0:990:2052:a166:fb35:830a',
       '/ip4/2600:1700:6cf0:990:2052:a166:fb35:830a/tcp/8080'
     );
-  });
-
-  test('get public IP', async () => {
-    const ip = await getPublicIp();
-    expect(ip).toBeDefined();
-    expect(isIP(ip)).toBeTruthy();
   });
 
   test('p2p multiaddr formatted string', async () => {
--- a/src/utils/p2p.test.ts
+++ b/src/utils/p2p.test.ts
@@ -1,10 +1,14 @@
+import { createEd25519PeerId } from '@libp2p/peer-id-factory';
 import { multiaddr, NodeAddress } from '@multiformats/multiaddr';
+import { isIP } from 'net';
 import {
   parseAddress,
   checkNodeAddrs,
   addressInfoFromParts,
   ipMultiAddrStrFromAddressInfo,
   addressInfoFromNodeAddress,
+  getPublicIp,
+  p2pMultiAddrStr,
 } from '~/utils/p2p';
 
 describe('p2p utils tests', () => {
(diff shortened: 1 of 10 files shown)


src/utils/p2p.ts
```diff
@@ -91,3 +93,44 @@ export const ipMultiAddrStrFromAddressInfo = (addressInfo: AddressInfo) => {
...
+};
+
+/**
+ *
+ * Fetches the publicly visible IP address of the running process
+ *
+ * @returns the public IPv4 or IPv6 as a string
+ */
+let lastIpFetch = { timestamp: new Date().getTime(), ip: '' };
+export const getPublicIp = async (): Promise<string> => {
```
varunsrin, line 115: seems like a place where we could use a Result type since it's possible to fail?

src/utils/p2p.ts
```diff
@@ -91,3 +93,44 @@ export const ipMultiAddrStrFromAddressInfo = (addressInfo: AddressInfo) => {
...
+  return new Promise((resolve, reject) => {
+    const now = new Date().getTime();
+    const since = now - lastIpFetch.timestamp;
+    if (since <= 10 * 60 * 1000 && lastIpFetch.ip != '') {
+      logger.debug({ component: 'utils/p2p', ip: lastIpFetch.ip }, `Cached public IP`);
+      resolve(lastIpFetch.ip);
+      return;
+    }
+    try {
+      get({ host: 'api64.ipify.org', port: 80, path: '/' }, (resp) => {
```
varunsrin, line 125: this may be pre-mature optimization, but you'll probably want to implement some retry logic here. defer to you as to whether that's done now or later.

src/utils/p2p.ts
```diff
@@ -91,3 +93,44 @@ export const ipMultiAddrStrFromAddressInfo = (addressInfo: AddressInfo) => {
...
+        resp.on('data', (ip: Buffer) => {
+          logger.info({ component: 'utils/p2p', ip: ip.toString() }, `Fetched public IP`);
+          lastIpFetch = { timestamp: now, ip: ip.toString() };
+          resolve(ip.toString());
+        });
+      });
+    } catch (err: any) {
+      reject(err);
```
varunsrin, line 133: probably worth logging here? unless you're guaranteeing that it will happen upstream

src/utils/p2p.test.ts
```diff
@@ -37,6 +41,26 @@ describe('p2p utils tests', () => {
     );
   });
 
+  test('get public IP', async () => {
```
varunsrin, line 44: if i'm reading this right, sounds like we are making a network call in the test? 

would recommend stubbing it or not testing, since it increases flakiness in CI which gets annoying when we have a lot of these

varunsrin: if you do decide to mock, the thing worth testing here is the caching behavior (i..e if you can fast forward time does it return a cached value ?)

varunsrin (approved):

src/utils/p2p.ts
sagar-a16z: Agree. I think it's fine for now since the caller does "retry" periodically.

src/utils/p2p.test.ts
sagar-a16z: Yeah we don't need to necessarily be testing someone else's API. Removed it

sagar-a16z: The caching would be useful to test, I might revisit that later.

sagar-a16z:
This PR has especially poor coverage because it's adding a method to fetch IP addresses from ipify.org. We don't want our unit tests to be dependent on a third-party api so that function and all the code branches that stem from it have no coverage.

PR #226, opened 2022-11-09
feat: prototype new error handling pattern and refactor one method to use it 

varunsrin:
## Motivation

See: https://github.com/farcasterxyz/hub/issues/213

## Change Summary

Introduces a new error class (HubError), a new error cataloging system (HubErrorCodes) and a standard result type for error prone functions (HubResult) that is built on neverthrow's Result. Also refactors one arbitrary method to demonstrate the syntax of the method in the application and tests. 

## Merge Checklist

### DIFF
--- a/src/network/p2p/node.test.ts
+++ b/src/network/p2p/node.test.ts
@@ -116,7 +116,10 @@ describe('node unit tests', () => {
   test('port and transport addrs in the Ip MultiAddr is not allowed', async () => {
     const node = new Node();
     const options = { ipMultiAddr: '/ip4/127.0.0.1/tcp/8080' };
-    await expect(node.start([], options)).rejects.toThrow();
+    const error = (await node.start([], options))._unsafeUnwrapErr();
+
+    expect(error.errCode).toEqual('unavailable');
+    expect(error.message).toMatch('unexpected transport/port information');
     expect(node.isStarted()).toBeFalsy();
     await node.stop();
   });
@@ -125,7 +128,11 @@ describe('node unit tests', () => {
     const node = new Node();
     // an IPv6 being supplied as an IPv4
     const options = { ipMultiAddr: '/ip4/2600:1700:6cf0:990:2052:a166:fb35:830a' };
-    await expect(node.start([], options)).rejects.toThrow();
+    expect((await node.start([], options))._unsafeUnwrapErr().errCode).toEqual('unavailable');
+    const error = (await node.start([], options))._unsafeUnwrapErr();
+
+    expect(error.errCode).toEqual('unavailable');
+    expect(error.message).toMatch('no protocol with code: 108');
     expect(node.isStarted()).toBeFalsy();
     await node.stop();
   });
--- a/src/network/p2p/node.ts
+++ b/src/network/p2p/node.ts
@@ -6,9 +6,10 @@ import { Mplex } from '@libp2p/mplex';
 import { TCP } from '@libp2p/tcp';
 import { Multiaddr } from '@multiformats/multiaddr';
 import { createLibp2p, Libp2p } from 'libp2p';
-import { err, ok, Result } from 'neverthrow';
+import { err, ok, Result, ResultAsync } from 'neverthrow';
 import { TypedEmitter } from 'tiny-typed-emitter';
-import { FarcasterError, ServerError } from '~/utils/errors';
+import { ServerError } from '~/utils/errors';
+import { HubError, HubResult } from '~/utils/hubErrors';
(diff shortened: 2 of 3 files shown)


Raynos:
Apparently `error.cause` was recently natively builtin to nodejs ( https://nodejs.org/api/errors.html#errorcause ) may want to add that too.

src/utils/errors.ts
```diff
@@ -1,31 +1,69 @@
...
+    // If no message input, use the error input's message, if present
+    if (!options.message) {
+      options.message = options.error?.message || '';
+    }
+    super(options.message);
 
-export class ServerError extends FarcasterError {
-  public override readonly statusCode = 500;
-}
+    this.code = code;
```
Raynos, line 34: nitpick: conflicts with `err.code` in node core ( https://nodejs.org/api/errors.html#errorcode ) for net.js / tcp errors, may want to choose another field and also if `options.error.code` exists and its a nodejs err from `net.connect()` etc copy over the `nodeCode` onto the `HubError`

src/utils/errors.ts
```diff
@@ -1,31 +1,69 @@
...
-export class ValidationError extends FarcasterError {
-  public override readonly statusCode = 400;
-}
+/**
+ * ErrorCode defines all the types of errors that can be raised in the Hub.
+ *
+ * A string union type is chosen over an enumeration since TS enums are unusual types that generate
+ * javascript code and may cause subtle, downstream issues.
+ */
+type ErrorCode =
```
Raynos, line 48: This is an error category on the order of magnitude of ~10 ; you may want to have a named error on the magnitude of ~100 to be able to drill in way deeper. ( https://github.com/uber/tchannel-node/blob/master/errors.js#L662-L779 )

It was useful when looking at a dashboard to know the difference between `tchannel.socket-closed` and `tchannel.socket-local-closed` , aka did I close the socket or the did the remote close the socket. There were maybe 10 different callsites where we create new socket closed errors with 10 different error messages, but quickly drilling down in DataDog between socket-closed & socket-local-closed added value compared to the high level categories of `unknown` or `network`

src/utils/errors.ts
```diff
@@ -1,31 +1,69 @@
...
+ */
+export class HubError extends Error {
+  /* Classifies the type of error being thrown */
+  public readonly code: ErrorCode;
 
-export class UnknownUserError extends FarcasterError {
-  public override readonly statusCode = 412; // Precondition Failed
-}
+  /* Indicates if if error message can be presented to the user */
+  public readonly presentable: boolean = false;
```
sagar-a16z, line 25: I don't think this should be optional. We should always allow users to be presentable, no?

src/utils/errors.ts
```diff
@@ -1,31 +1,69 @@
...
 
-export class BadRequestError extends FarcasterError {
-  public override readonly statusCode = 400;
-}
+  constructor(code: ErrorCode, options: Partial<HubErrorOpts>) {
```
sagar-a16z, line 27: @pfletcherhill has previously expressed a preference for being explicit about optionality in the interface declaration. 
https://github.com/farcasterxyz/hub/pull/130#discussion_r982877481

src/utils/errors.ts
```diff
@@ -1,31 +1,69 @@
...
+  | 'not_found'
+  /* TBD */
+  | 'db_error'
+  /* The request could not be completed, it may or may not be safe to retry */
+  | 'unavailable'
+  /* An unknown error was encountered */
+  | 'unknown';
+
+// TODO: move elsewhere
+export type HubResult<T> = Result<T, HubError>;
```
sagar-a16z, line 65: into `src/types/index.ts` maybe?

src/utils/errors.ts
varunsrin: i think @pfletcherhill's complaint was about mixing required and optional params in a partial (which i agree with)

here, all the params in options are optional (as the name implies), the Partial is just syntatic sugar to avoid excessively long function declarations

src/utils/errors.ts
varunsrin: will do at the end of the PR

src/utils/errors.ts
varunsrin: in most codebases you end up with errors that are clearly obvious and helpful to everyone (your signature isn't invalid) and errors that are only helpful to the developers of the project (component-5 failure). 

by explicitly requiring the former type to be tagged, the request layer can be opinionated about what it returns over the api/rpc in regular mode. it also gives you the flexibility to allow a "verbose" mode where unfriendly errors are surfaced. but we want to make sure this is explicitly turned on since it can unintentionally leak data that the operator of the hub may want to keep private from the caller.

Issue #213, opened 2022-11-04
feat: refactor methods to return Result<T, Err> when a failure is possible 


varunsrin:
**What is the feature you would like to implement?**

Refactor the codebase to implement error handling using neverthrow's Result in all public method calls that may fail

**Why is this feature important?**

Today, we have inconsistent error handling across our application. Some parts of our codebase use neverthrow’s Result (an Either monad inspired by Rust) extensively, some use it sparingly for public interfaces and some don’t use it at all and just throw. Our goal is to pick a single error handling strategy and use it consistently to make the code easier to reason about.

Hubs  are an eventually consistent system where correctness is paramount. Introducing bugs and errors, especially in parts that deal with consistency, will be very painful to recover from. It is important that people can make and release changes safely. Hubs are also an open source project that will have a few core contributors and a long tail of casual contributors who may be less familiar with the codebase. It is important that the core contributors can audit and safely reviews the changes being proposed. 

Our error handling system must also be ergonomic in Typescript. We chose a flexibly-typed language because it lets us choose between moving quickly and moving safely. Any solution we choose must also reflect that philosophy and not enforce unnecessary rigidity into our system.

We reject the throw-try-catch approach because it introduces invisible side effects. While it is ergonomic to use in javascript and the most common approach it makes method effects hard to understand. This increases the work that reviewers have to put in to keep the codebase safe. 

**Will the protocol spec need to be updated??**

No

**How should this feature be built? (optional)**

We propose using neverthrow’s Result type along with its corresponding [eslint plugin](https://github.com/mdbetancourt/eslint-plugin-neverthrow). If we cannot show that a method call succeeds without throwing an exception, it should return a Result. Methods on the “edge” that call unsafe code or other libraries that use a try-catch approach must wrap all their calls into a Result safely. 

This has the following benefits: 

1. Readers can immediately tell whether a function is safe or unsafe
2. Readers also know the type of error that may get thrown
3. Authors can never accidentally ignore an error. 

As long as reviewers of code enforce that all unsafe calls to external sources are wrapped in Results, code can be changed safely without introducing unhandled exceptions as shown below

```tsx
async read(): Result<string, JSONError> {
    if (succeeded) {
      return new Ok(['hello']);
    }
    return new Ok(response.result);
  }

const readResult = read();

// Handling errors and values separately (option 1)
if(readResult.isOk()) {
  log.info(value);
} else {
   log.warn(`Error`);
}

// Handling errors and values separately (option 2)
[...]

PR #227, opened 2022-11-10
refactor(p2p): network/p2p uses HubResults and HubErrors

varunsrin:
## Motivation

See: #213 

## Change Summary

Refactors all files under network/p2p and utils/p2p to use HubResults and HubErrors for error handling

## Merge Checklist

### DIFF
--- a/src/network/p2p/connectionFilter.ts
+++ b/src/network/p2p/connectionFilter.ts
@@ -21,7 +21,7 @@ export class ConnectionFilter implements ConnectionGater {
   }
 
   denyDialPeer = async (peerId: PeerId): Promise<boolean> => {
-    const deny = await this.shouldDeny(peerId.toString());
+    const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyDialPeer: denied a connection with ${peerId}`);
     }
@@ -29,7 +29,7 @@ export class ConnectionFilter implements ConnectionGater {
   };
 
   denyDialMultiaddr = async (peerId: PeerId, _multiaddr: Multiaddr): Promise<boolean> => {
-    const deny = await this.shouldDeny(peerId.toString());
+    const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyDialMultiaddr: denied a connection with ${peerId}`);
     }
@@ -45,7 +45,7 @@ export class ConnectionFilter implements ConnectionGater {
   };
 
   denyOutboundConnection = async (peerId: PeerId, _maConn: MultiaddrConnection): Promise<boolean> => {
-    const deny = await this.shouldDeny(peerId.toString());
+    const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyOutboundConnection: denied a connection with ${peerId}`);
     }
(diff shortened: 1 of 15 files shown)


src/utils/p2p.ts
```diff
@@ -4,55 +4,50 @@ import { err, ok, Result } from 'neverthrow';
...
+  if (options.port !== undefined || options.transport !== undefined) {
+    return err(new HubError('bad_request', 'unexpected multiaddr transport/port information'));
+  }
+
+  return parseAddress(listenCombinedAddr).match(
     (addr) => {
-      if (!addr) return err(new ServerError('Invalid Node MultiAddr: could not parse multiaddr'));
+      if (!addr) return err(new HubError('bad_request.parse_failure', 'invalid multiaddr listenCombinedAddr'));
+
+      // DISCUSS: can toOptions throw here? Should we fail if it does?
```
varunsrin, line 42: @sagar-a16z see comment

src/network/p2p/connectionFilter.ts
```diff
@@ -93,6 +93,9 @@ export class ConnectionFilter implements ConnectionGater {
   /*                               Private Methods                              */
   /* -------------------------------------------------------------------------- */
 
+  // DISCUSS: why are we accepting a null and why are we returning a promise?
```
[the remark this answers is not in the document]
sagar-a16z: 1. Null seems like an oversight. 
2. Promise here was just to establish the interface. I was imagining a more complex filter check down the road. Can definitely remove it since it's unnecessary right now.

src/utils/p2p.ts
```diff
@@ -134,3 +101,41 @@ export const getPublicIp = async (): Promise<Result<string, FarcasterError>> =>
...
+
+/* -------------------------------------------------------------------------- */
+/*                               Private Methods                              */
+/* -------------------------------------------------------------------------- */
+
+const checkIpAddr = (ipAddr: string): HubResult<void> => {
+  const parseListenIpAddrResult = parseAddress(ipAddr);
+  if (parseListenIpAddrResult.isErr()) return err(parseListenIpAddrResult.error);
+
+  // DISCUSS: we previously had a null check here (which seems impossible), but we have no
```
sagar-a16z, line 113: `parseAddress` here won't return an empty string. So we should be okay overall.

src/utils/p2p.ts
```diff
@@ -134,3 +101,41 @@ export const getPublicIp = async (): Promise<Result<string, FarcasterError>> =>
...
+  if (options.port !== undefined || options.transport !== undefined) {
+    return err(new HubError('bad_request', 'unexpected multiaddr transport/port information'));
+  }
+  return ok(undefined);
+};
+
+const checkCombinedAddr = (ipAddr: string): HubResult<void> => {
+  return parseAddress(ipAddr).match(
+    (addr) => {
+      // DISCUSS: can toOptions throw here? Should we fail if it does?
```
sagar-a16z, line 134: Yes, good catch. this can throw.

src/utils/p2p.ts
varunsrin: if `ipAddr` is an empty string, parseAddress returns an empty string

src/utils/p2p.ts
varunsrin, line 134: @sagar-a16z should we be throwing if this throws? or is ignoring the error intentional?

src/utils/p2p.ts
sagar-a16z: Oh! Okay I just had to write a test to check it and you're right. That's not good. I'd say we should make parseAddress return an error for empty strings.

src/utils/p2p.ts
sagar-a16z: We should just catch and return the error.

PR #247, opened 2022-11-15
feat: schema for gossip flatbuffers

sagar-a16z:
## Motivation

Gossip should to use Flatbuffers as its message container

## Change Summary

Add the schema for Gossip messages

## Merge Checklist


sagar-a16z:
Need to add typeguards for the new types here. Will do that in a follow-up.

sagar-a16z:
@sanjayprabhu I can also just add versioning to this schema. 
Quick question about that though, in your PR you used an Enum, which is fine I suppose but doesn't allow for minor or patch versions in the schema. 

Did you and @varunsrin settle on that versioning scheme?

sanjayprabhu:
I don't think we need minor and patch versions. IMO, we should only bump this when there's a breaking change, and it's not tied to the hub or protocol version. Open to other suggestions, but that's how I was thinking about it.

sagar-a16z:
> I don't think we need minor and patch versions. IMO, we should only bump this when there's a breaking change, and it's not tied to the hub or protocol version. Open to other suggestions, but that's how I was thinking about it.

Updated to include the version enum from your PR

src/utils/schemas/gossip.fbs
```diff
@@ -0,0 +1,75 @@
...
+ * @content - The message content to be broadcasted
+ * @topics - The topics this message belongs to. Multiple topics can be passed.
+ */
+table GossipMessage {
+  content: GossipContent (required);
+  topics: [string] (required);
+  version: GossipVersion = 1;
+}
+
+root_type GossipMessage;
```
pfletcherhill, line 75: Add a new line

src/utils/schemas/gossip.fbs
```diff
@@ -0,0 +1,75 @@
...
+ * @gossipAddress - The address at which this node is listening for Gossip messages. Unset if Gossip is not public.
+ * @rpcAddress - The address at which this node is serving RPC requests. Unset if RPC is not offered.
+ * @excludedHashes - The excluded hashes of the sender's current trie snapshot
+ * @count - The number of messages under the root
+ */
+table ContactInfoContent {
+  peer_id: [ubyte] (required);
+  gossip_address: GossipAddressInfo;
+  rpc_address: GossipAddressInfo;
+  excluded_hashes: [string];
```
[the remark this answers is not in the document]
sagar-a16z: I'll have to refactor sync for this, so not going to do that in this PR.

src/utils/generated/farcaster/contract-event-content.ts
```diff
@@ -0,0 +1,76 @@
+// automatically generated by the FlatBuffers compiler, do not modify
```
pfletcherhill, line 1: I think this file is out of date and you can delete it?

pfletcherhill (approved):
Approving, though can you delete the generated flatbuffer files and re-generate them to ensure all the files are relevant to the most recent schema?

PR #253, opened 2022-11-26
fix(RPCServer): use result value instead of result in multiaddr str

adaddeo:
## Motivation

The multiaddr was interpolating a Result object rather than the value inside the result, resulting in an incorrect mutliaddr string:
```
"[object Object]/tcp/50458"
```

## Change Summary

Updated the multiaddr function to check the result of `ipMultiAddrStrFromAddressInfo` and if successful return a string with the result value interpolated in the multiaddr string, else return undefined.

## Merge Checklist

## Additional Context

### DIFF
--- a/src/network/rpc/server.ts
+++ b/src/network/rpc/server.ts
@@ -139,7 +139,9 @@ export class RPCServer {
   get multiaddr() {
     const addr = this.address;
     if (!addr) return undefined;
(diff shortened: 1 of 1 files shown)


varunsrin:
good catch, i probably introduced this when i was refactoring return values to be results

Issue #163, opened 2022-10-11
chore:  review and fix or suppress eslint-security warnings


varunsrin:
**What is the feature you would like to implement?**

Review and fix all eslint security warnings or suppress them if they are false positives. 

```
➜  hub git:(varunsrin/eslint-security) yarn lint:fix

/Users/varun/src/farcaster/hub/src/network/sync/merkleTrie.ts
  60:18  warning  Variable Assigned to Object Injection Sink  security/detect-object-injection

/Users/varun/src/farcaster/hub/src/storage/db/rocksdb.ts
  90:11  warning  Found fs.open with non literal argument at index 0  security/detect-non-literal-fs-filename

/Users/varun/src/farcaster/hub/src/urls/chainAccountUrl.ts
  35:30  warning  Found non-literal argument to RegExp Constructor  security/detect-non-literal-regexp

/Users/varun/src/farcaster/hub/src/urls/chainUrl.ts
  28:30  warning  Found non-literal argument to RegExp Constructor  security/detect-non-literal-regexp

/Users/varun/src/farcaster/hub/src/urls/utils.ts
  14:12  warning  Generic Object Injection Sink                     security/detect-object-injection
  28:8   warning  Found non-literal argument to RegExp Constructor  security/detect-non-literal-regexp
  32:28  warning  Found non-literal argument to RegExp Constructor  security/detect-non-literal-regexp
  32:39  warning  Generic Object Injection Sink                     security/detect-object-injection

/Users/varun/src/farcaster/hub/src/utils/crypto.ts
  57:16  warning  Generic Object Injection Sink        security/detect-object-injection
  59:21  warning  Function Call Object Injection Sink  security/detect-object-injection
```

avichalp:
Hey @varunsrin,

If no one has already started working on it, I would like to take a stab. I see that is it a good opportunity to familiarize myself with the codebase.

varunsrin:
tim had a WIP PR here a few days ago: https://github.com/farcasterxyz/hub/pull/262

i would ping him there and see if he is still working on it, if not its all yours

varunsrin:
@avichalp and @djma i'm not sure if tim is working on it anymore, so this is up for grabs

TimDaub:
I think it should be merged asap to reduce potential merge conflicts

TimDaub:
actually whatever, there are too many now. I will close

PR #262, opened 2022-12-05
Fix and suppress eslint security warnings

TimDaub:
hey, so I tried adressing #163. Here's the current left-over output of yearn lint

```
yarn lint
yarn run v1.22.18
warning package.json: License should be a valid SPDX license expression
$ eslint  src/ --color --ext .ts

/Users/timdaub/Projects/farcaster/hub/src/urls/chainAccountUrl.ts
  35:30  warning  Found non-literal argument to RegExp Constructor  security/detect-non-literal-regexp

✖ 1 problem (0 errors, 1 warning)
```

☝🏼 seeking advice on how to fix the last remaining warning.

I'll comment inline of this PR to document some of the decisions I took.

- security/detect-object-injection: Is a security concern when a user can control the key to an object access, so in all cases where we're aware that the user can by no means control the key, we can suppress the warning.

### DIFF
--- a/src/network/p2p/connectionFilter.ts
+++ b/src/network/p2p/connectionFilter.ts
@@ -1,6 +1,5 @@
-import { ConnectionGater, MultiaddrConnection } from '@libp2p/interface-connection';
+import { ConnectionGater } from '@libp2p/interface-connection';
 import { PeerId } from '@libp2p/interface-peer-id';
-import { Multiaddr } from '@multiformats/multiaddr';
 import { logger } from '~/utils/logger';
 
 /**
@@ -28,7 +27,7 @@ export class ConnectionFilter implements ConnectionGater {
     return deny;
   };
 
-  denyDialMultiaddr = async (peerId: PeerId, _multiaddr: Multiaddr): Promise<boolean> => {
+  denyDialMultiaddr = async (peerId: PeerId): Promise<boolean> => {
     const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyDialMultiaddr: denied a connection with ${peerId}`);
@@ -36,7 +35,7 @@ export class ConnectionFilter implements ConnectionGater {
     return deny;
   };
 
-  denyInboundConnection = async (_maConn: MultiaddrConnection): Promise<boolean> => {
+  denyInboundConnection = async (): Promise<boolean> => {
     /**
      * A PeerId is not always known on incipient connections.
      * Don't filter incipient connections, later filters will catch it.
@@ -44,7 +43,7 @@ export class ConnectionFilter implements ConnectionGater {
     return false;
   };
 
-  denyOutboundConnection = async (peerId: PeerId, _maConn: MultiaddrConnection): Promise<boolean> => {
+  denyOutboundConnection = async (peerId: PeerId): Promise<boolean> => {
     const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyOutboundConnection: denied a connection with ${peerId}`);
@@ -52,7 +51,7 @@ export class ConnectionFilter implements ConnectionGater {
     return deny;
   };
 
-  denyInboundEncryptedConnection = async (peerId: PeerId, _maConn: MultiaddrConnection): Promise<boolean> => {
+  denyInboundEncryptedConnection = async (peerId: PeerId): Promise<boolean> => {
     const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyInboundEncryptedConnection: denied a connection with ${peerId}`);
@@ -60,7 +59,7 @@ export class ConnectionFilter implements ConnectionGater {
     return deny;
   };
 
-  denyOutboundEncryptedConnection = async (peerId: PeerId, _maConn: MultiaddrConnection): Promise<boolean> => {
+  denyOutboundEncryptedConnection = async (peerId: PeerId): Promise<boolean> => {
     const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyOutboundEncryptedConnection: denied a connection with ${peerId}`);
@@ -68,7 +67,7 @@ export class ConnectionFilter implements ConnectionGater {
     return deny;
   };
 
-  denyInboundUpgradedConnection = async (peerId: PeerId, _maConn: MultiaddrConnection): Promise<boolean> => {
+  denyInboundUpgradedConnection = async (peerId: PeerId): Promise<boolean> => {
     const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyInboundUpgradedConnection: denied a connection with ${peerId}`);
@@ -76,7 +75,7 @@ export class ConnectionFilter implements ConnectionGater {
     return deny;
   };
 
-  denyOutboundUpgradedConnection = async (peerId: PeerId, _maConn: MultiaddrConnection): Promise<boolean> => {
+  denyOutboundUpgradedConnection = async (peerId: PeerId): Promise<boolean> => {
     const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyOutboundUpgradedConnection: denied a connection with ${peerId}`);
--- a/src/network/sync/merkleTrie.ts
+++ b/src/network/sync/merkleTrie.ts
@@ -63,6 +63,8 @@ class MerkleTrie {
   public getDivergencePrefix(prefix: string, excludedHashes: string[]): string {
     const ourExcludedHashes = this.getSnapshot(prefix).excludedHashes;
     for (let i = 0; i < prefix.length; i++) {
+      // NOTE: `i` is controlled by for loop and hence not at risk of object injection.
+      // eslint-disable-next-line security/detect-object-injection
       if (ourExcludedHashes[i] !== excludedHashes[i]) {
         return prefix.slice(0, i);
       }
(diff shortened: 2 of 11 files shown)


src/network/sync/merkleTrie.ts
```diff
@@ -63,6 +63,7 @@ class MerkleTrie {
   public getDivergencePrefix(prefix: string, excludedHashes: string[]): string {
     const ourExcludedHashes = this.getSnapshot(prefix).excludedHashes;
     for (let i = 0; i < prefix.length; i++) {
+      // eslint-disable-next-line security/detect-object-injection
```
TimDaub, line 66: `i` is controlled by the for loop, so not dangerous

src/network/sync/syncEngine.ts
```diff
@@ -73,6 +73,7 @@ class SyncEngine {
     const ourSnapshot = this.snapshot;
     const excludedHashesMatch =
       ourSnapshot.excludedHashes.length === excludedHashes.length &&
+      // eslint-disable-next-line security/detect-object-injection
       ourSnapshot.excludedHashes.every((value, index) => value === excludedHashes[index]);
```
TimDaub, line 77: `index` is controlled by `every`

src/network/sync/trieNode.ts
```diff
@@ -46,7 +46,7 @@ class TrieNode {
    * every node that was traversed.
    */
   public insert(key: string, value: string, current_index = 0): boolean {
-    const char = key[current_index];
+    const char = key[parseInt(current_index, 10)];
```
TimDaub, line 49: honestly, I'm not a big fan of using `parseInt` here. What are alternatives?

src/storage/db/binaryrocksdb.ts
```diff
@@ -88,6 +88,7 @@ class RocksDB {
       } else {
         mkdir(this._db.location, { recursive: true }, (fsErr: Error | null) => {
           if (fsErr) reject(parseError(fsErr));
+          // eslint-disable-next-line security/detect-non-literal-fs-filename
```
TimDaub, line 91: it seems that eslint is falsely matching this one, as `this._db.open` isn't `fs.open`.

src/test/perf/scenario.ts
```diff
@@ -75,9 +75,8 @@ const makeMessages = async (userInfos: UserInfo[], config: ScenarioConfig, event
   if (event > MockFCEvent.Verification) throw 'Invalid event type for messages';
 
   const total = userInfos.length * (config.Adds + config.Removes + config.RemovesWithoutAdds);
-  // safe to disable here since `event` is validated above
```
TimDaub, line 78: IMO it's misleading to document why eslint can be disabled. For safety, it's better if anyone looking at this code interprets it at the time of inspection. Otherwise, this comment may serve as an excuse to not thoroughly inspect the code.

src/test/perf/scenario.ts
```diff
@@ -75,9 +75,8 @@ const makeMessages = async (userInfos: UserInfo[], config: ScenarioConfig, event
...
-  // eslint-disable-next-line security/detect-object-injection
   const progress = new ProgressBar(
+    // eslint-disable-next-line security/detect-object-injection
```
TimDaub, line 79: Moved eslint disable statement such that it caught the warning again.

src/urls/utils.ts
```diff
@@ -11,6 +11,7 @@ export const getParams = <T>(id: string, spec: IdentifierSpec): T => {
   const arr = splitParams(id, spec);
   const params: any = {};
   arr.forEach((value, index) => {
+    // eslint-disable-next-line security/detect-object-injection
```
TimDaub, line 14: `index` is controlled by `forEach`

src/urls/utils.ts
```diff
@@ -25,10 +26,14 @@ export const joinParams = (params: Params, spec: IdentifierSpec): string =>
     .join(spec.parameters.delimiter);
 
 export const isValidId = (id: string, spec: IdentifierSpec): boolean => {
+  // NOTE: Please don't call this function with a user-controlled `spec`
```
TimDaub, line 29: this isn't great, I have to say.

src/network/sync/trieNode.ts
msoica: `public insert(key: string, value: string, current_index: number = 0): boolean ` ?

TimDaub: I remember that I tried that (adding :number to current_index) but the eslint warning persisted

TimDaub: Also, technically can't Typescript infer from "= 0" that it must be a number?

src/utils/crypto.ts
```diff
@@ -54,8 +54,10 @@ const removeProps = (obj: Record<string, any>): void => {
   if (typeof obj === 'object' && obj != null) {
     Object.getOwnPropertyNames(obj).forEach((key) => {
       if (key[0] === '_') {
+        // eslint-disable-next-line
```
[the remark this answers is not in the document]
varunsrin: can you disable the specific rule as you did on line 31?

src/urls/utils.ts
```diff
@@ -25,10 +26,14 @@ export const joinParams = (params: Params, spec: IdentifierSpec): string =>
...
+  // variable.
+  // eslint-disable-next-line security/detect-non-literal-regexp
   if (!new RegExp(spec.regex).test(id)) return false;
   const params = splitParams(id, spec);
   if (params.length !== Object.keys(spec.parameters.values).length) return false;
   const matches = params
+    // eslint-disable-next-line
```
varunsrin, line 36: can you disable the specific rule as you did on line 31?

varunsrin:
@TimDaub great work so far, would you mind adding a comment above every disable explaining why its safe to disable? You've added the descriptions in the PR, but it is useful to have side by side in the code so that a future reader can easily understand why a decision was made, and evaluate if it still holds true.

src/network/p2p/connectionFilter.ts
```diff
@@ -28,55 +27,55 @@ export class ConnectionFilter implements ConnectionGater {
...
+  denyInboundEncryptedConnection = async (peerId: PeerId): Promise<boolean> => {
     const deny = this.shouldDeny(peerId.toString());
     if (deny) {
       logger.info(`ConnectionFilter denyInboundEncryptedConnection: denied a connection with ${peerId}`);
     }
     return deny;
   };
 
-  denyOutboundEncryptedConnection = async (peerId: PeerId, _maConn: MultiaddrConnection): Promise<boolean> => {
+  denyOutboundEncryptedConnection = async (peerId: PeerId): Promise<boolean> => {
```
varunsrin, line 62: the calls to these new method signatures in connectionFilter.test.ts are broken since they pass in two arguments instead of one

src/network/sync/trieNode.ts
varunsrin: what is the specific error that eslint is unhappy about?

varunsrin (changes requested):

src/network/sync/trieNode.ts
TimDaub:
```
/Users/timdaub/Projects/farcaster/hub/src/network/sync/trieNode.ts
  49:18  warning  Variable Assigned to Object Injection Sink  security/detect-object-injection

```

TimDaub:
> @TimDaub great work so far, would you mind adding a comment above every disable explaining why its safe to disable? You've added the descriptions in the PR, but it is useful to have side by side in the code so that a future reader can easily understand why a decision was made, and evaluate if it still holds true.

ok

src/network/p2p/connectionFilter.ts
TimDaub: ok so instead of removing the beforehand-assumed "unnecessary" argument options, rather we should suppress the warning?

src/test/perf/scenario.ts
TimDaub: @varunsrin I think actually that the eslint checking for a security issue like detect object injection in a test file is really unnecessary as the test environment is controlled by the developer and at their own risk. Shall I check if it's possible to disable security/detect-object-injection for the entire `test` directory?

src/network/sync/merkleTrie.ts
TimDaub: added comment

src/network/sync/trieNode.ts
TimDaub: I reverted `parseInt` as it failed the tests.

src/storage/db/rocksdb.ts
```diff
@@ -87,6 +87,7 @@ class RocksDB {
       } else {
         mkdir(this._db.location, { recursive: true }, (fsErr: Error | null) => {
           if (fsErr) reject(parseError(fsErr));
+          // eslint-disable-next-line security/detect-non-literal-fs-filename
```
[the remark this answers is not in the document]
TimDaub: added comment in the code.

src/urls/chainUrl.ts
```diff
@@ -25,6 +25,7 @@ export class ChainURL extends BaseChainURL {
       const chainId = new ChainId(chainIdParams);
 
       // check for extra invalid data before or after the chain ID
+      // eslint-disable-next-line security/detect-non-literal-regexp
```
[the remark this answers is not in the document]
TimDaub: added comment in code

TimDaub:
> @TimDaub great work so far, would you mind adding a comment above every disable explaining why its safe to disable? You've added the descriptions in the PR, but it is useful to have side by side in the code so that a future reader can easily understand why a decision was made, and evaluate if it still holds true.

Done.

TimDaub:
Screenshot 2022-12-07 at 17 58 15

I'm confused, I didn't close the PR...

TimDaub:
Damn I think it's because Github doesn't recognize the newly forced pushed commit :/

varunsrin:
@sagar-a16z do you want to quickly review the changes to the sync and see if you are OK with them?

@TimDaub i think we're ready to merge once conflicts are resolved

sagar-a16z:
Sync and p2p related changes look good 👍

PR #266, opened 2022-12-06
feat: Track Fnames

adityapk00:
## Motivation

Track fname updates in the hub so users can be identified by their preferred fname.

## Change Summary

Update Hub to track `NameRegistry` events, persist them in RocksDB. A new message type `UserNameAdd` allows users to broadcast their preferred fname. 

## Merge Checklist

## Additional Context
See https://github.com/farcasterxyz/hub/issues/128 for the feature spec. 

Since this is my first PR, I thought I'll send it for some early feedback. 

- This is missing a few test cases for some of the edge scenarios. 
- I'm having some issues signing the commits, so I might have to update the commits


adityapk00:
Tagging @varunsrin @pfletcherhill

src/storage/sets/flatbuffers/signerStore.ts
```diff
@@ -175,6 +181,24 @@ class SignerStore {
     this._eventHandler.emit('mergeContractEvent', event);
   }
 
+  /**
+   * Merges a NameRegistryEvent into the SignerStore, storing the causally latest event at the key:
+   * <RootPrefix:User><UserPostfix:NameRegistryEvent><fname>
+   */
+  async mergeNameRegistryEvent(event: NameRegistryEventModel): Promise<void> {
```
pfletcherhill, line 188: This method should go in the UserDataStore. The `mergeIdRegistryEvent` method is in the SignerStore, because it relates to signers, but new NameRegistryEvents could cause UserNameAdd messages to be revoked, so it should be in the UserDataStore.

src/storage/sets/flatbuffers/userDataStore.ts
```diff
@@ -133,11 +170,53 @@ class UserDataStore {
...
+  private async mergeNameAdd(message: UserNameAddModel): Promise<void> {
+    // TODO: Check that the user sending this message actually owns the fname
+
+    let tsx = await this.resolveUserNameMergeConflicts(this._db.transaction(), message);
+
+    // No-op if resolveMergeConflicts did not return a transaction
+    if (!tsx) return undefined;
+
+    // Revoke the name from the previous user
+    tsx = await this.revokeNameFromPreviousUser(tsx, message);
```
pfletcherhill, line 200: I don't think this should happen here. The UserDataStore should resolve conflicts between UserNameAdd messages (i.e. two from the same fid) but not affect data from other users. A UserNameAdd message should be revoked if/when:
* a NameRegistryEvent is merged that transfers ownership of the fname
* the fname expires

src/network/rpc/flatbuffers/client.ts
```diff
@@ -216,6 +223,13 @@ class Client {
     );
   }
 
+  async getUserName(fid: Uint8Array): HubAsyncResult<UserNameAddModel> {
+    return this.makeUnaryUserNameRequest(
```
pfletcherhill, line 227: I think you can use the `makeUnaryMessageRequest` method here instead of making a new method

src/network/rpc/flatbuffers/client.ts
```diff
@@ -312,6 +326,48 @@ class Client {
...
+            resolve(err(fromServiceError(e)));
+          } else if (response) {
+            resolve(ok(new NameRegistryEventModel(response)));
+          }
+        }
+      );
+    });
+  }
+
+  private makeUnaryUserNameRequest<RequestType, ResponseMessageType extends MessageModel>(
```
pfletcherhill, line 350: See comment above. You can get rid of this method in favor of `makeUnaryMessageRequest`.

pfletcherhill (changes requested):
Okay, I went through most of the files. Let me know if you feel differently, but I think we can refactor `UserNameAdd` to be just another `UserDataAdd` with its own `UserDataType` value. We can re-use a lot of code that way.

pfletcherhill (approved):
Approving, though please address comments before merging

PR #272, opened 2022-12-06
feat: Index custody address -> IdRegistryEvent

adityapk00:
## Motivation

There are instances (for eg. when handling NameRegistryEvents) when we need to lookup an `fid` by a custody address. This PR indexes `IdRegistryEvent`s so they can be looked up

## Change Summary

- Index IdRegistryEvents by custody address, so we can look up `fid` by custody addresses.
- Change `ContractEvent` to `IdRegistryEvent`

## Merge Checklist

## Additional Context

This is a big PR, but most of it is just the renaming.


src/storage/flatbuffers/types.ts
```diff
@@ -70,6 +70,8 @@ export enum UserPostfix {
   UserDataAdds = 17,
   /* Used to store a user data message */
   UserDataMessage = 18,
+  /* Used to index a custody address event in the add set */
+  CustodyAddressAdds = 19,
```
pfletcherhill, line 74: We should have `IdRegistryEventsByCustodyAddress` as a value in the `RootPrefix` enum rather than here. If it were here, it would mean that the key would be prefixed by <user byte, fid, ...> which isn't the case.

pfletcherhill (changes requested):

adityapk00:
@pfletcherhill Addressed all the comments, and fixed merge conflicts. This is now ready for review :)

src/storage/flatbuffers/idRegistryEventModel.ts
```diff
@@ -0,0 +1,112 @@
+import { ByteBuffer } from 'flatbuffers';
+import RocksDB, { Transaction } from '~/storage/db/binaryrocksdb';
+import { RootPrefix, UserPostfix } from '~/storage/flatbuffers/types';
+import { IdRegistryEvent, IdRegistryEventType } from '~/utils/generated/id_registry_event_generated';
+import SignerStore from '../sets/flatbuffers/signerStore';
+
+/** ContractEventModel provides helpers to read and write Flatbuffers ContractEvents from RocksDB */
```
pfletcherhill, line 7: Nit: `IdRegistryEventModel` instead of `ContractEventModel`

src/storage/flatbuffers/idRegistryEventModel.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  static from(bytes: Uint8Array) {
+    const event = IdRegistryEvent.getRootAsIdRegistryEvent(new ByteBuffer(bytes));
+    return new this(event);
+  }
+
+  /** <user prefix byte, fid, ID Registry event prefix byte> */
+  static primaryKey(fid: Uint8Array): Buffer {
+    return Buffer.concat([Buffer.from([RootPrefix.CustodyEvent]), Buffer.from(fid)]);
```
pfletcherhill, line 22: We should rename the enum value to `RootPrefix.IdRegistryEvent`

src/storage/flatbuffers/idRegistryEventModel.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Generates a unique key used to store the current custody address of a user -> fid
+   *
+   * @param address the custody address of the user
+   *
+   * @returns RocksDB key of the form <RootPrefix>:<address>
+   */
+  static byCustodyAddressKey(address: Uint8Array): Buffer {
+    return Buffer.concat([Buffer.from([RootPrefix.CustodyEventByCustodyAddress]), Buffer.from(address)]);
```
pfletcherhill, line 33: The root prefix enum value should be `RootPrefix.IdRegistryEventByCustodyAddress`

src/storage/flatbuffers/idRegistryEventModel.ts
```diff
@@ -0,0 +1,112 @@
...
+  static async getByCustodyAddress(db: RocksDB, custodyAddress: Uint8Array): Promise<IdRegistryEventModel> {
+    const buffer = await db.get(IdRegistryEventModel.byCustodyAddressKey(custodyAddress));
+    return IdRegistryEventModel.from(new Uint8Array(buffer));
+  }
+
+  static putTransaction(tsx: Transaction, event: IdRegistryEventModel): Transaction {
+    tsx = tsx.put(event.primaryKey(), event.toBuffer());
+
+    // This works for both register events and transfer events.
+    // TODO: For Register events, the `to` address is the custody address, right?
```
pfletcherhill, line 50: Yeah, `to` address is correct for both transfers and registers

PR #274, opened 2022-12-07
fix:Ensure fname is owned when adding UserDataName

adityapk00:
## Motivation

When processing `UserDataAdd` messages for `fname`, ensure that the fname is actually owned by the user.

## Change Summary

- Ensure the custody address of the `fid` and the `fname` are the same, ensuring that the `fid` owns the `fname` at the time of processing

## Merge Checklist

## Additional Context

I've added the checks in the `UserDataStore` since it already throws HubError for validation, but I can also move it to the engine's `validateMessage`

### DIFF
--- a/src/storage/sets/flatbuffers/userDataStore.ts
+++ b/src/storage/sets/flatbuffers/userDataStore.ts
@@ -10,7 +10,6 @@ import StoreEventHandler from '~/storage/sets/flatbuffers/storeEventHandler';
 import NameRegistryEventModel from '~/storage/flatbuffers/nameRegistryEventModel';
 import { eventCompare } from '~/utils/contractEvent';
 import { NameRegistryEventType } from '~/utils/generated/nameregistry_generated';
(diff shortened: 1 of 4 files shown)


src/storage/sets/flatbuffers/userDataStore.ts
```diff
@@ -113,6 +114,27 @@ class UserDataStore {
   /** Merges a UserDataAdd message into the set */
   async merge(message: MessageModel): Promise<void> {
     if (isUserDataAdd(message)) {
+      if (message.body().type() == UserDataType.Fname) {
```
pfletcherhill, line 117: Can we put this in the `validateMessage` method in the engine instead? Thus far stores only handle merge conflicts between messages, and the engine handles the validity of the message itself in the broader system.

pfletcherhill (commented):
Just one comment about moving the check into the engine, otherwise looks good

PR #279, opened 2022-12-09
fix: remove race condition from SyncEngine test

### LINKED ISSUE #215: bug: flakiness in test  "SyncEngine › should not sync if messages were added within the sync threshold"
**What is the bug?**

A sync will trigger unnecessarily if the timestamp change happens to be from 9s -> 10s, so approximately 10% of the time. See details here: https://github.com/farcasterxyz/hub/pull/279#discussion_r1048644390, thanks to @adaddeo for investigating.

See above

This bug causes two tests to fail sporadically locally and in CI 

```
Summary of all failing tests
FAIL src/network/sync/syncEngine.test.ts (12.641 s)
  ● SyncEngine › should not sync if messages were added within the sync threshold

    expect(received).toBeFalsy()

    Received: true

      118 |     // due to randomness
      119 |     const shouldSync = syncEngine.shouldSync(snapshot.excludedHashes, snapshot.numMessages + 1);
    > 120 |     expect(shouldSync).toBeFalsy();
          |                        ^
      121 |   });
      122 |
      123 |   test('initialize populates the trie with all existing messages', async () => {

      at Object.<anonymous> (src/network/sync/syncEngine.test.ts:120:24)
```

```
  ● differentialSync › sync converges even if signer removal is not within sync threshold

    expect(received).toBeLessThanOrEqual(expected)

    Expected: <= 1668118779.897
    Received:    1668118780

      237 |
      238 |       const serverSnapshot = hubASyncEngine.snapshot;
    > 239 |       expect(hubASyncEngine.snapshotTimestamp).toBeLessThanOrEqual(now / 1000);
          |                                                ^
      240 |       expect(hubBSyncEngine.shouldSync(serverSnapshot.excludedHashes)).toBeTruthy();
      241 |
      242 |       await hubBSyncEngine.performSync(serverSnapshot.excludedHashes, hubARPCClient);

      at Object.<anonymous> (src/test/e2e/differentialSync.test.ts:239:48)

```

adaddeo:
## Motivation

A race condition existed in a SyncEngine test that caused it to fail ~1/10 of the time. Closes #215.

## Change Summary

Used the [Jest Fake Timers API](https://jestjs.io/docs/jest-object#fake-timers) to fudge the time such that the test would never run into the race condition where a change in a higher order digit resulted in messages within the same sync threshold to still result in divergent snapshots.

## Merge Checklist

## Additional Context

This edge case can be repro'd 100% of the time by changing the faked system time to `Math.ceil(Date.now() / 100_000) * 100_000`.

It can also be reprod'd with `seq 1 12 | xargs -I{} yarn test syncEngine.test.ts --testNamePattern 'should not sync'` where one of the executions will have a snapshot timestamp ending in two zeros (i.e. `1670556600`).

### DIFF
--- a/src/network/sync/syncEngine.test.ts
+++ b/src/network/sync/syncEngine.test.ts
@@ -1,3 +1,4 @@
+import { jest } from '@jest/globals';
 import { Factories } from '~/test/factories';
 import Engine from '~/storage/engine';
 import { SyncEngine } from '~/network/sync/syncEngine';
@@ -123,13 +124,25 @@ describe('SyncEngine', () => {
 
   test('should not sync if messages were added within the sync threshold', async () => {
     const user = await mockFid(engine, faker.datatype.number());
+
+    /**
+     * When the time elapsed between two messages causes a change in a higher
+     * order digit of the timestamp, a comparison of the snapshots will find a
+     * divergence despite the messages occurring within the sync threshold.
+     *
+     * We work around this by fudging the system time to this avoid scenario.
+     */
+    jest.useFakeTimers();
+    jest.setSystemTime(Math.ceil(Date.now() / 100_000) * 100_000 + 50_000);
     const snapshotTimestamp = syncEngine.snapshotTimestamp;
     await addMessagesWithTimestamps(user, [snapshotTimestamp - 3, snapshotTimestamp - 2, snapshotTimestamp - 1]);
 
     const snapshot = syncEngine.snapshot;
+
     // Add a message after the snapshot, within the sync threshold
     await addMessagesWithTimestamps(user, [snapshotTimestamp + 3]);
     expect(syncEngine.shouldSync(snapshot.excludedHashes)).toBeFalsy();
+    jest.useRealTimers();
   });
 
   test('initialize populates the trie with all existing messages', async () => {


adaddeo:
feel free to edit my comment, that was my best attempt being newly acquainted with the algorithm

src/network/sync/syncEngine.test.ts
```diff
@@ -123,13 +124,25 @@ describe('SyncEngine', () => {
...
+
+    /**
+     * When the time elapsed between two messages causes a change in a higher
+     * order digit of the timestamp, a comparison of the snapshots will find a
+     * divergence despite the messages occurring within the sync threshold.
+     *
+     * We work around this by fudging the system time to this avoid scenario.
+     */
+    jest.useFakeTimers();
+    jest.setSystemTime(Math.ceil(Date.now() / 100_000) * 100_000 + 50_000);
```
varunsrin, line 136: can you explain why the specific limits of 100_000 and 50_000 are chosen here?

src/network/sync/syncEngine.test.ts
```diff
@@ -123,13 +124,25 @@ describe('SyncEngine', () => {
 
   test('should not sync if messages were added within the sync threshold', async () => {
     const user = await mockFid(engine, faker.datatype.number());
+
+    /**
+     * When the time elapsed between two messages causes a change in a higher
```
varunsrin, line 129: doesn't this indicate a bug in the sync logic itself? that is, we should not be triggering the sync but we will be and this change just suppresses the test

varunsrin (changes requested):

varunsrin:
there is also another test `differentialSync › sync converges even if signer removal is not within sync threshold` that might require a similar fix

src/network/sync/syncEngine.test.ts
adaddeo: after another look I think this should be considered bug

the root cause is that within excludedHashes there are two representations of if a trie has no excluded items at a specified path, either `hash(null)` if a node exists and there are no excluded items or no value if the node doesn't exist

in this test, when the snapshot timestamp ends in `00` seconds, messages merged a few seconds before and a few seconds after will be inserted on different branches

snapshot A has value 099, excludedHashes for prefix 10 is [hash(0)]
snapshot B has values 099 and 101, excludedHashes for prefix 10 is [hash(0), hash(null)]

a sync is triggered even though there is no difference in excluded items for the tries along the prefix

possible solutions:
- excludedHashes contains a value for every index of the prefix (`len(excludedHashes) === len(prefix)`), whether the full path exists or not (e.g. pad with "empty hash" value if a path terminates before the prefix ends)
- leave excludedHashes as is but when comparing excludedHashes take into account that no item at an index represents no excluded items

in my /very humble/ opinion the first solutions seems easier to reason about since it provides explicit values and theres something nice about the `len(excludedHashes) === len(prefix)` invariant. a downside is extra data needs to be sent, but as the trees becomes less sparse this will matter less. could optimize what value represents empty, null instead of hash(null).

interested to hear thoughts of you or someone w/ a deeper understanding than me

varunsrin: ok, im actually thinking through some other sync design problems today which may or may get rid of this logic anyway. let's hit pause on this until i'm done with the doc and revisit. 

thanks for investigating and finding this, i'll update the ticket to indicate that it's a bug

varunsrin:
cc @adityapk00 who is working on sync

pfletcherhill:
Should this PR still be open?

varunsrin:
i think it can be closed, but will defer to @adaddeo

adityapk00:
I think this can be closed

adaddeo:
@adityapk00  the test is still in the codebase but commented out, do you want me to remove it?

adityapk00:
Yes, probably a good idea to remove the commented out test. 

The syncEngine now depends on gossip to catch up to messages that are <10s old, so this test is not needed any more.

PR #284, opened 2022-12-09
feat: Pull log events from Id and Name registries

adityapk00:
## Motivation

Syncing Id and Name registries from an Ethereum node. 

## Change Summary

- Sync events from the goreli testnet from the Id and Name contracts 
- Events are logged, which are then pushed into the Hub store

## Merge Checklist

## Additional Context

This is a draft PR, I'll continue to make updates to this so we can track it.

- [X] Connect to ETH node
- [X] Sync ongoing events as they happen from Id and Name registries
- [X] Sync missed (old) events on Hub startup, so that the hub can catch up after being offline
- [X] New Tests for `EthEventsProvider`


src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,300 @@
...
+      idEvents = [];
+      this._idEventsByBlock.set(blockNumber, idEvents);
+    }
+    idEvents.push(idRegistryEventModel);
+  }
+
+  private async addIdRegistryEvent(idRegistryEventModel: IdRegistryEventModel) {
+    const r = await this._engine.mergeIdRegistryEvent(idRegistryEventModel);
+    if (r.isErr()) {
+      log.error({ err: r._unsafeUnwrap() }, 'IdRegistryEvent failed to merge');
```
varunsrin, line 246: You shouldn't need to unsafe unwrap because the if clause above guarantees that it's an error

varunsrin (commented):
Directionally looks good to me, happy to review again once tested and todos are resolved

adityapk00:
The test failure seems to be a flaky test, but everything else seems to pass fine.

src/hub.ts
```diff
@@ -101,6 +108,18 @@ export class Hub extends TypedEmitter<HubEvents> implements RPCHandler {
     this.gossipNode = new Node();
     this.rpcServer = new RPCServer(this);
     this.syncEngine = new SyncEngine(this.engine);
+
+    // TODO: This should be the primary engine
```
varunsrin, line 112: are you planning to do this in a follow on PR?

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    // Call Eth Node to check connection
+    this.callEthNode();
+  }
+
+  /**
+   *
+   * Setup a Eth Events Provider with Goreli testnet, which is currently used for Production Farcaster Hubs.
+   */
+  public static makeWithGoreli(
```
varunsrin, line 96: typo - i believe it should be goerli 

this happens in a few comments and other places, so worth a find and replace

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+  private _nameEventsByBlock: Map<number, Array<NameRegistryEventModel>>;
+
+  private _lastBlockNumber: number;
+
+  constructor(
+    engine: Engine,
+    jsonRpcProvider: providers.BaseProvider,
+    idRegistryContract: Contract,
+    nameRegistryContract: Contract,
+    numConfirmations: number
```
varunsrin, line 47: should we hard code this for now, and provide rationale for why we chose a specific number? worry is that someone without understanding the nuances tweaks it later to makes this faster and introduces other problems

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+  /**
+   *
+   * Setup a Eth Events Provider with Goreli testnet, which is currently used for Production Farcaster Hubs.
+   */
+  public static makeWithGoreli(
+    engine: Engine,
+    networkUrl: string,
+    IdRegistryAddress: string,
+    NameRegistryAddress: string,
+    numConfirmations = 6
```
varunsrin, line 101: should we hardcode confirmations (as mentioned above) and add logic for why we chose six? 

the worry is that someone will tweak this without understand the implications on downstream sync

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+      idRegistryContract,
+      nameRegistryContract,
+      numConfirmations
+    );
+
+    return provider;
+  }
+
+  /** Connect to Ethereum RPC */
+  private async callEthNode() {
```
varunsrin, line 121: this method does a lot more than just calling a node, perhaps `syncHistoricalState` or something like that would be more descriptive?

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+        type === NameRegistryEventType.NameRegistryTransfer ? BigNumber.from(0) : event.args?.at(1);
+
+      await this.handleNameRegistryEvent(from, to, tokenId, type, expiry, event);
+    }
+  }
+
+  /** Handle a new block. Processes all events in the cache that have now been confirmed */
+  private async handleNewBlock(blockNumber: number) {
+    // Get all blocks that have been confirmed into a single array and sort.
+    const confirmedBlocksSet = new Set([...this._nameEventsByBlock.keys(), ...this._idEventsByBlock.keys()]);
```
varunsrin, line 196: nit: should we call these cached blocks to make their purpose more obvious to the reader? my understanding is that these are events from blocks that are confirmed from the blockchain, but we don't accept them and keep them in cache

src/storage/engine/flatbuffers/providers/ethEventsProvider.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  const addBlocks = async (startAt: number, n: number): Promise<number> => {
+    for (let i = 0; i < n; i++) {
+      expect(mockRPCProvider.emit('block', startAt + i)).toBeTruthy();
+      await waitForBlock(startAt + i);
+    }
+
+    return startAt + n;
+  };
+
+  test('processes Register events', async () => {
```
varunsrin, line 104: we have a test for IdRegistry Register and NameRegistry transfer

should we also add IdRegistry Transfer, NameRegistry Register and NameRegistry Renew for thoroughness and to make it easier for future folks to refactor without breaking functionality?

varunsrin (approved):
Appoving, with some minor suggestions for naming and test improvements. Feel free to merge once you have resolved the ones you agree with.

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
adityapk00: Oops. Fixed everywhere.

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
adityapk00: Done

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+    // Sync old Name events
+    await this.syncOldNameEvents(NameRegistryEventType.NameRegistryTransfer, numBlocksToSync);
+    await this.syncOldNameEvents(NameRegistryEventType.NameRegistryRenew, numBlocksToSync);
+  }
+
+  /**
+   * Sync old Id events that may have happened before hub was started. We'll put them all
+   * in the sync queue to be processed later, to make sure we don't process any unconfirmed events.
+   */
+  private async syncOldIdEvents(type: IdRegistryEventType, numBlocksToSync: bigint) {
```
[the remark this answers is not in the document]
adityapk00: I changed the methods to try-catch and log errors. Basically, these errors should never happen, because we're parsing logs coming from the Eth contracts, and it should never be unexpected. If anything here throws, it is a code bug, so we should fix them. 

Basically, there are no "expected errors" here.

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+    const typeString = type === IdRegistryEventType.IdRegistryRegister ? 'Register' : 'Transfer';
+
+    const oldIdEvents = await this._idRegistryContract.queryFilter(typeString, -1 * Number(numBlocksToSync));
```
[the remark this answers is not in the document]
adityapk00: This will not happen because we register the `on('Register')` handlers on the contracts first, before we get the latest block. So, any new blocks produced while we're going through this will be handled via that code path.

It is possible that a block might be processed twice, but our regular `merge()` logic will handle that just fine.

src/storage/engine/flatbuffers/providers/ethEventsProvider.ts
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  /**
+   * Sync old Name events that may have happened before hub was started. We'll put them all
+   * in the sync queue to be processed later, to make sure we don't process any unconfirmed events.
+   */
+  private async syncOldNameEvents(type: NameRegistryEventType, numBlocksToSync: bigint) {
+    const typeString = type === NameRegistryEventType.NameRegistryTransfer ? 'Transfer' : 'Renew';
+
+    const oldNameEvents = await this._nameRegistryContract.queryFilter(typeString, -1 * Number(numBlocksToSync));
```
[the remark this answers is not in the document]
adityapk00: Using the `-numblocks` is fine because we register the `on(event)` handlers on the contracts first, so we'll never miss a block.

src/storage/engine/flatbuffers/providers/ethEventsProvider.test.ts
adityapk00: Expanded the tests to add the additional tests

PR #288, opened 2022-12-15
bugfix: Edge case where a block would not be synced

adityapk00:
## Motivation
Fix a bug in the initial (historical) sync where a block might be missed

## Change Summary
If a block is created just before we start historical sync, the earliest block might get missed. 

## Merge Checklist

## Additional Context

Follow up from #284


varunsrin (approved):
LGTM

PR #294, opened 2022-12-17
chore: remove set tests since sets will be deprecated

varunsrin:
## Motivation

Sets are being replaced with Stores, but cannot yet be removed since sync has dependencies on them that need to be refactored. However, their tests can be removed safely and this should speed up our CI runs which are taking 8+ minutes.  

## Change Summary

Tests for all Sets are being removed

## Merge Checklist


varunsrin:
@pfletcherhill tagging as an FYI

PR #309, opened 2022-12-20
refactor: order imports for easier readability

varunsrin:
## Motivation

When looking at files with a lot of imports, its hard to parse dependencies when they are unordered

## Change Summary

Orders dependencies

## Merge Checklist


varunsrin:
this package adds a lot of deps like babel, which bloats the environment. it also thinks that `./` imports are external libs which is not ideal since we use that syntax a lot. holding off on merging for now.

PR #318, opened 2022-12-21
refactor: minimize external dependencies in RPC client

### LINKED ISSUE #312: refactor:  minimize external dependencies in RPC client

Refactor Client to eliminate external dependencies so that it can be extracted into a package

Client is a folder that will be shared between consumers of the Hub and the Hub itself. It is important that this has as few external dependencies as possible so that it can be easily extracted into a package in the future. 

No

**How should this feature be built? (optional)**

Refactor Client to eliminate dependencies on 'net', 'logger', 'utils/p2p'. It may be a good idea to introduce a HubRPCClient that wraps around this with additional functionality, or it may be possible to just push the additional functionality upstream to the callers within the Hub. Also consider eliminating the HubResult dependency if its not too convoluted. Ideally, the client should only need to depend on things within its folder, grpc and flatbuffers.

adaddeo:
## Motivation

Client is a folder that will be shared between consumers of the Hub and the Hub itself. It is important that this has as few external dependencies as possible so that it can be easily extracted into a package in the future. Closes #312 

## Change Summary

The 'net', 'logger', 'utils/p2p' dependencies are removed from Client by refactoring hub specific logic into HubRPCClient and introducing ClientError.

## Merge Checklist

### DIFF
--- a/app/src/rpc/server/index.ts
+++ b/app/src/rpc/server/index.ts
@@ -95,38 +95,6 @@ export const toTrieNodeMetadataResponse = (metadata: NodeMetadata): flatbuffers.
   return response;
 };
 
-export const fromNodeMetadataResponse = (response: flatbuffers.TrieNodeMetadataResponse): NodeMetadata => {
-  const children = new Map<string, NodeMetadata>();
-  for (let i = 0; i < response.childrenLength(); i++) {
-    const child = response.children(i);
-
-    const prefix = new TextDecoder().decode(child?.prefixArray() ?? new Uint8Array());
-    // Char is the last char of prefix
-    const char = prefix[prefix.length - 1] ?? '';
-
-    children.set(char, {
-      numMessages: Number(child?.numMessages()),
-      prefix,
-      hash: new TextDecoder().decode(child?.hashArray() ?? new Uint8Array()),
-    });
-  }
-
-  return {
-    prefix: new TextDecoder().decode(response.prefixArray() ?? new Uint8Array()),
-    numMessages: Number(response.numMessages()),
-    hash: new TextDecoder().decode(response.hashArray() ?? new Uint8Array()),
-    children,
-  };
-};
-
-export const fromSyncIdsByPrefixResponse = (response: flatbuffers.GetAllSyncIdsByPrefixResponse): string[] => {
-  const ids = [];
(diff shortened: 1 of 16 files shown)


Niravprajapati1 (approved):
Awesome 👍

varunsrin:
Looks great, but I think some conflicts need to be resolved now

adaddeo:
> Looks great, but I think some conflicts need to be resolved now

rebased

varunsrin:
LGTM 

I don't love having to introduce a new ClientError type, but I'm not sure there is a cleaner solution. @pfletcherhill thoughts on this?

adaddeo:
@varunsrin after taking a look at this again I think updating the PR to drop ClientError in favor of continuing with using the HubError in the client module is a better path for the PR. Progress is made with 'net', 'logger', 'utils/p2p' dependencies dropped and we avoid prematurely introducing another class

varunsrin:
@adaddeo i like that approach better for now. this may need to be rebased over the latest conflicts

adaddeo:
@varunsrin rebased

app/src/rpc/server/index.ts
```diff
@@ -95,38 +95,6 @@ export const toTrieNodeMetadataResponse = (metadata: NodeMetadata): flatbuffers.
   return response;
 };
 
-export const fromNodeMetadataResponse = (response: flatbuffers.TrieNodeMetadataResponse): NodeMetadata => {
```
adaddeo, line 98: moved to `client/index.ts`

app/src/rpc/server/serviceImplementations/syncImplementation.ts
```diff
@@ -154,23 +154,3 @@ export const createSyncServiceRequest = (fid: Uint8Array): flatbuffers.GetAllMes
     new ByteBuffer(builder.asUint8Array())
   );
 };
-
-export const createByPrefixRequest = (prefix: Uint8Array): flatbuffers.GetTrieNodesByPrefixRequest => {
```
adaddeo, line 158: moved to `client/serviceRequests/syncRequests`

adaddeo:
@varunsrin rebased after mono repo changes

PR #330, opened 2022-12-27
perf: use swc to transform TS in jest

### LINKED ISSUE #314: perf: use swc or esbuild toolchain to transpile code locally

All local actions (yarn start, yarn test) should use a fast transpiler like esbuild or swc, and remote CI actions should compile the code at most once before running tests.  

Faster build times makes it easier to work with the Hub codebase

**How should this feature be built? (optional)**

Currently, we use tsx for yarn start which uses esbuild under the hood, and nothing special for yarn test. In CI, we run yarn tsc and then yarn test with some CI flags. Ideally, we'd do a full typecheck in CI but when running locally we'd just do a fast transpile and execute. We may also be double compiling in CI adding to our run times, needs investigation. 

**Additional context**

There is some complexity due to ESM where additional configuration is needed to get certain packages to work. For example, swc is a drop in replacement in jest, but esbuild requires some fiddling with configurations. Similarly, tsx was the only drop in replacement for yarn start that actually worked.

adaddeo:
## Motivation

Closes #314. Speed up `yarn test` by ~2x.

## Change Summary

Replace `ts-jest` with `@swc/jest` to transform our TS files in Jest.

## Merge Checklist

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,7 +1,6 @@
-import type { JestConfigWithTsJest } from 'ts-jest';
+import type { Config } from 'jest';
 
-const jestConfig: JestConfigWithTsJest = {
-  preset: 'ts-jest/presets/default-esm',
+const jestConfig: Config = {
   testEnvironment: 'node',
   moduleNameMapper: {
     '^~/(.*)$': '<rootDir>/src/$1',
@@ -9,15 +8,13 @@ const jestConfig: JestConfigWithTsJest = {
   },
   coveragePathIgnorePatterns: ['<rootDir>/build/', '<rootDir>/node_modules/', '<rootDir>/src/utils/generated/'],
   testPathIgnorePatterns: ['<rootDir>/build', '<rootDir>/node_modules', '<rootDir>/src/utils/generated/'],
-  // transform ts files with ts-jest and enable ESM
+  extensionsToTreatAsEsm: ['.ts'],
+  /**
+   * For high performance with minimal configuration transform with TS with swc.
+   * @see https://github.com/farcasterxyz/hub/issues/314
+   */
   transform: {
-    '^.+\\.tsx?$': [
-      'ts-jest',
-      {
-        tsconfig: 'tsconfig.json',
-        useESM: true,
-      },
-    ],
+    '^.+\\.(t|j)sx?$': '@swc/jest',
   },
 };
 
--- a/package.json
+++ b/package.json
@@ -34,6 +34,9 @@
   "devDependencies": {
     "@faker-js/faker": "^7.6.0",
     "@libp2p/interface-mocks": "^7.0.1",
+    "@swc/cli": "^0.1.57",
+    "@swc/core": "^1.3.24",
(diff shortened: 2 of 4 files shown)


adaddeo:
The security check that is failing is for the [install script](https://github.com/swc-project/swc/blob/17bc1a6d8659ad5a75bccace73e7083ee0a0aaba/node-swc/src/postinstall.ts) from @swc/core, here's what the script does:

> It checks if corresponding optional dependencies for native binary is installed and can be loaded properly.
> If it fails, it'll internally try to install `@swc/wasm` as fallback.

This alert can be ignored by replying with a comment starting `@SocketSecurity ignore @swc/core@1.3.24`

varunsrin:
I ran some benchmarks locally, and the improvement seems closer to ~ 15%, not 100%.  I still think its worth merging in, but I want to make sure this discrepancy isn't the result of some other config issue before we do that.  

I used hyperfine to benchmark, and here's the output when running yarn test from main and clearing the cache each time

```
➜  hub git:(main) ✗ hyperfine 'rm -rf build && yarn test'
Benchmark 1: rm -rf build && yarn test
  Time (mean ± σ):     12.751 s ±  0.262 s    [User: 60.905 s, System: 3.453 s]
  Range (min … max):   12.155 s … 13.006 s    10 runs
```

Here's the output when doing it on your branch: 

```
➜  hub git:(ad/swc-build) ✗ hyperfine 'rm -rf build && yarn test'
Benchmark 1: rm -rf build && yarn test
  Time (mean ± σ):     11.166 s ±  0.222 s    [User: 40.538 s, System: 2.851 s]
  Range (min … max):   11.004 s … 11.657 s    10 runs
  ```

**Comment by adaddeo** [comment]
Jest has its own cache for transformed modules separate from `./build`, what do you get w `hyperfine 'yarn test --no-cache'`?

ts-jest:
```
Benchmark 1: yarn test --no-cache
  Time (mean ± σ):     51.285 s ±  0.889 s    [User: 282.376 s, System: 15.140 s]
  Range (min … max):   50.079 s … 52.389 s    5 runs
```

swc:
```
Benchmark 1: rm -rf build && yarn test --no-cache
  Time (mean ± σ):     21.337 s ±  0.384 s    [User: 82.215 s, System: 7.419 s]
  Range (min … max):   20.936 s … 21.719 s    5 runs
```

**Comment by varunsrin** [comment]
That comes out much better, shows closer to a 50% improvement on an M1. I think its worth merging 

```
✨  Done in 19.70s.
  Time (mean ± σ):     18.909 s ±  0.646 s    [User: 109.488 s, System: 6.586 s]
  Range (min … max):   17.801 s … 19.797 s    10 runs
```

.```
✨  Done in 12.46s.
  Time (mean ± σ):     12.653 s ±  0.185 s    [User: 41.741 s, System: 3.459 s]
  Range (min … max):   12.454 s … 13.080 s    10 runs
```

varunsrin:
@SocketSecurity ignore @swc/core@1.3.24

PR #348, opened 2023-01-01
chore(shared-prettier-config) & chore(shared-tsconfig)

Pavan-Nambi:
Prettier configuration and tsconfig is extracted to a shared package and imported into each workspace

## Motivation

https://github.com/farcasterxyz/hub/issues/344#issue-1515272819
https://github.com/farcasterxyz/hub/issues/345#issue-1515273246

## Change Summary

made new prettier config package and tsconfig package and imported it into flatbuffers

## Merge Checklist

## Additional Context


packages/flatbuffers/tsconfig.json
```diff
@@ -11,8 +10,10 @@
     "baseUrl": "./src",
     "outDir": "build",
     "rootDir": "src",
-    "isolatedModules": true,
+    "isolatedModules": true
   },
+  //baseUrl is  ./src so..
```
varunsrin, line 15: i'm not sure i understand this comment, what does it mean?

packages/prettier-config-custom/index.json
```diff
@@ -0,0 +1,9 @@
+{
```
varunsrin, line 1: does an index.json export work when you test it? i've typically only seem them as index.ts with a modules.export on top.

packages/tsconfig-custom/base.json
```diff
@@ -0,0 +1,34 @@
+{
```
varunsrin, line 1: same comment about json, we should do the same thing that we do with the eslint package

varunsrin (commented):
Thanks for the PR! Two issues that will need to be addressed in addition to the comments: 

1. I noticed this PR added a `package-lock.json` which means you may be using a different package manager. Can you use yarn instead so we don't need this large diff?

2. The commits aren't signed, so Github won't permit a merge. Please see the contributing.md doc in the repo for instructions on how to get that working correctly.

packages/flatbuffers/tsconfig.json
Pavan-Nambi: i was checking path for config file there i wrote for myself forgot to remove.. sorry will remove it

packages/prettier-config-custom/index.json
Pavan-Nambi: acc to prettier docs it said so hmm ya i will update it with modules.export into a js file

packages/tsconfig-custom/base.json
Pavan-Nambi: for shared tsconfig i only saw using json...  and extending it on tsconfg.json
i will look into bit further ig thankyou for input

packages/prettier-config-custom/index.json
Pavan-Nambi: i have referred multiple projects regarding sharing tsconfig in all projects i have seen so far its done using either making a .json file and extending wherever needed... 
i am not sure how else this should be done can you show me some reference if possible?
ya eslint and prettier we can use modules.export but idk about tsconfig...

Pavan-Nambi: apologies if i have made any inconvenience i am bit new to open source....

packages/prettier-config-custom/index.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
varunsrin, line 1: there doesn't seem to be any config anymore?

packages/prettier-config-custom/index.json
varunsrin: i was referring to prettier here. 

the prettier config can probably be set up similar to eslint-config-custom in this repo, which uses a .js file 

the tsconfig as you said should probably have a json config, as defined here: https://turbo.build/repo/docs/handbook/linting/typescript

packages/flatbuffers/tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../tsconfig-custom/tsconfig.json",
```
varunsrin, line 2: we should be able to reference it via package name since it is added as a dependency. using a path makes it more brittle. see this guide: https://turbo.build/repo/docs/handbook/linting/typescript

varunsrin (changes requested):

packages/flatbuffers/tsconfig.json
Pavan-Nambi: ok

Pavan-Nambi:
hey and if you are free can you say what causes bin script confusion... thats the reason i reverted changes at first i tried googling but it just said due to changing bin to package.json but i haven't actually done anything of tht sort
and its saying its a high security threat soo what exactly caused it here..? or normally what sort of thing causes bin script confusion?

https://socket.dev/blog/npm-bin-script-confusion

varunsrin:
that may have been a problem introduced because you had the npm package lock checked in which retriggered some checks, it seems resolved now

packages/flatbuffers/package.json
```diff
@@ -14,7 +14,9 @@
     "lint:fix": "yarn run lint -- --fix"
   },
   "dependencies": {
+    "prettier-config-custom": "*",
```
varunsrin, line 17: it appears that the custom configs for prettier and ts-config are only used by the flatbuffers package. we'll also need to do this for the primary app package

packages/prettier-config-custom/index.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  printWidth: 100,
+  tabWidth: 4,
+  useTabs: false,
+  semi: true,
+  singleQuote: true,
+  trailingComma: "es5",
+  bracketSpacing: true,
+  jsxBracketSameLine: false,
+  arrowParens: "always",
```
varunsrin, line 10: let's not introduce new configuration here - this should just be a copy of the current prettier config

varunsrin (changes requested):
added a few comments. can you also test that it works correctly across all repos? 

a simple test would involving changing something (like printwidth) in the shared config, running prettier and making sure that all the code across all repos would change. 

similarly, we should check that the tsconfig is also inherited correctly.

packages/flatbuffers/package.json
Pavan-Nambi: oh yaa got it

varunsrin:
@Pavan-Nambi still planning to finish this? had a few people asking to pick this up, if youre no longer able to work on it

varunsrin:
closing for now,

Issue #337, opened 2022-12-30
feat: message model constructors 


varunsrin:
**What is the feature you would like to implement?**

Methods that will construct a correctly signed MessageModel of a specific type given a set of parameters or a simple object. For example, calling `makeSignerAdd()` with the right parameters should produce a SignerAdd message correctly. This is a higher order abstraction over the Flatbuffer builders that also handles the correct hashing and signing of objects using the appropriate libraries. 

**Why is this feature important?**

Farcaster applications will model user actions as JSON objects which must be converted into a Flatbuffer representation before being sent to the Hub. By providing a standardized library in the ~/flatbuffers/ folder which is exported along with the RPC Client we can make it easy for clients to generate these messages correctly. 

**Will the protocol spec need to be updated??**

No

**How should this feature be built? (optional)**

Imagine that the consumer of this library is an application that hosts thousands of users and wants to generate messages for them. We would like the interface to: 

- Be as ergonomic as possible given the consumer's needs. For example, it might allow an instance to be create that remembers the signing key, so that it can be easily invoked with new parameters instead of having to respecify the key each time. 

- Be extremely strict, so that any additional parameters or data provided will fail noisily, which will make debugging easy. 

- Be thoroughly tested, down to the exact bytes being output so that we can make changes to the logic without worrying about about introducing subtle byte ordering bugs. 

- Be easily exportable, having no internal dependencies outside of the flatbuffers folder (external deps are OK). 

**Additional Context**

Flatbuffer factories already construct messages given a set of parameters, but this is not ideal outside of a testing environment. Some features like auto population of values when not supplied could lead to subtle and hard to debug errors, and they also have a lot of dependencies that we shouldn't add into production environments.

varunsrin:
@adaddeo this might be a good one for you to pick up if you're interested

adaddeo:
@varunsrin yes I'll pick it up

adaddeo:
Thought: since Farcaster applications will be working with objects from `flatbuffers/models` (i.e. `MessageModel`), it may be worthwhile to break out the RocksDB functionality to avoid having application code with a rocksdb dependency and objects that expose rocksdb specific methods.

varunsrin:
yes, i think thats a good call. 

tagging @pfletcherhill who may have opinions on how to do this

adaddeo:
closed by #355

PR #355, opened 2023-01-02
feat: introduce MessageBuilder

adaddeo:
**WIP** for #337


adaddeo:
Looking for early feedback on this before implementing methods for building other message types.

And specifically:

Validation: using `flatbuffers/models/validations` to perform validations seems _mostly_ good with a couple caveats:
- Some of the validations like checking the signature and timestamp don't make as much sense in the context of constructing messages but may not hurt. We could just use the message specific (i.e. `validateSignerMessage`) to validate user input rather than using `validateMessage` like I have it currently.
- The validation functions return `Result<T, HubError>` with error codes like `bad_request.validation_failure`. We could return more generic error codes from the validations module and have consumers add whatever additional context (i.e. `bad_request`). 

Signers: I'm think the API I have for setting up ECDSA and EsDSA signers is decent but I'm not completely satisfied. Still noodling on it but open to suggestions.

pfletcherhill:
Do you think you could split this work into two pieces: (1) a set of stateless functions to generate flatbuffer messages from JSON versions of the messages and (2) a class like the MessageBuilder one you have that retains the fid, signer, etc for ease of use? What's nice about splitting it is that the stateless functions can be used to implement the MessageBuilder class but they can also be used in tests across the rest of the codebase.

pfletcherhill:
Also for error handling, I think it's fine to send the generated object through `validateMessage`. I think that's better than effectively writing a parallel set of validations. And I don't see any issue with returning `HubResult<...>` from all these methods.

If you're excited about doing it, I think it would be great to refactor the `validateMessage` method so that it checks a bunch of stuff in parallel and returns an aggregate list of validation errors. The fact that the user gets a single error, then runs into another, and another could be annoying.

adaddeo:
@pfletcherhill yes makes sense, I'll split it. Thanks for the feedback re errors, will revisit `validateMessage` after I wrap this up

app/src/flatbuffers/messageBuilder/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Ergonomic, type-safe message constructor.
+
+## Usage
+
+Constructing a signer message with an ECDSA signature:
+
+```typescript
+const ethersSigner = new EthersMessageSigner(signer.wallet);
+const builder = new MessageBuilder({ fid, eip712Signer: ethersMessageSigner });
```
varunsrin, line 11: does builder always need an ethers signer? most messages just require a signer

app/src/flatbuffers/messageBuilder/messageBuilder.ts
```diff
@@ -0,0 +1,154 @@
...
+  fid: number;
+  eip712Signer?: IMessageSigner;
+  ed2559Signer?: IMessageSigner;
+  hasher?: IMessageHasher;
+  network?: flatbuffers.FarcasterNetwork;
+};
+
+class MessageBuilder implements IMessageBuilder {
+  /** litte endian representation of fid */
+  private fid: number[];
```
varunsrin, line 41: nit: might be helpful to call this `fidBytes` to distinguish from input parameter fid

app/src/flatbuffers/messageBuilder/messageBuilder.ts
```diff
@@ -0,0 +1,154 @@
...
+    const fidBytesResult = numberToLittleEndianBytes(fid);
+    // TODO consider how errors should be reported
+    if (fidBytesResult.isErr()) {
+      throw new Error('invalid fid');
+    }
+    this.fid = Array.from(fidBytesResult.value);
+    this.network = network;
+    this.eip712Signer = eip712Signer;
+    this.ed2559Signer = ed2559Signer;
+    this.hasher = hasher ?? new Blake3Hasher();
```
varunsrin, line 63: i like the flexibility, but given that we only have one viable hash scheme, should we just hardcode this? or is there an argument for using alternative hashers for perf or other reasons?

app/src/flatbuffers/messageBuilder/messageBuilder.ts
```diff
@@ -0,0 +1,154 @@
...
+    const body = new flatbuffers.SignerBodyT(Array.from(arrayify(options.publicKey)));
+    const model = await this.buildMessage(
+      flatbuffers.MessageBody.SignerBody,
+      body,
+      flatbuffers.MessageType.SignerAdd,
+      this.eip712Signer
+    );
+
+    if (options.privateKey) {
+      this.ed2559Signer = new Ed25519MessageSigner(arrayify(options.privateKey), options.publicKey);
```
varunsrin, line 86: i'm wondering how explicit vs implict we should be here. i like the flexibility of "give me a signer or i'll make one for you". but i worry that people might forget to export the signer for example and so its better to force them to provide one as input instead.  

a more explicit model might involve two separate builder classes, one for ECDSA signatures and the other for EDDSA signers, explicitly requiring a key pair to be passed in each time. @pfletcherhill what are your thoughts here?

app/src/flatbuffers/messageBuilder/messageBuilder.test.ts
```diff
@@ -0,0 +1,102 @@
...
+            new MessageBuilder({
+              fid: -1,
+              eip712Signer: ethersSigner,
+            })
+        ).toThrow();
+      });
+    });
+  });
+
+  describe('instance methods', () => {
```
varunsrin, line 42: one issue im worried about people introducing in the future is a symmetrical bug that affects both the encode/decode side. for example if we have an error that inverts the order of the fid while encoding and decoding, it would pass such tests. 

a more robust, but tedious, option would be to use a fixture that is externally verified so we know that there are no errors in encoding/decoding. @pfletcherhill you may have thought more about this already.

varunsrin (commented):
overall looks excellent! left a few smaller comments, and will defer to @pfletcherhill for final approval

app/src/flatbuffers/messageBuilder/messageBuilder.ts
adaddeo: If the single hash scheme is stable I'm in favor of hardcoding.

The other reason I had in my head was cross platform support but it looks like `@noble/hashes/blake3` library doesn't use native bindings so it will work across browser/node/react-native environments, so not a concern.

Given likely usage perf doesn't seem to be a concern either.

app/src/flatbuffers/messageBuilder/README.md
adaddeo: no, it can just take an `ed2559Signer` if no signer messages are needed

app/src/flatbuffers/messageBuilder/messageBuilder.test.ts
adaddeo: how about a module that defines functions for serializing/deserializing "FC scalars" (i.e. timestamp, fid, ts_hash) between their byte and preferred Javascript representations.

If there were well-tested functions like:
```
serializeFid(number: BigNumber): Uint8Array
deserializeFid(bytes: UInt8Array): BigNumber
```

Test cases could be written like:
```
const someFid = 123;
const builder = new MessageBuilder({ fid, ... });
const message = builder.makeSignerAdd(...);
expect(message.data.fidArray()).toEqual(serializeFid(someFid));
```

should catch these types of errors and isolate the tedium of externally verified checks to a single module

app/src/flatbuffers/messageBuilder/messageBuilder.ts
varunsrin: great, lets do that for now and we can always extend it later

pfletcherhill:
@adaddeo can you replace calls to `arrayify` and `hexlify` with the appropriate methods from the [bytes.ts file](https://github.com/farcasterxyz/hub/blob/main/app/src/flatbuffers/utils/bytes.ts)? I just merged a bunch of new methods there, because `arrayify` and `hexlify` default to big endian but all data inside the hub should be little endian. Happy to give more context/explanation too if this doesn't make sense.

adaddeo:
Initial feedback has been incorporated.

- `constructors` export stateless functions for building message models that all take `fid`, `signer`, and `network` along with whatever specific properties are required for building that message type (i.e. `publicKey` for `SignerAdd`)
- there are two classes for building messages, one for signer messages and one for all others
- updated to use byte functions

@pfletcherhill looking for any more feedback on the general pattern and once you're satisfied I'll go ahead and implement the rest of the message types

One question: should network be optional? If so what default is preferred?

pfletcherhill:
I've been thinking a fair amount about the various use-cases for this, and I think we should focus mostly on the third party dev UX rather than internal hub usage. While that would seem to mean lots of abstractions over flatbuffers, the most immediate need is a package that Merkle can import into their backend to create valid flatbuffer messages and send them to our internal hub. We want the API for building those messages to be incredibly straightforward, but they should still be able to retrieve the flatbuffer itself and save that in a database.

To that end, I think we should write a new class (or just a json type) for representing a message that includes both the deserialized values and flatbuffer. The `MessageModel` class has a lot of code that has to do with rocksdb persistence inside hubs that shouldn't be given to external devs. And `MessageModel` also doesn't do enough deserialization—we want devs to have access to [json versions of all the message data](https://github.com/pfletcherhill/farcaster-hub/blob/js-package/packages/js/src/types.ts). It could be as simple as this (see below), though it's probably worth making a class that validates the provided data and makes it readonly:
```
// types.Message from https://github.com/pfletcherhill/farcaster-hub/blob/js-package/packages/js/src/types.ts
type Message = types.Message & {
  fbb: flatbuffers.Message;
  tsHash: string; // tsHash isn't part of the message and has to be computed
}
```

Also, I don't think persistence of fid or signer between messages is actually that important for a dev. I imagine they want to be able to create messages for many fids and many signers, so I think we should export the stateless functions rather than the `MessageBuilder` class.

Here's the API I'm imagining, which is very similar to what you already have with the exception of the arguments to methods like `makeCastAdd` being the json versions of body and data (i.e. no byte arrays).
```
import { FarcasterClient, makeCastAdd, Ed25519Signer, FarcasterNetwork } from '@hub/js';

const client = new FarcasterClient(`<insert hub address and port>`);

const signer = new Ed25519Signer('<insert private key>');

const castAdd = await makeCastAdd({ text: 'hello world' }, { fid: 10, network: FarcasterNetwork.Testnet }, signer);

const response = await client.submitMessage(castAdd);

const castResponse = await client.getCast(10, castAdd.tsHash)

expect(castResponse).toEqual(castAdd)
```

Also note that the end result should be an external package (`@hub/js`), though the code can be written inside that package or in `@hub/utils` and re-exported from there.

adaddeo:
Ok this clarifies things. The use case I had in mind was a frontend client interacting directly w/ a hub, but given this use case I see the appeal of stateless function and returning objects with json + flatbuffer representations. Model classes won't be used outside of Hubs (I just saw grpc Client got refactored to work with flatbuffer representations directly).

I also see the `json` module for going between scalar and byte representations, very nice 👍 

I think we're on the same page. How would you like to proceed from here?

adaddeo:
Should the validations module also be refactored to work with `flatbuffers.Message` instead of `MessageModel`?

pfletcherhill:
Great, here's the branch I was experimenting with an approach like this: https://github.com/pfletcherhill/farcaster-hub/tree/message-builder

Feel free to pull into your message-builder branch and continue working on it or just use as inspiration

adaddeo:
> [message-builder](https://github.com/pfletcherhill/farcaster-hub/tree/message-builder)

will do!

PR #390, opened 2023-01-06
feat: Merge returns false if it is a no-op, true otherwise

adityapk00:
## Motivation

Return a boolean from merge, indicating if the merge resulted in a no-op. We need this so that gossip knows if the submitted message was new, and should be gossiped out, or if it already existed, so no need to gossip out. 

## Change Summary

- `merge` methods now return boolean, depending if it was a no-op
- Update tests

## Merge Checklist


pfletcherhill:
Closing in favor of https://github.com/farcasterxyz/hub/pull/428

PR #392, opened 2023-01-07
feat: replace Eip712Signer with EthersSigner

adaddeo:
## Motivation

The existing implementation of `Eip712Signer` requires using a private key. Clients using this will want to support a variety of hot and cold wallets (i.e. MetaMask, Ledger) that don't expose private keys.
 
## Change Summary

- introduce `EthersSigner` that works with an any `ethers.Signer` sub-class

Once this gets sign off I will remove `Eip712Signer` and replace usages with the EthersSigner w/ `ethers.Wallet`.

## Merge Checklist

### DIFF
--- a/packages/utils/src/signers/ethersSigner.ts
+++ b/packages/utils/src/signers/ethersSigner.ts
@@ -1,0 +1,37 @@
+import { SignatureScheme } from '@hub/flatbuffers';
+import ethers from 'ethers';
+import { hexStringToBytes } from '../bytes';
+import { eip712 } from '../crypto';
+import { HubAsyncResult } from '../errors';
+import { VerificationEthAddressClaim } from '../types';
+import { Signer } from './signer';
+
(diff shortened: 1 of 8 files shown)


packages/utils/src/signers/ethersSigner.ts
```diff
@@ -0,0 +1,37 @@
...
+import { HubAsyncResult } from '../errors';
+import { VerificationEthAddressClaim } from '../types';
+import { Signer } from './signer';
+
+export class EthersSigner extends Signer {
+  /** 32-byte wallet address */
+  public declare readonly signerKey: Uint8Array;
+  private readonly ethersSigner: ethers.Signer;
+
+  static async fromEthers(ethersSigner: ethers.Signer) {
```
adaddeo, line 14: alternative: change the type of `signerKey` and `signerKeyHex` to be functions that return `HubAsyncResult` wrapped values

PR #409, opened 2023-01-11
feat: add getIdRegistryEvent and getNameRegistryEvent in @farcaster/js

### LINKED ISSUE #403: feat: add support for /getIdRegistryEvent and /getNameRegistryEvent gRPC methods in @farcaster/js package

adaddeo:
## Motivation

Adds `getIdRegistryEvent` and `getNameRegistryEvent` to the @farcaster/js Client, closes #403.

## Change Summary

- adds `getIdRegistryEvent` and `getNameRegistryEvent` to the @farcaster/js Client
- adds `validateFname` and use it in `validateMessage`
- reorganized `utils`, grouping serialization functions by scalars, message body, message, and registry events

## Merge Checklist

### DIFF
--- a/packages/js/src/builders.ts
+++ b/packages/js/src/builders.ts
@@ -1,18 +1,16 @@
 import * as flatbuffers from '@farcaster/flatbuffers';
 import {
-  bytesToHexString,
   HubAsyncResult,
   HubError,
   HubResult,
   MessageSigner,
   Signer,
   toFarcasterTime,
-  toTsHash,
   validations,
 } from '@farcaster/utils';
 import { blake3 } from '@noble/hashes/blake3';
 import { Builder, ByteBuffer } from 'flatbuffers';
-import { err, ok, Result } from 'neverthrow';
+import { err, ok } from 'neverthrow';
 import * as types from './types';
 import * as utils from './utils';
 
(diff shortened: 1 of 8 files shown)


packages/js/src/builders.ts
```diff
@@ -61,47 +59,6 @@ const buildMakeMessage = <
   };
 };
 
-/** Generic Methods */
-
-export const makeMessageFromFlatbuffer = (flatbuffer: flatbuffers.Message): HubResult<types.Message> => {
```
adaddeo, line 66: moved this into `utils` under `deserializeMessage`, made sense given the call signature

packages/utils/src/factories.ts
```diff
@@ -31,7 +31,7 @@ const FnameFactory = Factory.define<Uint8Array>(() => {
   const builder = new Builder(8);
   // Add 8 random alphabets as the fname
   for (let i = 0; i < 8; i++) {
-    builder.addInt8(faker.datatype.number({ min: 65, max: 90 }));
+    builder.addInt8(faker.datatype.number({ min: 97, max: 122 }));
```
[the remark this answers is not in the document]
pfletcherhill: Let's use the full character set from the contract: https://github.com/farcasterxyz/contracts/blob/643540b3801742274b1052f022a6e0e3aea4d9de/src/NameRegistry.sol#L1063

packages/js/src/client.ts
```diff
@@ -2,30 +2,55 @@ import * as flatbuffers from '@farcaster/flatbuffers';
...
   serializeEd25519PublicKey,
   serializeEthAddress,
   serializeFid,
+  serializeFname,
   serializeTsHash,
   serializeUserId,
 } from './utils';
 
+type ClientResult<TJson, TFlatbuffer> = HubAsyncResult<
+  Readonly<
```
pfletcherhill, line 20: Nit: move this abstraction (`Readonly<TJson & ....>`) into types.ts

packages/js/src/client.ts
```diff
@@ -2,30 +2,55 @@ import * as flatbuffers from '@farcaster/flatbuffers';
...
+const deserializeCall = async <TJson, TFlatbuffer>(
+  call: HubAsyncResult<TFlatbuffer>,
+  deserialize: (fbb: TFlatbuffer) => HubResult<TJson>
+): ClientResult<TJson, TFlatbuffer> => {
+  const response = await call;
+
+  return response.andThen((flatbuffer) =>
+    deserialize(flatbuffer).map((type) => ({
+      ...type,
+      flatbuffer,
```
pfletcherhill, line 36: This is a little weird to me. Rather than only adding the flatbuffer here, I think we should update the type in types.ts to always have the flatbuffer. In fact, it probably makes sense for all types that correspond to a flatbuffer (i.e. `Message`, `MessageData`, `NameRegistryEvent`, and `IdRegistryEvent`) to include those flatbuffers as well for ease-of-use.

pfletcherhill (changes requested):
Looks good! A few minor comments to address before merging.

packages/js/src/client.ts
adaddeo: In my head I was thinking we have a json type, flatbuffer type, and functions to map between them, with a combined convenience type being a composition. Whereas if the flatbuffer representation is on our types, then any `serialize*` functions require the object to be serialized to already include a serialized representation.

But so long as we only do this for the root types and not things like bodies and CastId it won't be a problem.

packages/utils/src/validations.ts
```diff
@@ -431,13 +433,31 @@ export const validateUserDataAddBody = (body: flatbuffers.UserDataBody): HubResu
...
+      return err(validatedFname.error);
     }
   } else {
     return err(new HubError('bad_request.validation_failure', 'invalid user data type'));
   }
 
   return ok(body);
 };
+
+export const validateFname = (fname?: string | null): HubResult<string> => {
```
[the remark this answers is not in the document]
adaddeo: sure thing, can you clarify whether you're wanting this fn to support `string | byteArray` inputs if you want to it just operate on `byteArray`?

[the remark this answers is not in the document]
pfletcherhill: Ideally both, though if you'd prefer to just do one for now I'd say bytes because most of the other validations use bytes. It would be nice to support both serialized and deserialized inputs for most of these validations in the future.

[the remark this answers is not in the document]
adaddeo: got it, done

packages/js/src/client.ts
adaddeo: ahhh ok that helps clarify!

root + nested vs everything else

pfletcherhill (approved):
Thanks for making changes, looks good!

PR #411, opened 2023-01-11
chore: separate yarn build into a typecheck and transpile step

### LINKED ISSUE #346: chore: separate yarn build into a typecheck and transpile step

The monorepo should define two high level tasks - `yarn typecheck` which checks files without emitting types, and `yarn build` which is dependent on typecheck in `turbo.json` and only transpiles files using esbuild or swc. 

`yarn build` is often used locally to check for type errors but it outputs .js files which makes the process slow. the typecheck command offers a lighteweight alternative, and chaining it with build is useful in CI where want to both check files and transpile the code. 

No

deodad:
## Motivation

closes #346, give devs the ability to quickly typecheck the entire project or individual packages with `yarn typecheck`

## Change Summary

- adds `yarn typecheck` support to all packages
- updates apps/hub to be built using es-build (via `tsup-node`)

## Merge Checklist

## Additional Context

A new `packages/tsconfig.base.json` file is added that extends the root tsconfig file with two settings:
- `composite: true` this allows for incremental compilation across packages
- `paths` this allows the TS to look up types from the local source files rather than built distribution files, re-enables TS-aware code assists like Jump to Definition

Each package  also needs to specify `references` with any dependencies in its `tsconfig.json`.

While it's a bit annoying to have to maintain an additional layer of configuration, it should be well worth the benefit of a smooth development experience.

A note on the built hub code:

Running `node dist/cli.js start` fails just as it did previously with:
```
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /Users/deodad/Repos/hub/apps/hub/.config/hub.config.ts
```

It looks like the build is unused anyway since the start script using `tsx`.


varunsrin:
@deodad is this still relevant?

deodad:
> @deodad is this still relevant?

I do think setting up project references so that live type checking can be performed is meaningful DX improvement. Not urgent though.

PR #421, opened 2023-01-13
fix: docker builds are broken

### LINKED ISSUE #347: bug: docker builds are broken
**What is the bug?**

Docker builds are broken since the move to the monorepo structure which left the Docker files out of date

Run `docker build .` locally, which should fail

- https://turbo.build/repo/docs/handbook/deploying-with-docker
- https://github.com/vercel/turbo/tree/main/examples/with-docker

kcchu:
## Motivation

Fix #347 broken docker builds

## Change Summary

- Update Dockerfile according to monorepo structure
- Use `turbo prune` to optimize docker build

## Merge Checklist

## Additional Context

Currently, the Docker build attempt to run `yarn build`, which compiles Typescript files into Javascript. But in the last stage of the Dockerfile, TS code of the hub is copied into the image and it is run with `tsx` instead of the JS compiled files.

I believed it is done in this way because the hub uses tsconfig's path aliases (i.e. `compilerOptions.paths`) to map `~` to the project root but this mapping isn't carried into the JS files generated by `tsc`. Therefore, the compiled JS files of hub can't be run directly with node because node can't resolve the imports.

I tried to add wrapper scripts to override the default module resolution to follow tsx behavior. It works but the code is more complex than I would prefer. So, this PR keeps the existing behavior (i.e. using tsx).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,5 +31,6 @@ coverage/
 
 # Turborepo
 .turbo
+out
 
 dist
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,28 +2,46 @@
 # https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/docker/multi_stage_builds.md
 
 ###############################################################################
-############## Stage 1: Build the code using a full node image ################
+############## Stage 1: Create pruned version of monorepo #####################
+###############################################################################
+
+FROM node:18 AS prune
+
+USER node
+RUN mkdir /home/node/app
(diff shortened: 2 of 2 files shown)


sds:
Thank you, @kcchu!

This was on my list to tackle. Agree that (for now) the pragmatic solution is to continue to use `tsx`. I had looked into figuring out the module resolution with `~` and was similarly disappointed with how seemingly ugly it made things. 

Appreciate your contribution.

PR #422, opened 2023-01-13
test: add tests for deserializeIdRegistryEvents and deserializeNameRegistryEvents

adaddeo:
## Motivation

Tests were missing for `deserializeIdRegistryEvent` and `deserializeNameRegistryEvent`.

## Change Summary

- adds tests for `deserializeIdRegistryEvent` and `deserializeNameRegistryEvent`
- improve related factories
- organize functions in `utils.ts`

## Merge Checklist

### DIFF
--- a/packages/js/src/utils.test.ts
+++ b/packages/js/src/utils.test.ts
@@ -632,3 +632,121 @@ describe('deserializeMessage', () => {
     expect(message.signer).toEqual(ed25519Signer.signerKeyHex);
   });
 });
+
+describe('deserializeNameRegistryEvent', () => {
+  let nameRegistryEventFbb: flatbuffers.NameRegistryEvent;
+  let nameRegistryEvent: types.NameRegistryEvent;
+
+  beforeAll(async () => {
+    nameRegistryEventFbb = await Factories.NameRegistryEvent.create();
+    nameRegistryEvent = utils.deserializeNameRegistryEvent(nameRegistryEventFbb)._unsafeUnwrap();
+  });
+
+  test('flatbuffer', () => {
+    expect(nameRegistryEvent.flatbuffer).toEqual(nameRegistryEventFbb);
+  });
+
+  test('blockNumber', () => {
+    expect(nameRegistryEvent.blockNumber).toEqual(nameRegistryEventFbb.blockNumber());
+  });
+
+  test('blockHash', () => {
+    expect(nameRegistryEvent.blockHash).toEqual(
+      bytesToHexString(nameRegistryEventFbb.blockHashArray() ?? new Uint8Array())._unsafeUnwrap()
+    );
+  });
+
+  test('transactionHash', () => {
+    expect(nameRegistryEvent.transactionHash).toEqual(
+      bytesToHexString(nameRegistryEventFbb.transactionHashArray() ?? new Uint8Array())._unsafeUnwrap()
+    );
+  });
+
+  test('logIndex', () => {
+    expect(nameRegistryEvent.logIndex).toEqual(nameRegistryEventFbb.logIndex());
+  });
+
+  test('fname', () => {
+    expect(nameRegistryEvent.fname).toEqual(
+      utils.deserializeFname(nameRegistryEventFbb.fnameArray() ?? new Uint8Array())._unsafeUnwrap()
+    );
+  });
+
+  test('to', () => {
+    expect(nameRegistryEvent.to).toEqual(
+      bytesToHexString(nameRegistryEventFbb.toArray() ?? new Uint8Array())._unsafeUnwrap()
+    );
+  });
+
+  test('from', () => {
+    expect(nameRegistryEvent.from).toEqual(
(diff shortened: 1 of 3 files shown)


packages/js/src/utils.ts
```diff
@@ -18,146 +18,174 @@ import { err, ok, Result } from 'neverthrow';
...
 /* -------------------------------------------------------------------------- */
 
-export const deserializeFid = (fid: Uint8Array): HubResult<number> => {
-  return validations.validateFid(fid).andThen((fid) => bytesToNumber(fid));
-};
+export const deserializeNameRegistryEvent = (
+  flatbuffer: flatbuffers.NameRegistryEvent
+): HubResult<types.NameRegistryEvent> => {
+  const deserialized = Result.combine([
+    bytesToHexString(flatbuffer.blockHashArray() ?? new Uint8Array()),
```
pfletcherhill, line 28: Block hash (`size: 64`), transaction hash (`size: 64`), to/from (as eth addresses) should all be fixed size hex strings and likely have their own deserialize functions so we don't have to remember the lengths elsewhere in the code. We already have `deserializeEthAddress` that ensures that the deserialized eth address is 40 chars (plus 0x).

packages/js/src/utils.ts
```diff
@@ -18,146 +18,174 @@ import { err, ok, Result } from 'neverthrow';
...
 };
 
-export const serializeEthAddress = (ethAddress: string): HubResult<Uint8Array> => {
-  return hexStringToBytes(ethAddress).andThen((ethAddress) => validations.validateEthAddress(ethAddress));
-};
+export const deserializeIdRegistryEvent = (
+  flatbuffer: flatbuffers.IdRegistryEvent
+): HubResult<types.IdRegistryEvent> => {
+  const deserialized = Result.combine([
+    bytesToHexString(flatbuffer.blockHashArray() ?? new Uint8Array()),
```
pfletcherhill, line 67: Same comment as above about fixed size hex strings

packages/utils/src/factories.ts
```diff
@@ -63,6 +63,10 @@ const BlockHashFactory = Factory.define<string>(() => {
   return faker.datatype.hexadecimal({ length: 64, case: 'lower' });
 });
 
+const TransactionHashFactory = Factory.define<string>(() => {
```
pfletcherhill, line 66: Nit: we're mixing factories that make byte arrays and factories that make hex strings in the same file. Would be nice to either have a separate file for factories that produce deserialized data or be able to specify the output format for each factory. Not worth trying to change in this PR, but just something to think about.

pfletcherhill (changes requested):
Looks good, couple minor changes

packages/utils/src/factories.ts
adaddeo: hbu `Factories.TransactionHash.String.build()`? a little long but keeps things co-located

adaddeo:
@pfletcherhill feedback incorporated

packages/js/src/utils.test.ts
```diff
@@ -40,6 +40,60 @@ describe('serializeFid', () => {
...
+  const transactionHashBytes = hexStringToBytes(transactionHash)._unsafeUnwrap();
+
+  test(`succeeds`, () => {
+    expect(utils.deserializeTransactionHash(transactionHashBytes)).toEqual(ok(transactionHash));
+  });
+});
+
+describe('serializeEip712Signature', () => {
+  const signatureBytes = Factories.Bytes.build(undefined, { transient: { length: 65 } });
+  const signature = bytesToHexString(signatureBytes)._unsafeUnwrap();
```
pfletcherhill, line 72: Nit: consider adding `{size: 130}` here just to be extra explicit. Right now this will succeed because the bytes factory always produces unpadded byte arrays, but if that code ever changes then this test will start failing every once in a while.

packages/js/src/utils.test.ts
```diff
@@ -40,6 +40,60 @@ describe('serializeFid', () => {
...
+
+  test(`succeeds`, () => {
+    expect(utils.serializeEip712Signature(signature)).toEqual(ok(signatureBytes));
+  });
+});
+
+describe('serializeEip712Signature', () => {
+  const signatureBytes = Factories.Bytes.build(undefined, { transient: { length: 65 } });
+  const signature = bytesToHexString(signatureBytes)._unsafeUnwrap();
```
pfletcherhill, line 81: Same nit as above

packages/utils/src/factories.ts
pfletcherhill: Do you think that's better than `Factories.TransactionHash` (bytes) and `Factories.TransactionHashHex`?

pfletcherhill: Just added an issue for us to think about this more at some point https://github.com/farcasterxyz/hub/issues/425

pfletcherhill (approved):
A couple tiny nits, but leaving it up to you to address or not before merge

packages/js/src/utils.test.ts
adaddeo: ok finally have by head around this—all byte arrays are unpadded!

Issue #350, opened 2023-01-02
feat: add filter arguments to subscribe gRPC method


pfletcherhill:
Update the subscribe gRPC method to let users filter events and receive historical events.

Right now the events service only sends new events over gRPC, but if a process is subscribed to the service and then disconnects, they should have a way to catch up from the last event they saw (within reason). In order to implement this, hubs will have to maintain a queue of their events in order to retrieve historical ones.

Additionally, a process may only want `mergeMessage` or `mergeIdRegistryEvent` events, and they should be able to supply a filter for what they want.

PR #429, opened 2023-01-15
feat: add support for /subscribe gRPC method in @farcaster/js package

### LINKED ISSUE #402: feat: add support for /subscribe gRPC method in @farcaster/js package

adaddeo:
## Motivation

Closes #402, adds event type filtering to complete part of #350

## Change Summary

- exposes a `subscribe` method on `@farcaster/js` client
- add event filter to `subscribe`
- adds `deserializeEventResponse` to deserialize an EventResponse flatbuffer object into our JSON object types

## Merge Checklist

## Additional Context

Ideally the `subscribe` method in `@farcaster/js` would be wrapping the grpc call in some way and performing the deserialization. However, I was uncertain about the best way to do this without layering a potentially undesirable abstraction over the `ClientReadableStream` object being returned by the grpc client.

### DIFF
--- a/apps/hub/src/rpc/test/eventService.test.ts
+++ b/apps/hub/src/rpc/test/eventService.test.ts
@@ -33,15 +33,20 @@ afterAll(async () => {
 });
 
 const fid = Factories.FID.build();
+const fname = Factories.Fname.build();
 const ethSigner = Factories.Eip712Signer.build();
 const signer = Factories.Ed25519Signer.build();
 let custodyEvent: IdRegistryEventModel;
+let nameRegistryEvent: NameRegistryEventModel;
 let signerAdd: SignerAddModel;
 let castAdd: CastAddModel;
 
 beforeAll(async () => {
   custodyEvent = new IdRegistryEventModel(
     await Factories.IdRegistryEvent.create({ to: Array.from(ethSigner.signerKey), fid: Array.from(fid) })
+  );
+  nameRegistryEvent = new NameRegistryEventModel(
+    await Factories.NameRegistryEvent.create({ to: Array.from(ethSigner.signerKey), fname: Array.from(fname) })
   );
 
   const signerAddData = await Factories.SignerAddData.create({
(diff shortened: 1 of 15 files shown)


apps/hub/src/rpc/test/eventService.test.ts
```diff
@@ -61,40 +61,80 @@ beforeAll(async () => {
...
-      [EventType.MergeIdRegistryEvent, custodyEvent],
-      [EventType.MergeMessage, signerAdd],
-      [EventType.MergeMessage, castAdd],
-    ]);
+  describe('with type filters', () => {
+    let stream: ClientReadableStream<EventResponse>;
+    let events: [EventType, MessageModel | IdRegistryEventModel | NameRegistryEventModel][];
+
+    beforeEach(async () => {
+      stream = (await client.subscribe([EventType.MergeMessage]))._unsafeUnwrap();
```
pfletcherhill, line 108: Nit: can you share this code with the tests above under the `without type filters` describe?

pfletcherhill: And can you add a test with multiple types?

packages/js/src/utils.ts
```diff
@@ -17,6 +17,54 @@ import { ByteBuffer } from 'flatbuffers';
...
+      ).map((nameRegistryEvent) => {
+        return {
+          flatbuffer: fbb,
+          type,
+          nameRegistryEvent,
+        };
+      });
+    }
+    default:
+      return err(new HubError('unknown', `unknown EventType '${type}'`));
```
pfletcherhill, line 64: I'd say this is `bad_request.invalid_param` rather than `unknown`, because we know the problem is that the given event has an invalid type

packages/flatbuffers/src/schemas/rpc.fbs
```diff
@@ -170,5 +170,7 @@ table GetTrieNodesByPrefixRequest {
 
 // Events Requests
 
-table SubscribeRequest {}
+table SubscribeRequest {
+  types: [EventType];
```
pfletcherhill, line 174: How do you feel about just `types` vs `event_types` to be extra explicit?

pfletcherhill (approved):
Looks great! Left a few minor comments for you to consider before merge.

packages/flatbuffers/src/schemas/rpc.fbs
adaddeo: `event_types` is better

apps/hub/src/rpc/test/eventService.test.ts
adaddeo: refactored and added!

I tried to add a test that subscribes to all event types so we can test they all get setup properly, but I wasn't getting any events back for `PruneMessage` or `RevokeMessage`.

I was trying:
```
      await engine.revokeMessagesBySigner(fid, signer.signerKey);
```
and 
```
      await engine.pruneMessages(fid);
```

pfletcherhill: When you called `revokeMessagesBySigner` or `pruneMessages`, are you sure there were merged messages that were available to revoke (i.e. signed by the given signer) or prune (i.e. with a timestamp earlier than the cutoff for the given store)?

PR #432, opened 2023-01-16
chore: refactor functions in `packages/*` to return HubResult

adaddeo:
## Motivation

moves #213 forward by refactoring @farcaster/utils to use HubResult

## Change Summary

- removes all usages of `throw` from `packages/*`

## Merge Checklist

### DIFF
--- a/packages/js/src/builders.ts
+++ b/packages/js/src/builders.ts
@@ -108,12 +108,15 @@ const makeMessageData = (
   }
 
   const timestamp = toFarcasterTime(dataOptions.timestamp || Date.now());
+  if (timestamp.isErr()) {
+    return err(timestamp.error);
+  }
 
   const dataT = new flatbuffers.MessageDataT(
     bodyType,
     bodyT,
     messageType,
-    timestamp,
+    timestamp.value,
     Array.from(serializedFid.value),
     dataOptions.network
   );
--- a/packages/utils/src/signers/eip712Signer.ts
+++ b/packages/utils/src/signers/eip712Signer.ts
@@ -20,15 +20,15 @@ export class Eip712Signer implements Signer {
 
   private readonly _typedDataSigner: TypedDataSigner;
 
-  constructor(typedDataSigner: TypedDataSigner, address: string) {
+  public static fromSigner(typedDataSigner: TypedDataSigner, address: string) {
+    const signerKeyHex = address.toLowerCase();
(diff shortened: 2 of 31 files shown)


packages/utils/src/signers/eip712Signer.ts
```diff
@@ -20,15 +20,15 @@ export class Eip712Signer implements Signer {
 
   private readonly _typedDataSigner: TypedDataSigner;
 
-  constructor(typedDataSigner: TypedDataSigner, address: string) {
+  public static fromSigner(typedDataSigner: TypedDataSigner, address: string) {
```
pfletcherhill, line 23: I don't necessarily disagree with this change, but what's your thinking/motivation behind it?

packages/js/src/builders.ts
```diff
@@ -102,18 +102,27 @@ const makeMessageData = (
   messageType: flatbuffers.MessageType,
   dataOptions: MessageDataOptions
 ): HubResult<flatbuffers.MessageData> => {
+  // const dataT = Result.combine([
```
pfletcherhill, line 105: Get rid of commented out code

pfletcherhill (commented):
Generally looks good. Address the merge conflicts and a couple minor comments below, and it should be good to merge.

packages/js/src/builders.ts
adaddeo: 😲 should have caught this my bad

packages/utils/src/signers/eip712Signer.ts
adaddeo: Constructors can't return a `HubResult`.

The current approach is to `throw` if an operation fails in the constructor.

The approach in this PR is to refactor constructors so that they always succeeds and move operations that can fail into a static factory function that returns a `HubResult` with the constructed object in it.

FYI most of the `*Model` classes in the `hub` are throwing in the constructor, so it's worth considering whether this is a pattern we want to adopt.

The alternatives I considered were:
- leave the `throw` and just say "constructors are one place where throw is ok and we should remember to handle failures"
  - having the exception to the rule doesn't really buy us anything, the complexity still shows up wherever the objects are constructed, and has the other drawbacks discussed in #213 that motivated using neverthrow
- allow for "invalid" objects to be created and push the error handling into other parts of the object
  - can be useful pattern in some places but is heavy handed as a default approach

PR #444, opened 2023-01-20
docs: add TSDoc annotations to time functions and properties

deodad:
## Motivation

Add hints for developers about what format different timestamp values are in.

## Change Summary

- add TSDoc comments to places in the public API that deal with time

## Merge Checklist

### DIFF
--- a/packages/js/src/types.ts
+++ b/packages/js/src/types.ts
@@ -24,6 +24,7 @@ export type MessageData<TBody = MessageBody, TType = protobufs.MessageType> = {
   _protobuf: protobufs.MessageData;
   body: TBody;
   type: TType;
+  /** Unix milleseconds */
   timestamp: number;
   fid: number;
   network: protobufs.FarcasterNetwork;
--- a/packages/js/src/builders.ts
+++ b/packages/js/src/builders.ts
@@ -19,6 +19,7 @@ import * as utils from './utils';
 type MessageDataOptions = {
   fid: number;
   network: protobufs.FarcasterNetwork;
+  /** Unix milliseconds */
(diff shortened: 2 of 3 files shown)


packages/js/src/types.ts
```diff
@@ -24,6 +24,7 @@ export type MessageData<TBody = MessageBody, TType = flatbuffers.MessageType> =
   flatbuffer: flatbuffers.MessageData;
   body: TBody;
   type: TType;
+  /** unix milliseconds timestamp */
   timestamp: number;
```
adaddeo, line 28: having this be the Farcaster time probably makes more sense than unix milliseconds

@pfletcherhill wyt?

pfletcherhill (commented):
My assumption was that the deserialized timestamp should be the same format as the input to the builder. That way a user of the package doesn't necessarily need to know what farcaster time is. Am I missing something?

adaddeo:
> My assumption was that the deserialized timestamp should be the same format as the input to the builder. That way a user of the package doesn't necessarily need to know what farcaster time is. Am I missing something?

My thinking was the `timestamp` value of `MessageData` being in the [protocol specified format](https://github.com/farcasterxyz/protocol#timestamps) is the most straightforward. There is a Merkle usage that assumes this is the case, which doesn't seem unreasonable given nothing says otherwise.

Fair point about the builder though. Either way seems fine as long was it's well-documented. Let me know which you prefer and I'll update the TSDoc comments accordingly.

adaddeo:
Sounds good. Updated the PR to just add TSDoc annotations specifying the timestamp format.

PR #457, opened 2023-01-26
feat: Stub all RPC methods

adityapk00:
## Motivation

Definitions for all RPC methods

## Change Summary

- Stubs for all the RPC methods
- Rename service to 'HubService' instead of 'SyncService' since it now has all the Hub's RPC methods 

## Merge Checklist


packages/protobufs/src/schemas/message.proto
```diff
@@ -49,6 +49,10 @@ enum UserDataType {
   USER_DATA_TYPE_FNAME = 6;
 }
 
+message UserId {
```
pfletcherhill, line 52: Why bring back `UserId`? I've been referring to amps just by target fid.

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -34,7 +36,100 @@ message SyncIds {
...
+}
+
+message ReactionsByCastRequest {
+    CastId cast_id = 1;
+    ReactionType reaction_type = 2;
+}
+
+message AmpRequest {
+    uint64 fid = 1;
+    UserId user_id = 2;
```
pfletcherhill, line 65: See comment above about `UserId`. This should just be `uint64 target_fid = 2`

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -34,7 +36,100 @@ message SyncIds {
...
+    uint64 fid = 1;
+    bytes address = 2;
+}
+
+message SignerRequest {
+    uint64 fid = 1;
+    bytes signer = 2;
+}
+
+service HubService {
```
pfletcherhill, line 87: How do you feel about splitting this into `StorageService` and `SyncService` (naming tbd) to represent the fact that many clients may want to implement only the storage service and not deal with hub-to-hub methods?

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -34,7 +36,100 @@ message SyncIds {
...
+    // Submit Methods
+    rpc SubmitMessage (Message) returns (Message);
+    rpc SubitIdRegistryEvent (IdRegistryEvent) returns (IdRegistryEvent);
+    rpc submitNameRegistryEvent (NameRegistryEvent) returns (NameRegistryEvent);
+
+    // Casts
+    rpc GetCast(CastId) returns (Message);
+    rpc GetCastsByFid(FidRequest) returns (stream Message);
```
pfletcherhill, line 95: We initially switched from using grpc streams to using unary calls for queries like this. Let me try to find the discussion and see if the reasons are still relevant.

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -34,7 +36,100 @@ message SyncIds {
...
+
+    // Verifications
+    rpc GetVerification(VerificationRequest) returns (Message);
+    rpc GetVerificationsByFid(FidRequest) returns (stream Message);
+
+    // Signer
+    rpc GetSigner(SignerRequest) returns (Message);
+    rpc GetSignersByFid(FidRequest) returns (stream Message);
+    rpc GetCustodyEvent(FidRequest) returns (IdRegistryEvent);
+    rpc GetFids(Empty) returns (FidsResponse);
```
pfletcherhill, line 122: Any reason this is a unary call but other multi-response queries are streams?

packages/protobufs/src/schemas/rpc.proto
adityapk00: This is a good idea. I'll split it in a new PR.

Issue #445, opened 2023-01-20
feat: Benchmark Sync + Gossip Performance


adityapk00:
**What is the feature you would like to implement?**
A test that can benchmark the performance of Sync and Gossip in the Hub.

**Why is this feature important?**
Hubs stay in sync through the use of Gossip (for recent messages) and Sync (for older messages). While we currently have tests that check correctness, we need a way to profile the performance of this to ensure 1) No regressions 2) Estimate how much compute/bandwidth will be required for Hub operators to stay in sync

**Will the protocol spec need to be updated??**
No

**How should this feature be built? (optional)**
Ideally this is a test or script that can be run both on a local machine and in CI that can measure Sync and Gossip performance
1. Test sets up a network of Hubs on the test machine (each running a separate rocksDB and port) and connects them via libp2p
2. Adds generated messages to one or more hubs 
3. Ensure all hubs converge to the same rootHash, fail otherwise
4. Measure the number of messages that were synced. Metrics we want to track are: 
   a. Gossip messages/sec
   b. Sync messages/sec
   c. Wall clock time to sync
   d. (Optional): Bandwidth used by each Hub, Amount of overhead gossip traffic, amount of overhead sync traffic. 

**Additional context**
- There is a script here (https://github.com/farcasterxyz/hub/blob/main/apps/hub/src/test/perf/scenario.ts) that can generate a variety of messages and push them to the hub.
- Optionally add a network latency delay that can be configured to emulate real-world conditions
- Right now, we're looking for about 10-20 hubs to stay in sync in a performant way, but obviously this will increase in the future.

PR #470, opened 2023-01-29
feat: add benchmark suite for MerkleTrie

kcchu:
## Motivation

Add benchmarking scripts for MerkleTrie (#445)

## Change Summary

The change add a simple cli script that run benchmarking suites. The command takes parameters that can be useful for experimenting with different configuration.

An initial benchmark for MerkleTrie is implemented. It measures the time and memory usage of insert and getSnapshot operations in the MerkleTrie implementation.

To run the benchmark, run `yarn bench` under `apps/hubble` folder.

```
$ yarn bench -h
Usage: farcaster-bench --benchmark <suite> [options]

Farcaster unit benchmark suites

Options:
  -V, --version              output the version number
  -b --benchmark <suite>     the benchmarking suite to run
  -n, --count <count>        size of input
  -c, --cycle <cycles>       run at least <cycle> times before taking measurements
  -o, --output <file>        path of the output file (default STDOUT)
  -s, --write-heap-snapshot  write V8 heap snapshot after each cycle (default: false)
  -h, --help                 display help for command
```

For example, running a MerkleTrie benchmark of 5M items:

```
yarn bench -b merkletrie -n 5000000 -o output.csv
```

The `--write-heap-snapshot`  write V8 heap snapshot after each cycle which is useful for debugging memory leak.

## Merge Checklist

## Additional Context

Some preliminary benchmark results can be found here: https://gist.github.com/kcchu/03c07ce4c7a582c04bb6678e3fa8b275

### DIFF
--- a/apps/hubble/src/test/bench/helpers.ts
+++ b/apps/hubble/src/test/bench/helpers.ts
@@ -1,0 +1,50 @@
+import Chance from 'chance';
+
+import { Message } from '@farcaster/protobufs';
+import { SyncId } from '~/network/sync/syncId';
+
+/**
+ * Generate `n` SyncId. The SyncIds are sampled from FIDs between 1 to `numFids`.  Timestamp in
+ * returned SyncIds are strictly increasing by a uniformly random number between 1 to `maxTsDelta`.
+ * The generated result set is repeatable.
+ *
+ * @param n the total number of SyncId
+ * @param numFids number of FIDs
+ * @param maxTimeShift generate timestamp between 1 to maxTsDelta
+ */
+export const generateSyncIds = (n: number, numFids = 1, maxTimeShift = 1): SyncId[] => {
+  const chance = new Chance(1); // use constant seed for repeatable data
+  const fids: number[] = [];
+  for (let i = 1; i <= numFids; i++) {
+    fids.push(i);
+  }
+  const syncIds: SyncId[] = [];
+  let ts = 1;
+  for (let i = 0; i < n; i++) {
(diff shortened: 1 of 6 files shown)


adityapk00:
The `apps/hubs` is about to go away. Please put these scripts in `apps/protohub`, which uses the protobufs. Overall, the direction looks good to me!

apps/hubble/src/test/bench/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+  const chance = new Chance(1); // use constant seed for repeatable data
+  const fids: number[] = [];
+  for (let i = 1; i <= numFids; i++) {
+    fids.push(i);
+  }
+  const syncIds: SyncId[] = [];
+  let ts = 1;
+  for (let i = 0; i < n; i++) {
+    const fid = chance.pickone(fids);
+    const hash = chance.hash({ length: 20 });
```
adityapk00, line 25: this should be `length: 40`. Hashes are 20 bytes, 40 hex bytes

adityapk00 (approved):
It is curious that this implementation seems to be using more RSS (although the same amount of Heap). I wonder if the overhead is coming from the performance measure library itself?

As an experiment, I ran this script, but removed all the benchmarking and measuring code, and it was able to fit 3.5M SyncIDs into memory at the default memory size (which I believe is 4GB for node), which makes me think the perf measurement stuff in node might be causing the extra usage?

 It would be worthwhile to track this down.

Issue #473, opened 2023-01-30
feat: Move the protobuf RPC implementation into it's own package


adityapk00:
**What is the feature you would like to implement?**
Move the RPC service definition and implementation out of `packages/protobufs into it's own package

**Why is this feature important?**
Right now, the `packages/protobufs` contains both the proto definitions and the gRPC service definitions, which bring in grpc-js. It would be nice to have the `protobufs`  package to just contain the protobuf definitions, which will make it easier for clients to depend on instead of pulling in all the grpc dependencies. 

**Will the protocol spec need to be updated??**
No

**How should this feature be built? (optional)**

**Additional context**

payton:
Hey @adityapk00 is this still of interest?

PR #500, opened 2023-02-02
feat: Move the protobuf RPC implementation into it's own package

outricked:
## Motivation

https://github.com/farcasterxyz/hubble/issues/473

## Change Summary

- Created a new package to hold RPC implementation called grpc. Removed all grpc requirements from protobuf package. Redirected all imports to grpc.

## Merge Checklist

## Additional Context


outricked:
@adityapk00 Created a pull request to address issue #473. Would love a review!

adityapk00:
This is pretty cool! Thank you. 

The one comment I have is can you have it such that the `packages/js` and `packages/utils` import the protobufs from the `protobufs` package instead of the new `grpc` package? Only the `apps/hubble` package should depend on the new `grpc` package. 

The reason for this is that the `grpc` stuff only works in a node environment, but we want the `packages/js` and `packages/utils` to be usable in a browser environment too, so it should not pull in any grpc code. 

Somewhat specifically, the problem is that to build the grpc stubs, we need the additional ESBuild config (currently in `tsup.config.ts` file, which overrides the build with the additional imports in the `banner` property. This doesn't work in a browser environment. (see https://github.com/farcasterxyz/hubble/issues/494)

outricked:
One issue with the ask is that utils use a lot of things from the grpc package.
Specifically: all the imports [here](https://github.com/farcasterxyz/hubble/blob/a25c8c9654bf7e7476d52335db41023701843ee4/packages/utils/src/client.ts#L1) require stuff from the grpc library. Not sure how I can handle that without changing the utils file significantly

adityapk00:
That makes sense. How about we just do the `packages/js` for the first pass. That should help with the apps in the browser. 

I suspect to move the utils package, we'll have to also add grpc-web, (since grpc doesn't work in the browser anyway)

varunsrin:
@outricked you'll need to sign your commits and might need to push the branch up again. see here for more details: https://github.com/farcasterxyz/hubble/blob/main/CONTRIBUTING.md#22-signing-commits

pfletcherhill:
Right now I think `@farcaster/js` still indirectly depends on `@grpc/grpc-js` because it includes its own gRPC client for submitting messages and querying hubs. The package also uses BigInt via `@noble/hashes`, which causes more compatibility issues with react native. So my take is that this problem will take more spec work ([see this ticket](https://github.com/farcasterxyz/hubble/issues/497)) and we should hold off on implementing something right now.

pfletcherhill (commented):
Right now I think @farcaster/js still indirectly depends on @grpc/grpc-js because it includes its own gRPC client for submitting messages and querying hubs. The package also uses BigInt via @noble/hashes, which causes more compatibility issues with react native. So my take is that this problem will take more spec work (https://github.com/farcasterxyz/hubble/issues/497) and we should hold off on implementing something right now.

outricked:
Agreed. I think it's a bigger overhaul to actually fully separate grpc. RIP

packages/grpc/src/typeguards.ts
```diff
@@ -0,0 +1,144 @@
+import * as protobufs from '@farcaster/protobufs/src/generated/message';
```
adaddeo, line 1: any reason to include `types.ts` and `typeguards.ts` here instead of in `protobufs` where they could be used w/o needing to add grpc-js dependency?

PR #520, opened 2023-02-04
chore: Test gRPC server is running in Docker image

sds:
## Motivation

Ensure we know if we accidentally break the gRPC service for the hub.

## Change Summary

Add a simple test to our build step that ensures the gRPC service is running.

## Merge Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,31 @@ jobs:
         uses: docker/setup-buildx-action@v2
 
       - name: Build Docker image
+        id: docker-image
         uses: docker/build-push-action@v4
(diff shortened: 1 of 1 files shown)


sds:
This isn't working because our hub implementation waits until it has performed an initial sync before starting the gRPC server. We should probably start the gRPC server right away.

varunsrin:
@sds this still relevant to keep open?

sds:
@varunsrin I think so since it allows us to have a basic smoke test as part of every hub PR.

The ask I had of the team was to update the hubs to start the RPC server ASAP, rather than waiting for all ID registry events to be fetched.

PR #535, opened 2023-02-07
fix: incorrect prefix used by SyncEngine.getSnapshot()

kcchu:
## Motivation

SyncEngine.getSnapshot() divides timestamp by 10 inadvertently and makes the prefix to be shifted right, causing the prefix to be mismatch with with messages in MerkleTrie.

## Change Summary

Change SyncEngine.getSnapshot() to not divide timestamp by 10.

## Merge Checklist

## Additional Context

The function intents to use a 10-seconds resolution timestamp to give the messages a chance to settle before they are picked up by the sync algorithm. The timestamp rounding logic is implemented in `snapshotTimestamp` and dividing timestamp by 10 in SyncEngine.getSnapshot() is unnecessary and incorrect.


adityapk00 (approved):

adityapk00:
Thanks! Merged.

PR #562, opened 2023-02-12
feat: grpc api documentation

varunsrin:
## Motivation

Documenting the protobufs and gRPC endpoints will make it easier for developers to work with Hubs

## Change Summary

Adding Markdown files that contain documentation for both

## Merge Checklist

### DIFF
--- a/packages/js/src/client.ts
+++ b/packages/js/src/client.ts
@@ -43,17 +43,17 @@ export class Client {
     this._grpcClient = getHubRpcClient(address);
   }
 
-  // /* -------------------------------------------------------------------------- */
-  // /*                                Submit Methods                              */
-  // /* -------------------------------------------------------------------------- */
+  /* -------------------------------------------------------------------------- */
+  /*                                Submit Methods                              */
+  /* -------------------------------------------------------------------------- */
 
   async submitMessage(message: types.Message): HubAsyncResult<types.Message> {
     return wrapGrpcMessageCall(this._grpcClient.submitMessage(message._protobuf));
   }
 
-  // /* -------------------------------------------------------------------------- */
-  // /*                                 Cast Methods                               */
-  // /* -------------------------------------------------------------------------- */
+  /* -------------------------------------------------------------------------- */
+  /*                                 Cast Methods                               */
+  /* -------------------------------------------------------------------------- */
 
   async getCast(fid: number, hash: string): HubAsyncResult<types.CastAddMessage> {
     const castId = utils.serializeCastId({ fid, hash });
@@ -83,9 +83,14 @@ export class Client {
     return wrapGrpcMessagesCall(this._grpcClient.getCastsByMention(fidRequest));
   }
 
-  // /* -------------------------------------------------------------------------- */
-  // /*                                Amp Methods                              */
-  // /* -------------------------------------------------------------------------- */
+  async getAllCastMessagesByFid(fid: number): HubAsyncResult<(types.CastAddMessage | types.CastRemoveMessage)[]> {
+    const request = protobufs.FidRequest.create({ fid });
+    return wrapGrpcMessagesCall(this._grpcClient.getAllCastMessagesByFid(request));
+  }
+
+  /* -------------------------------------------------------------------------- */
+  /*                                Amp Methods                              */
+  /* -------------------------------------------------------------------------- */
 
   async getAmp(fid: number, targetFid: number): HubAsyncResult<types.AmpAddMessage> {
     const ampRequest = protobufs.AmpRequest.create({ fid, targetFid });
@@ -102,9 +107,14 @@ export class Client {
     return wrapGrpcMessagesCall(this._grpcClient.getAmpsByUser(fidRequest));
   }
 
-  // /* -------------------------------------------------------------------------- */
-  // /*                               Reaction Methods                             */
-  // /* -------------------------------------------------------------------------- */
+  async getAllAmpMessagesByFid(fid: number): HubAsyncResult<(types.AmpAddMessage | types.AmpRemoveMessage)[]> {
+    const request = protobufs.FidRequest.create({ fid });
+    return wrapGrpcMessagesCall(this._grpcClient.getAllAmpMessagesByFid(request));
+  }
+
+  /* -------------------------------------------------------------------------- */
+  /*                               Reaction Methods                             */
+  /* -------------------------------------------------------------------------- */
 
   async getReaction(
     fid: number,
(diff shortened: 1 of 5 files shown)


varunsrin:
A few issues that came up while writing the documentation: 

1. Is there any difference between `GetAllUserDataMessagesByFid` and `GetUserDataByFid` ?

2. Should we remove Amps for now? It is possible that we may change implementation details or not ship this feature for a long time, and supporting it at the protocol level first might lock us in to a design. 

3. VerificationAddEthAddressMessage is 32 chars and long and unwieldy, especially in documentation. Consider a shorter name across the protocol like `VerificationAddEthMessage`

4. There are some differences in type specificity: the gRPC endpoints define the return value generically as `repeated Message` while the farcaster/js methods are much more specific like `CastAddMessage[]`. In practice, the specificity is valid because the code underneath calls the right methods on the engine and this is probably safe to ignore.

varunsrin:
This PR specifically does not document the farcaster/js methods which will be done in a follow on PR

varunsrin:
Remaining items 

- [ ] Add entrypoint in README to navigate to docs 
- [ ] Update after resolving issues with Amps, Location
- [ ] Try to auto-generate docs from rpc and comment descriptions
- [ ] Add high level descriptions to Services

packages/protobufs/docs/services.md
```diff
@@ -0,0 +1,166 @@
...
+| -------------------- | ------------------------ | ----------------- | ------------------------------------------------------ |
+| GetFids              | Empty                    | FidsResponse      | Returns the most recent Fids that were registered      |
+| GetIdRegistryEvent   | FidRequest               | IdRegistryEvent   | Returns the most recent IdRegistryEvent for an Fid     |
+| GetNameRegistryEvent | NameRegistryEventRequest | NameRegistryEvent | Returns the most recent NameRegistryEvent for an Fname |
+
+#### NameRegistryEvent Request
+
+| Field | Type       | Label | Description |
+| ----- | ---------- | ----- | ----------- |
+| name  | [bytes](#) |       |             |
```
adityapk00, line 160: This is used in the gRPC to represent an fname, to request data by fname

packages/protobufs/docs/services.md
```diff
@@ -0,0 +1,166 @@
...
+
+#### Fids Response
+
+| Field | Type        | Label    | Description |
+| ----- | ----------- | -------- | ----------- |
+| fids  | [uint64](#) | repeated |             |
```
adityapk00, line 166: This is used in the gRPC to represent an FID, to request data by FID. This is needed because gRPC doesn't allow you to specify just a uin64 as a argument type, it has to be a protobuf.

adityapk00 (approved):

adityapk00:
> Is there any difference between GetAllUserDataMessagesByFid and GetUserDataByFid
They are doing the exact same thing. I think both are there for consistency with the rest of the API, but we'll be OK to get rid of one of them. 

> There are some differences in type specificity: the gRPC endpoints define the return value generically as repeated Message while the farcaster/js methods are much more specific like CastAddMessage[]

At least some of the methods - Like the `GetAllMessagesBySyncIds` will return different types of messages in the same response, so we need to use `repeated Message` instead of a specific `CastAddMessage`

For the rest, like `GetCastsByFid`, we could potentially change it to be `repeated CastAddMessage`

pfletcherhill:
> Is there any difference between GetAllUserDataMessagesByFid and GetUserDataByFid ?

Adding to what Aditya said above...while they are functionally the same, I have a slight preference towards keeping them both for consistency with the rest of the API. The `GetAll...` methods are for clients looking to replicate state of the hub and want to get both adds and removes, so it's convenient to have a full set of `GetAll...` methods rather than having to know that you should use `GetUserDataByFid` for the UserDataStore instead. If there's a clear cost to having both, then I'm happy to remove one though.

> Should we remove Amps for now? It is possible that we may change implementation details or not ship this feature for a long time, and supporting it at the protocol level first might lock us in to a design.

I'm open to either, but I'm a bit concerned about removing it from the protocol and then not having much incentive to ship it at all. Having it in the protocol is a clear signal to the community that while we're not publishing follows, we're going to add amps imminently. This could also be a great opportunity for another team in the community to build a client for amps to take the pressure off Merkle.

> VerificationAddEthAddressMessage is 32 chars and long and unwieldy, especially in documentation. Consider a shorter name across the protocol like VerificationAddEthMessage

I don't have a strong opinion either way, though while the name is long, it does correctly represents the concept. Other than annoyance to write, what's the cost of a long name? 

> There are some differences in type specificity: the gRPC endpoints define the return value generically as repeated Message while the farcaster/js methods are much more specific like CastAddMessage[]. In practice, the specificity is valid because the code underneath calls the right methods on the engine and this is probably safe to ignore.

It's worth mentioning that `CastAddMessage` is just a type and not a value defined in the Message protobuf schema, so I don't think we'd be able to be more specific in the protobuf schema. We could be less specific with the js package and grpc methods, but it's very useful to be able to type those return values more tightly. So I'm in favor of keeping it the way it is.

> Same problem as 2, but with Location User Data type

The problem is that the Merkle team hasn't yet shipped the location feature, so we want to hold off adding it until we do? Would the issue be that we're not sure it should be a UserDataAdd vs some sort of Cast? Or the string length?

> VerificationRemove appears generic compared to VerificationAddEthereumAddress, but it contains an address property which binds it only address based verifications. Should we call it VerificationRemoveEth` and scope verifications to always point to keypairs?

Yeah, I'm happy to change it to `VerificationRemoveEthAddress`. Alternatively, we could just call them both `VerificationAdd` and `VerificationRemove` now and know that for this version they're referring to eth addresses.

packages/protobufs/docs/protos.md
```diff
@@ -0,0 +1,198 @@
...
+| network   | [FarcasterNetwork](#FarcasterNetwork)                                                                                                                                                                                                                                       |       | Farcaster network the message is intended for  |
+| body      | [CastAddBody](#CastAddBody), [CastRemoveBody](#CastRemoveBody), [ReactionBody](#ReactionBody), [VerificationAddEthAddressBody](#VerificationAddEthAddressBody), [VerificationRemoveBody](#VerificationRemoveBody), [SignerBody](#SignerBody), [UserDataBody](#UserDataBody) | oneOf |                                                |
+
+### 1.3 HashScheme
+
+Type of hashing scheme used to produce a digest of the message&#39;s data
+
+| Name               | Number | Description                             |
+| ------------------ | ------ | --------------------------------------- |
+| HASH_SCHEME_NONE   | 0      |                                         |
```
pfletcherhill, line 41: Do you think we should include the NONE values in the docs? Constructing a message with any of them will result in an invalid message.

pfletcherhill:
Can you add a changeset since you're updating packages?

packages/protobufs/docs/protos.md
varunsrin: let me look at the conventions for other apis, the code generators seem to produce them by defaylt

varunsrin:
Responding to @pfletcherhill's comments: 

1. On `GetAllUserDataMessagesByFid` and `GetUserDataByFid` it's confusing when reading docs because you assume that something with an All in the name must be different than something without it. I understand your concern, but I wonder if there is another better way to express this that minimizes confusion. 

2. On Amps and location, my experience from building casts and likes is that building a working prototype leads to uncovering issues with the data format. Since we have not yet implemented either in any client, and since data format upgrades might be expensive, let's bias to adding them post prototyping. 

3. On `VerificationAddEthereumMessage` the verbosity makes several things tedious (e.g. markdown tables for the documentation are less readable). The name can be shortened without losing meaning, so I'm supportive of `VerificationAddEth/VerificationRemoveEth`.  If you strongly prefer `VerificationAdd/VerificationRemove` I'm supportive but with the caveat that we will likely rename this again in the next year since we will most likely have other verification types. 

4. The changeset repo seemed to indicate that this should be done only for meaningful changes, and this PR has no changes to the functionality of the app. What's our philosophy on when a changeset is needed vs when it isn't?  

Screen Shot 2023-02-17 at 10 38 42 AM

packages/protobufs/docs/protos.md
varunsrin: discussing and handling in follow on pr

varunsrin:
This PR is ready for merge, items to be resolved in follow ons: 

1. Naming for `VerificationAddEthereumMessage`
2. Duplication of functionality between `GetAllUserDataMessagesByFid` and `GetUserDataMessagesByFid`
3. Removal of Amps and Location
4. Making a changeset

PR #568, opened 2023-02-13
feat: add mentionsPositions to CastAddBody

pfletcherhill:
## Motivation

We can avoid including tokens (like `%0%`) in cast text by storing the positions of mentions in a separate list.

## Change Summary

* Add `mentions_positions` list to CastAddBody schema
* Validate that positions are integers, in-range of text, and the same number as mentions

## Merge Checklist

## Additional Context

### DIFF
--- a/packages/utils/src/validations.ts
+++ b/packages/utils/src/validations.ts
@@ -280,10 +280,6 @@ export const validateCastAddBody = (body: protobufs.CastAddBody): HubResult<prot
     return err(new HubError('bad_request.validation_failure', 'mentions and mentionsPositions must match'));
   }
 
-  if (body.mentionsPositions.length > 0 && body.mentionsPositions.length !== new Set(body.mentionsPositions).size) {
-    return err(new HubError('bad_request.validation_failure', 'mentionsPositions must be unique'));
-  }
-
   for (let i = 0; i < body.mentions.length; i++) {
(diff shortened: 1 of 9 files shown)


packages/utils/src/validations.ts
```diff
@@ -268,11 +272,34 @@ export const validateCastAddBody = (body: protobufs.CastAddBody): HubResult<prot
...
-  // TODO: validate mentions are actually fids
   if (body.mentions.length > 5) {
     return err(new HubError('bad_request.validation_failure', 'mentions > 5'));
   }
 
+  if (body.mentions.length !== body.mentionsPositions.length) {
+    return err(new HubError('bad_request.validation_failure', 'mentions and mentionsPositions must match'));
+  }
+
+  if (body.mentionsPositions.length > 0 && body.mentionsPositions.length !== new Set(body.mentionsPositions).size) {
```
sds, line 283: Not sure if worthwhile, but if we guaranteed `mentionsPositions` was in sorted order, wouldn't need to validate using a `Set` (just check for any duplicate items next to each other in the array, achievable in linear time).

Also probably makes sense to have the `mentionsPositions` be sorted since as a client goes through the list of positions they will update the cast text using a string builder, and if these aren't sorted that gets a little bit trickier (can't use a simple string builder—need to recreate the string each time).

Issue #555, opened 2023-02-11
docs: add docs for all api calls in the repository 


varunsrin:
**What is the feature you would like to implement?**

Add a markdown file that contains all hubble gRPC API calls. 

**Why is this feature important?**

Make it easier for developers to get started with the Hubs. 

**Will the protocol spec need to be updated??**
No

**How should this feature be built? (optional)**
Ideally try to autogen some of the code and then document it, but write by hand if necessary.

vinliao:
Hi, I'm trying to familiarize myself with the codebase. I figured out writing this documentation is a pretty good way to do it.

Correct me if I'm wrong, but it seems like the gRPC API endpoints are the ones found in `/packages/js/src/utils.ts`?

Here's a sample of the docs:

`async getCast(fid: number, hash: Uint8Array)`
Description: Retrieves a single cast object by it the author's fid and the cast's hash.
Parameters:
- fid: the id of user in the Farcaster network
- hash: the cast hash
Returns: TODO: example response

`async getCastsByFid(fid: number)`
Description: Retrieves an array of cast objects from a specific user
Parameters:
- fid: the id of user in the Farcaster network
Returns: TODO: example response

`async getCastsByParent(parentId: protobufs.CastId)`
Description: Retrieves an array of cast objects that are children of a specified parent cast.
Parameters:
- parentId: The ID of the parent cast. `CastId: { fid: number, hash: Uint8Array }`
Returns: TODO: example response

Still figuring out how to run Hubble locally so I can actually see the what the return values of these functions are. It seems beginner-friendly enough that I can help contribute to this Issue.

varunsrin:
So there are two level of documentation we will need to write: 

1. docs for farcaster/protobufs, which document the raw .proto files and rpc methods (I've started this [here](https://github.com/farcasterxyz/hubble/pull/562))
2. docs for farcaster/js which documents the typescript endpoints (currently not being worked on) 


If you wanted to pick up #2, that would be much appreciated! Some considerations: 

1. Can we auto-generate the docs by adding comments into the code and using some tool to produce files?
2. Make sure that (1) doesn't conflict with auto-generated ts-proto files, since changes will not be preserved in those files
3. Can we generate them as simple markdown files or something that renders in github? 

I would suggest picking a couple of complicated API methods and documenting it thoroughly, so we can discuss style, format etc. Once we have that in place we can expand to do the others.

vinliao:
[...]

PR #590, opened 2023-02-18
docs: @farcaster/js

vinliao:
## Motivation

Add docs to @farcaster/js package.

## Change Summary

Initial file and table structure for the docs. Not done yet, just wanted to solicit feedback for these changes before we continue further.

## Merge Checklist

## Additional Context

"Description" column is generated by GPT, so it may be weird at times. Just wanted to get the file and table structure right for now. Check out https://github.com/vinliao/hubble/tree/md-docs/packages/js. Please give feedback if you have any!


varunsrin:
@deodad would be worth a review from you

deodad:
@vinliao thanks for the PR!

In #555 you mention of using an automated tool to build these docs but I don't see anything here. Are these docs generated manually or automatically?

I'd like to see something like:
- annotations are made following [TSDoc](https://tsdoc.org/) directly in code files (easier to maintain + shows up in hover information)
- markdown is generated using TypeDoc or a similar tool
- package README has docs table of contents + section about how the docs are generated (see #562)

vinliao:
I actually generated a bunch of markdown files with TypeDoc, but it's not pretty: https://github.com/vinliao/fcjs-docs-md. The markdowns in the PR is mostly manual. Re TSDoc: #555 seems to only be about markdown, that's why this PR is only markdown. But do let me know if TSDoc is necessary and I'll add it.

deodad:
If we can make annotations using TSDoc and generate markdown with TypeDoc I think it's a big gain in terms of usability and maintenance costs, so I'm in favor of going with your original approach.

Taking a look at one of the TypeDoc outputted [markdown files](https://github.com/vinliao/fcjs-docs-md/blob/master/classes/Eip712Signer.md), it doesn't look bad imo, just sparse since there aren't many comments in that file. Wondering more specifically what you didn't like when you said it's not pretty? Thinking if we took the examples and descriptions you added to [Eip712Signer.md](https://github.com/farcasterxyz/hubble/blob/b933e0c59b4d39f98b65200c6db7e3b67c95278c/packages/js/docs/Eip712Signer.md), added them as [TSDoc annotations](https://tsdoc.org/) to `eip712Signer.ts`, and re-generated it'd be pretty darn nice.

vinliao:
With TypeDoc: https://github.com/vinliao/hubble/blob/md-docs/packages/js/docs/classes/js_src.Eip712Signer.md. I've configured `docs` script on the root `package.json`, so now it's the matter of adding docs to the TS files (still missing a lot!).

varunsrin:
I like the direction in the Eip712Signer file [here](https://github.com/vinliao/hubble/blob/md-docs/packages/js/docs/classes/Eip712Signer.md). I think with enough annotations the auto-generated format looks decent. 

@vinliao one suggestion is to do break this up into smaller PR's to parallelize the work. the initial PR can just set the guidelines for how to comment and generate documentation, and do that for a single file. follow on pr's can document each file, and they can be done slowly and by different people if necessary.

vinliao:
The initial structure is actually done I think, the rest is about adding TSDoc comments and running `yarn docs` to generate the docs. Is the current branch good enough to be merged? I think I can write a small contribution guide in `contributing.md` or something inside `packages/js`. Thoughts? Anything I'm missing?

varunsrin (changes requested):
Generally, looks good, I'd make a few suggestions here: 

1. Rebase on the latest and regenerate docs, since some of the methods have been recently changed 

2. Exclude files that only have boilerplate documentation like Client, modules, types and utils so that we remember to add useful documentation and regenerate later. 

3. Open up a list of files that are excluded and must be documented in the ticket, so that someone can pick up the follow ups.

CONTRIBUTING.md
```diff
@@ -105,6 +105,7 @@ All changes should have supporting documentation that makes reviewing and unders
 - Do not comment obvious changes (e.g. `starts the db` before the line `db.start()`)
 - Add a `Safety: ..` comment explaining every use of `as`.
 - Prefer active, present-tense doing form (`Gets the connection`) over other forms (`Connection is obtained`, `Get the connection`, `We get the connection`, `will get a connection`)
+- To generate docs for `packages/*`, run `yarn docs` in the root of the monorepo. To see how the TSDoc comment is formatted, see [signMessageHash() in ed25519Signer.ts](packages/utils/src/signers/ed25519Signer.ts).
```
varunsrin, line 108: this isn't strictly true, since the protobufs package has a more involved generation process that is documented in its README.md.

varunsrin (changes requested):

deodad (dismissed):
A few small changes. 

Agree w/ @varunsrin suggestion to that excluding docs from certain parts would be nice, but technically none of it is wrong so I don't think this needs to block.

One more thought: what do you think of using setting up `typedoc` in `packages/js` instead of the root level? A modular approach may make it easier to setup each package to generate its own `docs` folder (for instance packages/protobufs generates its documentation using protoc-gen-doc).

CONTRIBUTING.md
vinliao: Ah, yep, forget about the protobufs, change `packages/*` to `packages/js` and it should be good to go?

vinliao:
@deodad The TypeDoc was initially inside `packages/js`, but since it depends on utils (which means it has to go `../utils`, many things were broken. Are there downside to putting the docs script on the root compared to `packages/js`?

CONTRIBUTING.md
varunsrin: yeah

varunsrin:
> @deodad The TypeDoc was initially inside `packages/js`, but since it depends on utils (which means it has to go `../utils`, many things were broken. Are there downside to putting the docs script on the root compared to `packages/js`?

fine for this release i think

packages/js/docs/README.md
```diff
@@ -0,0 +1,282 @@
+@farcaster/js / [Exports](modules.md)
+
+# @farcaster/js
+
+A collection of Typescript classes and methods for easily creating Farcaster messages and interacting with Farcaster hubs over gRPC.
+
+TODO: explain what protobufs are, and what this package does under the hood
```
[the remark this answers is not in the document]
vinliao: Seems useful to give a one (or two) sentence explanation about how Hubble works, so JS devs can understand what they're working with. Maybe point to the protocol docs instead?

packages/js/docs/classes/Ed25519Signer.md
```diff
@@ -0,0 +1,223 @@
...
+
+A HubResult containing an Ed25519Signer instance on success, or an error message on failure.
+
+#### Overrides
+
+BaseEd25519Signer.fromPrivateKey
+
+#### Defined in
+
+[js/src/signers.ts:128](https://github.com/vinliao/hubble/blob/7629cfa/packages/js/src/signers.ts#L128)
```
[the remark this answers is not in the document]
vinliao: Made the change. Is there a way to also disable "inherited from" text with another TypeDoc option? (Not critical, but I think it's quite verbose).

image

vinliao:
Should there be docs TODOs on utils (https://github.com/vinliao/hubble/blob/md-docs/packages/js/docs/modules/utils.md)? All seems to be internal functions that JS devs don't have to touch.

deodad:
> Should there be docs TODOs on utils (https://github.com/vinliao/hubble/blob/md-docs/packages/js/docs/modules/utils.md)? All seems to be internal functions that JS devs don't have to touch.

I think it's fine to leave these out for now

packages/js/docs/README.md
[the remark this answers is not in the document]
deodad: That seems resonable!

packages/js/docs/README.md
```diff
@@ -0,0 +1,282 @@
+@farcaster/js / [Exports](modules.md)
```
deodad, line 1: I see that this same file exists at `packages/js/docs/README.md` and `packages/js/README.md`, is this intentional (e.g. file copied over when typedoc is run) or should we pick one?

vinliao: `js/docs/README.md` is generated from `/js/README.md`. It's kind of redundant, but not sure how to turn that off. `js/README.md` is what users will see, and the links inside it will point to the client docs, signer docs, etc.

varunsrin (approved):

PR #591, opened 2023-02-19
fix: Allow empty cast text

sds:
## Motivation

Since a cast could consist of just a mention, a valid cast can have no actual text (i.e. is the empty string), only a mention.

## Change Summary

Don't fail validation if an empty text string is provided for the cast.

## Merge Checklist


pfletcherhill (approved):
lgtm, thanks for making the change

PR #596, opened 2023-02-20
fix: Allow duplicate `mentionsPositions`

sds:
## Motivation

It is possible for a cast to have multiple mentions with the same position. Consider the following cast:

    @a@b@c

In this case, we are mentioning the users `a`, `b`, and `c`, without separating their names with any other text. The resulting `mentionsPositions` will be `[0, 0, 0]`, since there is no change in position.

## Change Summary

Relax our validations to allow this case.

## Merge Checklist

### DIFF
--- a/.changeset/slimy-jobs-do.md
+++ b/.changeset/slimy-jobs-do.md
@@ -1,0 +1,5 @@
+---
+'@farcaster/utils': patch
+---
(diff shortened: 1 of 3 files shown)


pfletcherhill (approved):
good catch

PR #607, opened 2023-02-24
feat: add ability to customize log level from the cli

jcliff:
An environment variable is utilized for simplicity

Also, clarify that --version is available from the cli in the help menu

## Motivation

This allows users to configure the minimum log level they want outputted.

## Change Summary

Add FARCASTER_LOG_LEVEL to the environment variables utilized.

## Merge Checklist

### DIFF
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -8,7 +8,7 @@ import { ResultAsync } from 'neverthrow';
 import { dirname, resolve } from 'path';
 import { exit } from 'process';
 import { APP_VERSION, Hub, HubOptions } from '~/hubble';
-import { logger } from '~/utils/logger';
+import { logger, loggerLevels } from '~/utils/logger';
 import { addressInfoFromParts, ipMultiAddrStrFromAddressInfo, parseAddress } from '~/utils/p2p';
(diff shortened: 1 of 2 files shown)


apps/hubble/src/cli.ts
```diff
@@ -26,7 +26,8 @@ app.name('hub').description('Farcaster Hub').version(APP_VERSION);
 
 app
   .command('start')
-  .description('Start a Hub')
+  .description("Start a Hub\n")
+  .description('Set minimum log level with FARCASTER_LOG_LEVEL from ' + loggerLevels )
```
varunsrin, line 30: is there a reason you prefer the env variable over the cli flag?

varunsrin (changes requested):

apps/hubble/src/cli.ts
varunsrin: environment variables are a little more of a pain to configure in cloud environments, and so its a nicer devex to do it all with cli or a config file (ideally both) 

i can take a look and see if there's a way to wrangle it into a cli

varunsrin:
thanks for rebasing! we're in a code freeze now due to the imminent release of hubs, will start merging after the release

Issue #356, opened 2023-01-02
feat: pagination for list queries and gRPC


pfletcherhill:
**What is the feature you would like to implement?**
* Pagination for all list queries (i.e. `getCastsByFid`, `getAmpsByUser`, `getFids`)
* Update gRPC requests to include pagination params

varunsrin:
@deodad do you have cycles to pick this up this week? @pfletcherhill will write up more details on this

deodad:
yes I can pick this up

deodad:
@pfletcherhill I have some spare cycles, any further details?

PR #610, opened 2023-02-24
feat: pagination for list queries and gRPC

deodad:
## Motivation

See #356 

## Change Summary

* Support pagination on list methods and let user specify `pageSize` (number of records to return, up to 10k) and `pageToken` (token representing the last key they saw)
* Responses include `nextPageToken`
* Returns messages in chronological order within the context of an fid

## Merge Checklist

## Additional Context


packages/protobufs/src/schemas/rpc.proto
```diff
@@ -57,12 +57,18 @@ message SyncIds {
     repeated bytes sync_ids = 1;
 }
 
+message ListRequest {
```
varunsrin, line 60: is there a reason we call this ListRequest and not FidsRequest in line with our other request names?

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -57,12 +57,18 @@ message SyncIds {
...
+  int32 page_size = 2;
```
varunsrin, line 61: why do we start at property 2 and not 1?

packages/protobufs/src/generated/rpc.ts
```diff
@@ -130,12 +130,18 @@ export interface SyncIds {
   syncIds: Uint8Array[];
 }
 
+export interface ListRequest {
```
varunsrin, line 133: same q about why not FidsRequest

deodad:
Notes from first look:
- extend GRPC service requests as outlined by https://cloud.google.com/apis/design/design_patterns#list_pagination
  - use base64url encoded strings `page_token` / `next_page_token` on request / response protobufs
- use the prefix of the last element from page n - 1 to identify page n
- how to handle `getAll*` endpoints (i.e. [getAllCastsMessagesByFid](https://github.com/farcasterxyz/hubble/blob/main/apps/hubble/src/storage/engine/index.ts#L202-L216)) that perform concatenation on two engine result sets?

apps/hubble/src/storage/engine/index.ts
```diff
@@ -406,8 +406,14 @@ class Engine {
     return ResultAsync.fromPromise(this._signerStore.getIdRegistryEvent(fid), (e) => e as HubError);
   }
 
-  async getFids(): HubAsyncResult<number[]> {
-    return ResultAsync.fromPromise(this._signerStore.getFids(), (e) => e as HubError);
+  async getFids(
```
varunsrin, line 409: i'd avoid exposing the terminology of pages into the engine and rocksdb, since that's an api level concern. 

would recommend names like `startPrefix`, `nextPrefix`  and `returnSize` or something along those lines

packages/protobufs/src/generated/rpc.ts
```diff
@@ -130,12 +130,18 @@ export interface SyncIds {
...
+  pageSize: number;
```
varunsrin, line 134: should we be opinionated on the max page size allowed? i assume we want some limit here, even if very large

apps/hubble/src/storage/stores/signerStore.ts
```diff
@@ -155,13 +155,30 @@ class SignerStore {
...
+      keyAsBuffer: true,
+      values: false,
+      limit: pageSize,
+    })) {
       fids.push(Number((key as Buffer).readUint32BE(1)));
     }
-    return fids;
+
+    const buf = Buffer.alloc(4);
+    buf.writeUInt32BE(fids[fids.length - 1] ?? 0, 0);
```
varunsrin, line 176: a comment explaining this logic would be helpful for future readers

apps/hubble/src/storage/db/rocksdb.ts
```diff
@@ -167,6 +167,31 @@ class RocksDB {
...
+   * order, so this iterator will continue serving keys in order until it receives one that has a lexicographic order
+   * greater than the prefix.
+   */
+  iteratorByPrefixPaged(
+    prefix: Buffer,
+    pagePrefix?: Buffer,
+    options: AbstractRocksDB.IteratorOptions = {}
+  ): AbstractRocksDB.Iterator {
+    if (pagePrefix) {
+      options.gt = pagePrefix;
```
varunsrin, line 182: very minor nit, but should nextPagePrefix be the first item of the next page instead of the last item of the current page? 

semantically, its what i would expect when reading the code, and it also allows us to do a simpler gte here for both cases

apps/hubble/src/rpc/server/test/signerService.test.ts
```diff
@@ -120,14 +124,36 @@ describe('getIdRegistryEvent', () => {
...
+  test('returns pageSize results', async () => {
+    await engine.mergeIdRegistryEvent(custodyEvent);
+    await engine.mergeIdRegistryEvent(custodyEvent2);
+    // const result = await client.getFids();
+    // expect(result._unsafeUnwrap()).toEqual([custodyEvent.fid()]);
+    const result = await client.getFids(protobufs.ListRequest.create({ pageSize: 1 }));
     expect(result._unsafeUnwrap().fids).toEqual([custodyEvent.fid]);
   });
 
+  test('returns results after pageToken', async () => {
```
varunsrin, line 143: a few more test cases: 

1. what happens when pageSize > available items or invalid? 
2. what happens when pageToken is invalid ?

apps/hubble/src/storage/db/rocksdb.ts
deodad: yes, it also gives the benefit that the presence or absence of `nextPrefix` hints to whether there are more results or not

apps/hubble/src/storage/stores/signerStore.ts
deodad: found `makeIdRegistryEventPrimaryKey`

packages/protobufs/src/generated/rpc.ts
deodad: good idea. how about 10k? and should a default value of 0 in pageSize indicate the limit?

deodad:
Feedback is addressed and I added `getAllSignerMessagesByFid` to get down a pattern for getAll*

packages/protobufs/src/generated/rpc.ts
varunsrin: im less familiar with gRPC, so im not sure what an appropriate size here is. would depending on a few things: 

1. what is the largest possible size of 10k messages? 
2. what does gRPC recommend for size of messages? is there an upper limit? what does google usually do? 
3. are there any perf reasons on our side that would make us want to keep this smaller?  @adityapk00 might have thoughts about reading from the engine/rocksdb

adityapk00: Casts are the largest message type, with an average size of ~400bytes encoded. However, with mentions and stuff, a message could be as large as 1kb. 

We should generally have this be equal to how much we want to load in memory. With 10k limit, and a 1kb max size, we're at 10MB response size, which seems Ok. Anything larger and we're going to have grpc bottlenecks.

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -48,14 +48,23 @@ message SyncIds {
 
 message FidRequest {
   uint64 fid = 1;
+  int32 page_size = 2;
```
adityapk00, line 51: Instead of adding page_size and page_token here, we should have the RPC methods accept a second argument with a new protobuf `PagedRequest` with these 2 vars.

apps/hubble/src/rpc/server/index.ts
```diff
@@ -470,10 +474,27 @@ export default class Server {
         );
       },
       getFids: async (call, callback) => {
-        const result = await this.engine?.getFids();
+        const request = call.request;
+
+        if (request.pageSize < 0 || request.pageSize > 10_000) {
```
adityapk00, line 479: pagesize should be uint32, so it can never be negative

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -48,14 +48,23 @@ message SyncIds {
...
+  string page_token = 3;
+}
+
+message FidsRequest {
+  int32 page_size = 1;
```
adityapk00, line 56: This should be uint. A negative page_size is not allowed.

packages/protobufs/src/schemas/rpc.proto
```diff
@@ -48,14 +48,23 @@ message SyncIds {
 
 message FidRequest {
   uint64 fid = 1;
+  int32 page_size = 2;
+  string page_token = 3;
```
adityapk00, line 52: this should be bytes, right? The page_token and next_page_token are indexes into the rocks db keys, so we should make them bytes.

deodad: Yes, ultimately they're used as indexes into rocks db keys.

The idea, taken from [here](https://cloud.google.com/apis/design/design_patterns#list_pagination), is to make them url-safe base64 encoded strings so they can "evolve without compatibility issues". Not sure how that applies in our situation but having them be in a standard format and as url-safe string didn't seem like a bad idea so they could be included in urls.

I'm fine w/ bytes though.

adityapk00: Yes, I'd prefer to use `bytes`. The general philosophy here is that the protobufs use the "closest to metal" format for the fields, and if needed, the higher layers can encode/decode into strings/base64 as needed.

deodad: @adityapk00 makes sense, in that spirit do you prefer using `limit`, `startPrefix`, and `nextPrefix` in the protobufs instead of `pageSize`, `pageToken`, `nextPageToken`?

also wondering your thoughts on a dedicated protobuf type for holding the pagination properties vs adding them to directly to the request types?

adityapk00: I'd prefer to keep the existing terms, because if we change it to `{start|next}Prefix`, we're inviting the caller to interpret these as prefixes, and we don't want them to do that. We might change the interpretation of these in the future (If we change the rocksDB keys, for example in a future migration). Using `token` is preferable. 

As for the dedicated protobuf, I'll leave that to you - Whatever you think makes it easier. I don't have a preference one way or another.

PR #616, opened 2023-02-26
chore: Remove unspecified/unrecognized enum values from generated types

sds:
# WIP. Looking for feedback on whether we want to continue pursuing

## Motivation

Currently, our various enums are created to look like the following:

    enum MyEnum {
      MY_ENUM_NONE = 0,
      ...
      UNRECOGNIZED = -1,
    }

The `*_NONE` exists because proto3 has a hard requirement on the first value being 0.

The `UNRECOGNIZED` value can be disabled with a flag.

We want to remove these values because it makes it easier for us to perform type narrowing in TypeScript, which makes working with the library more pleasant/convenient since TypeScript's typechecking can do more heavy lifting.

## Change Summary

We disable `UNRECOGNIZED` since it doesn't serve much practical purpose for our uses, and we export a number of types from `@farcaster/js` that explicitly excludes the `*_NONE` value from the various enums we reference.

## Merge Checklist

### DIFF
--- a/.changeset/orange-falcons-reply.md
+++ b/.changeset/orange-falcons-reply.md
@@ -1,0 +1,6 @@
+---
+'@farcaster/protobufs': minor
+'@farcaster/js': patch
+---
+
+Remove unspecified/unrecognized enum values from generated types
--- a/packages/js/src/types.ts
+++ b/packages/js/src/types.ts
@@ -1,32 +1,30 @@
 import * as protobufs from '@farcaster/protobufs';
 
-export {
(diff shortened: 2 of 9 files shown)


varunsrin:
will defer to @pfletcherhill on making this decision

sds:
Decision was to remove `UNRECOGNIZED`, but we'll need to leave in `NONE` for now since it's a little more nuanced to remove due to how the protobuf TS code is auto-generated. We can work around it for now.

PR #623, opened 2023-03-01
feat: add ability to change contract configuration via cli

### LINKED ISSUE #622: feat: add ability to change contract configuration via cli
Ability to override the default contract addresses (and corresponding syncing configurations) when starting a hub.

Exposing this will allow developers to create fully isolated testing environments if they wish to run a hub against a local/private Ethereum node.

**Will the protocol spec need to be updated?**
No.

**How should this feature be built?**
Expose all of `GoerliEthConstants` as CLI options.
https://github.com/farcasterxyz/hubble/blob/a237a6554a12d8822e74271d943169a355efdb8f/apps/hubble/src/eth/ethEventsProvider.ts#L14-L19

**Additional context**
Features originally proposed in https://github.com/farcasterxyz/hubble/issues/585

payton:
## Motivation

Exposing this will allow developers to create fully isolated testing environments if they wish to run a hub against a local/private Ethereum node (resolves #622). 

## Change Summary

Expose all of `GoerliEthConstants` as CLI options.
https://github.com/farcasterxyz/hubble/blob/a237a6554a12d8822e74271d943169a355efdb8f/apps/hubble/src/eth/ethEventsProvider.ts#L14-L19

## Merge Checklist

- [x] (N/A) Changes to the protocol specification have been merged

## Additional Context

In the future, I think it will be worth adding an abi override option for testing changes to `farcasterxyz/contracts`. This feels less important in the current state as I imagine the vast majority of cases using this feature will be to reproduce production.

### DIFF
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -29,7 +29,13 @@ app
   .description('Start a Hub')
   .option('-e, --eth-rpc-url <url>', 'RPC URL of a Goerli Ethereum Node')
   .option('-c, --config <filepath>', 'Path to a config file with options', DEFAULT_CONFIG_FILE)
-  .option('-f, --fir-address <address>', 'The address of the FIR contract')
+  .option('--fir-address <address>', 'The address of the Farcaster ID Registry contract')
+  .option('--fnr-address <address>', 'The address of the Farcaster Name Registry contract')
+  .option('--first-block <number>', 'The block number to begin syncing events from Farcaster contracts')
+  .option(
+    '--chunk-size <number>',
+    'The number of blocks to batch when syncing historical events from Farcaster contracts'
+  )
   .option('-b, --bootstrap <peer-multiaddrs...>', 'A list of peer multiaddrs to bootstrap libp2p')
   .option('-a, --allowed-peers <peerIds...>', 'An allow-list of peer ids permitted to connect to the hub')
   .option('--ip <ip-address>', 'The IP address libp2p should listen on. (default: "127.0.0.1")')
@@ -110,7 +116,10 @@ app
       gossipPort: hubAddressInfo.value.port,
       network: cliOptions.network ?? hubConfig.network,
       ethRpcUrl: cliOptions.ethRpcUrl ?? hubConfig.ethRpcUrl,
-      IdRegistryAddress: cliOptions.firAddress ?? hubConfig.firAddress,
+      idRegistryAddress: cliOptions.firAddress ?? hubConfig.firAddress,
+      nameRegistryAddress: cliOptions.fnrAddress ?? hubConfig.fnrAddress,
+      firstBlock: cliOptions.firstBlock ?? hubConfig.firstBlock,
+      chunkSize: cliOptions.chunkSize ?? hubConfig.chunkSize,
       bootstrapAddrs,
       allowedPeers: cliOptions.allowedPeers ?? hubConfig.allowedPeers,
       rpcPort: cliOptions.rpcPort ?? hubConfig.rpcPort,
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -20,7 +20,7 @@ const PEER_ID_FILENAME = 'id.protobuf';
 const DEFAULT_PEER_ID_DIR = './.hub';
 const DEFAULT_PEER_ID_FILENAME = `default_${PEER_ID_FILENAME}`;
 const DEFAULT_PEER_ID_LOCATION = `${DEFAULT_PEER_ID_DIR}/${DEFAULT_PEER_ID_FILENAME}`;
-const DEFAULT_CHUNK_SIZE = '10000';
+const DEFAULT_CHUNK_SIZE = 10000;
 
 const app = new Command();
 app.name('hub').description('Farcaster Hub').version(APP_VERSION);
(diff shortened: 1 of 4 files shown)


apps/hubble/src/cli.ts
```diff
@@ -29,7 +29,13 @@ app
   .description('Start a Hub')
   .option('-e, --eth-rpc-url <url>', 'RPC URL of a Goerli Ethereum Node')
   .option('-c, --config <filepath>', 'Path to a config file with options', DEFAULT_CONFIG_FILE)
-  .option('-f, --fir-address <address>', 'The address of the FIR contract')
+  .option('--fir-address <address>', 'The address of the Farcaster ID Registry contract')
+  .option('--fnr-address <address>', 'The address of the Farcaster Name Registry contract')
+  .option('--first-block <number>', 'The block number to begin syncing events from Farcaster contracts')
+  .option(
```
adityapk00, line 35: Can you add a default for this? Maybe 10k blocks

apps/hubble/src/cli.ts
```diff
@@ -110,7 +116,10 @@ app
       gossipPort: hubAddressInfo.value.port,
       network: cliOptions.network ?? hubConfig.network,
       ethRpcUrl: cliOptions.ethRpcUrl ?? hubConfig.ethRpcUrl,
-      IdRegistryAddress: cliOptions.firAddress ?? hubConfig.firAddress,
+      idRegistryAddress: cliOptions.firAddress ?? hubConfig.firAddress,
+      nameRegistryAddress: cliOptions.fnrAddress ?? hubConfig.fnrAddress,
+      firstBlock: cliOptions.firstBlock ?? hubConfig.firstBlock,
+      chunkSize: cliOptions.chunkSize ?? hubConfig.chunkSize,
```
adityapk00, line 122: cliOptions.chunkSize ?? hubConfig.chunkSize ?? DEFAULT_CHUNK_SIZE

varunsrin:
@payton you'll need to sign your commits for it to be accepted: https://github.com/farcasterxyz/hubble/blob/main/CONTRIBUTING.md#22-signing-commits

adityapk00 (approved):
Please add a default, but approved

adityapk00:
You also need to sign your commits, you won't be able to merge without it. Please see `Contributing.md`

apps/hubble/src/cli.ts
payton: @adityapk00 is it preferred to pull this out of the `GoerliEthConstants` and into a generic constant?

Default chunk size should be relative to the given network size, no?

payton:
@varunsrin @adityapk00 apologies. I was following `Contributing.md`, but I goofed. 

- [x] Signed last commit.

apps/hubble/src/cli.ts
payton: (also, see my above message about defaults in the hub constructor)

adityapk00: It should be network specific, so fine to leave it in the GoerliEthConstants. I meant that if none is specified, we should default to 10k.

The reason we have chunks is actually so that the calls to Alchemy/Infura don't time out. However, if a developer is running a local net or a fork of the mainnet locally, we don't want them to additionally think about the chunk size. So I would:

1. define a new constant "DEFAULT_CHUNK_SIZE" which is 10k
2. If the chunk size is specified neither in the CLI options nor in the config, just use the DEFAULT_CHUNK_SIZE.

So, if you replace line 122 with `cliOptions.chunkSize ?? hubConfig.chunkSize ?? DEFAULT_CHUNK_SIZE` you'll get this. Also maybe add it to the help string for the option

apps/hubble/src/cli.ts
adityapk00: Yes, because I'm expecting that we'll refactor this further in the future for local devnets and testnets.

apps/hubble/src/cli.ts
payton: Thanks for the explanation. Understood!

payton:
@adityapk00 all set.

apps/hubble/src/cli.ts
```diff
@@ -29,7 +30,14 @@ app
...
   .option('-e, --eth-rpc-url <url>', 'RPC URL of a Goerli Ethereum Node')
   .option('-c, --config <filepath>', 'Path to a config file with options', DEFAULT_CONFIG_FILE)
-  .option('-f, --fir-address <address>', 'The address of the FIR contract')
+  .option('--fir-address <address>', 'The address of the Farcaster ID Registry contract')
+  .option('--fnr-address <address>', 'The address of the Farcaster Name Registry contract')
+  .option('--first-block <number>', 'The block number to begin syncing events from Farcaster contracts')
+  .option(
+    '--chunk-size <number>',
+    'The number of blocks to batch when syncing historical events from Farcaster contracts. (default: 10000)',
+    DEFAULT_CHUNK_SIZE
```
adityapk00, line 39: Sorry to be so pedantic, but can you remove line 39. If you specify the DEFAULT_CHUNK_SIZE here, then cliOptions.chunkSize will always be set, and it will override anything that the user sets in the config file. We don't want that. 

The order we want is:
explicitly specified on CLI -> specified in config file -> default. 

If you remove it from here (but keep the text), I think we'll be set.

payton: Ah, no problem. My mistake :sweat_smile: .

Removed the option default and changed `DEFAULT_CHUNK_SIZE` to a number from a string.

adityapk00:
Excellent! Please "Squash and merge" it!

If the button doesn't show, let me know, and I can do it on your behalf.

payton:
@adityapk00 no write access, so I can't squash and merge!

varunsrin:
i got you

PR #633, opened 2023-03-10
fix: validate length limits in terms of utf8 bytes and allow up to 10 mentions

sds:
## Motivation

We want to know exactly how many bytes we're storing.

Furthermore, we want to guarantee that regardless of future iterations to the Unicode specification that we have an unambiguous way to take text and mention positions and reconstruct cast text from that by inserting dynamic content at those positions (e.g. username, which can change).

## Change Summary

Check using utf8 byte length instead of character length.

While here, based on a team discussion and existing cast data we also increased the maximum number of mentions from 5 to 10.

## Merge Checklist

### DIFF
--- a/packages/utils/src/validations.test.ts
+++ b/packages/utils/src/validations.test.ts
@@ -218,18 +218,16 @@ describe('validateCastAddBody', () => {
       hubErrorMessage = 'text is missing';
     });
 
-    test('when text is longer than 320 characters', () => {
+    test('when text is longer than 320 ASCII characters', () => {
       body = Factories.CastAddBody.build({ text: faker.random.alphaNumeric(321) });
-      hubErrorMessage = 'text > 320 chars';
+      hubErrorMessage = 'text > 320 bytes';
     });
 
     test('when text is longer than 320 bytes', () => {
-      let text = '';
-      for (let i = 0; i < 320; i++) {
-        text = text + '🤓';
-      }
+      const text = faker.random.alphaNumeric(318) + '🤓';
+      expect(text.length).toEqual(320);
       body = Factories.CastAddBody.build({ text });
-      hubErrorMessage = 'text > 320 chars';
+      hubErrorMessage = 'text > 320 bytes';
     });
 
     test('with more than 2 embeds', () => {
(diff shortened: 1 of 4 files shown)


packages/utils/src/validations.ts
```diff
@@ -388,32 +398,47 @@ export const validateUserDataType = (type: number): HubResult<protobufs.UserData
...
-        return err(validatedFname.error);
+
+  const textUtf8BytesResult = utf8StringToBytes(value);
+  if (textUtf8BytesResult.isErr()) {
+    return err(new HubError('bad_request.invalid_param', 'value cannot be encoded as utf8'));
+  }
+
+  const valueBytes = textUtf8BytesResult.value;
+
+  switch (type) {
```
sds, line 409: Switching to a case statement for performance, especially as the number of data types grows.
https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html#if-else_versus_switch

packages/utils/src/validations.test.ts
```diff
@@ -282,6 +280,24 @@ describe('validateCastAddBody', () => {
       hubErrorMessage = 'mentionsPositions must be a position in text';
     });
 
+    test('with mentionsPositions within byte length of text', () => {
```
pfletcherhill, line 283: I think this should fail and mentions positions should be based on chars

pfletcherhill (changes requested):
Lgtm though I suggest we limit length by bytes and use character positions for mentions

packages/utils/src/validations.ts
```diff
@@ -277,7 +283,7 @@ export const validateCastAddBody = (body: protobufs.CastAddBody): HubResult<prot
     if (typeof position !== 'number' || !Number.isInteger(position)) {
       return err(new HubError('bad_request.validation_failure', 'mentionsPositions must be integers'));
     }
-    if (position < 0 || position > text.length) {
+    if (position < 0 || position > textBytes.length) {
```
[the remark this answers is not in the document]
sds: Yeah, I wish it were that simple. Counter-example:

```ts
[..."👉🏿"].length // 2 because this is technically two Unicode codepoints, but a single character
```

Trying to see if there's an off-the-shelf solution available.

PR #634, opened 2023-03-10
fix: getCastsByFid should return items in reverse chronological order

bstchow:
## Motivation
Fixes https://github.com/farcasterxyz/hubble/issues/611

## Change Summary
Modifies CastStore#getCastAddsByFid to fetch all messages for a given FID rather than only CastAdds from the DB.
Fetching all messages utilizes thash so casts are retrieved in chronological order.

## Merge Checklist

- [x] Changes to the protocol specification have been merged

## Additional Context
Not sure that I fully understood approach described in https://github.com/farcasterxyz/hubble/issues/611
If I understand correctly, it seems like this is a bandaid, and long term it might make sense to have a parallel index RocksDB table stored with thash keys?

### DIFF
--- a/apps/hubble/src/storage/stores/castStore.ts
+++ b/apps/hubble/src/storage/stores/castStore.ts
@@ -4,7 +4,6 @@ import AsyncLock from 'async-lock';
 import { err, ok, ResultAsync } from 'neverthrow';
 import {
   deleteMessageTransaction,
-  getAllMessagesByFid,
   getAllMessagesBySigner,
   getManyMessages,
   getManyMessagesByFid,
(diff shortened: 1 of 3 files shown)


apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -143,12 +144,9 @@ class CastStore {
 
   /** Gets all CastAdd messages for an fid */
   async getCastAddsByFid(fid: number): Promise<protobufs.CastAddMessage[]> {
-    const castAddsPrefix = makeCastAddsKey(fid);
-    const messageKeys: Buffer[] = [];
-    for await (const [, value] of this._db.iteratorByPrefix(castAddsPrefix, { keys: false, valueAsBuffer: true })) {
-      messageKeys.push(value);
-    }
-    return getManyMessagesByFid(this._db, fid, UserPostfix.CastMessage, messageKeys);
+    return (await getAllMessagesByFid(this._db, fid))
```
pfletcherhill, line 147: This should still use a rocksdb iterator, but instead of using the cast adds prefix, it should use the cast messages prefix (i.e. `Buffer.concat([makeUserKey(fid), Buffer.from([UserPostfix.CastMessage])])`. Also, I think it's okay to return chronological order rather than reversed chronological.

pfletcherhill (changes requested):

apps/hubble/src/storage/stores/castStore.ts
bstchow: Gotcha - that makes sense. Let me know if the latest changes have addressed this.

apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -143,11 +154,20 @@ class CastStore {
...
-      messageKeys.push(value);
+    const castMessagePrefix = makeCastMessageKey(fid);
+
+    const messageKeys: Uint8Array[] = [];
+    for await (const [key, value] of this._db.iteratorByPrefix(castMessagePrefix, {
+      keyAsBuffer: true,
+      valueAsBuffer: true,
+    })) {
+      const tsHash = Uint8Array.from(key).subarray(castMessagePrefix.length);
+      const msg = protobufs.Message.decode(value);
```
bstchow, line 165: Seemed like the most declarative way to get the MessageType, also not sure what offset I should expect type to be at

pfletcherhill:
Closing this PR since we've integrated the commits into a separate PR that involves ordering and pagination: https://github.com/farcasterxyz/hubble/pull/610/

PR #671, opened 2023-03-18
feat: Do validations in webworkers

adityapk00:
## Motivation

The largest CPU bottleneck right now is validation of messages (Specifically, checking the signatures of messages). If we offload the validation to another thread, we can improve the ingestion speed of the hubs by ~70%

## Change Summary

- Compile target for validation webworker
- Fallback to main thread if we can't create a webworker
- Do validations in the webworker thread

## Merge Checklist

## Additional Context
This PR is for review only right now. Needs more testing before we can merge


apps/hubble/package.json
```diff
@@ -16,12 +16,13 @@
...
     "bench": "NODE_OPTIONS='--max-old-space-size=8192' tsx src/test/bench",
-    "build": "yarn clean && tsc --project ./tsconfig.json",
+    "webworker": "tsc --project ./tsconfig.ww.json",
+    "build": "yarn clean && yarn webworker && tsc --project ./tsconfig.json",
     "clean": "rimraf ./build",
     "dev": "yarn start | yarn pino-pretty",
     "lint": "eslint  src/ --color --ext .ts",
     "lint:fix": "yarn run lint -- --fix",
-    "start": "tsx src/cli.ts start",
+    "start": "yarn webworker && tsx src/cli.ts start",
```
pfletcherhill, line 25: What's the reason for building the webworker on both start and build?

apps/hubble/tsconfig.ww.json
```diff
@@ -0,0 +1,18 @@
+{
```
pfletcherhill, line 1: There's not a way of building both this and the rest of the package at the same time with a single tsconfig?

apps/hubble/src/storage/engine/index.ts
```diff
@@ -41,6 +47,32 @@ class Engine {
...
+              const message = protobufs.Message.decode(messageBytes);
+              resolve(ok(message));
+            } else {
+              resolve(err(new HubError(errCode, errMessage)));
+            }
+          }
+        });
+      }
+    } catch (e) {
+      logger.warn({ workerPath, e }, 'failed to create validation worker, falling back to main thread');
```
pfletcherhill, line 74: Are there particular situations where you expect this to happen? Or would this be a very unexpected error?

apps/hubble/package.json
```diff
@@ -16,12 +16,13 @@
   "type": "module",
   "scripts": {
     "bench": "NODE_OPTIONS='--max-old-space-size=8192' tsx src/test/bench",
-    "build": "yarn clean && tsc --project ./tsconfig.json",
+    "webworker": "tsc --project ./tsconfig.ww.json",
```
pfletcherhill, line 19: Nit: maybe rename to `build-webworker` to make it clear it's building the code not running it?

pfletcherhill (commented):
Overall this seems cool! Just a few comments about building.

apps/hubble/package.json
adityapk00: Removed this entirely to have it be built just as a part of the normal build

apps/hubble/package.json
adityapk00: removed

apps/hubble/src/storage/engine/index.ts
adityapk00: This would be very unexpected. I'm logging this just in case something strange is happening either inside docker or fargate, but this should happen.

apps/hubble/tsconfig.ww.json
adityapk00: Turns out it does build in the same tsconfig, so removed.

PR #672, opened 2023-03-18
chore: separate build step into typecheck and transpile

### LINKED ISSUE #346: chore: separate yarn build into a typecheck and transpile step

vinliao:
## Motivation

Fixes #346.

## Change Summary

Changes:
1. Build uses `swc`
2. Typecheck command: `tsc --noEmit`
3. `turbo build` depends on `turbo typecheck`

Some screenshots:

Screenshot from 2023-03-18 13-36-29
Screenshot from 2023-03-18 13-36-38
Screenshot from 2023-03-18 13-38-28

## Merge Checklist

## Additional Context

Read somewhere on docs Hubble doesn't use Turborepo caching because it's "not worth the risk" or something. (Forgot where it's located.) Both `typecheck` and `build` are cached in this PR. Is this an unwanted behavior?

### DIFF
--- a/apps/hubble/package.json
+++ b/apps/hubble/package.json
@@ -16,7 +16,8 @@
   "type": "module",
   "scripts": {
     "bench": "NODE_OPTIONS='--max-old-space-size=8192' tsx src/test/bench",
-    "build": "yarn clean && tsc --project ./tsconfig.json",
+    "build": "yarn clean && swc src -d dist",
+    "typecheck": "tsc --noEmit",
     "clean": "rimraf ./build",
     "dev": "yarn start | yarn pino-pretty",
     "lint": "eslint  src/ --color --ext .ts",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "test": "./node_modules/.bin/turbo run test",
     "test:ci": "./node_modules/.bin/turbo run test:ci -- --passWithNoTests",
     "lint": "./node_modules/.bin/turbo run lint --parallel",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "typecheck": "./node_modules/.bin/turbo run typecheck"
   },
   "engines": {
     "npm": ">=8.0.0",
--- a/packages/hub-nodejs/package.json
+++ b/packages/hub-nodejs/package.json
@@ -23,12 +23,13 @@
     "neverthrow": "^6.0.0"
   },
   "scripts": {
-    "build": "tsup src/index.ts --format cjs,esm --dts --clean",
+    "build": "swc src -d dist",
     "clean": "rimraf ./dist",
     "lint": "eslint  src/ --color --ext .ts",
     "test": "NODE_OPTIONS=--experimental-vm-modules jest",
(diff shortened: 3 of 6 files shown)


vinliao:
@varunsrin thoughts on this PR? I know @deodad has one at https://github.com/farcasterxyz/hubble/pull/411, but it seemed outdated and stale, so I made another one with less diff. Curious what you both think.

deodad:
Thanks for picking this back up @vinliao 

A couple things need to happen during the build that aren't happening here:
- output a `.d.ts` declaration files so the types are included in the npm package
- output a CSJ entrypoint `index.js` and an ESM entrypoint `index.mjs` (see [package.json](https://github.com/farcasterxyz/hubble/blob/main/packages/utils/package.json#L4-L13))

Taking a step back and looking at #346 I don't think it's possible as specified since:
- types need to be included in the build
- to type check one package requires having the latest types of any dependencies, and since currently these are looked up in `dist`, type checking one requires building its dependencies first

My approach in #411 uses [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) to achieve:
- type checking of any package without needing to build its dependencies since project references make it so type checking is done on the source code of referenced dependencies rather than their `dist` code
- `yarn build` and `yarn typecheck` are fast since with incremental building the TS compiler understands the dependency tree and maintains a cache

deodad:
When `tsc --noEmit` is run, the TS compiler will look at the local `dist` folders of the other hubble dependencies for their types and if those don't exist or are not up to date it won't compile as expected. As an example, if you run `yarn clean` in the root which will clear all `dist` folders and then run `yarn tsc --noEmit` from `apps/hubble` and it won't compile. So developing across the repo requires keeping the builds of all repo-located dependencies up to date.

#411 never got reviewed and got stale, so it's possible there may still be dragons (mostly around validating that the CSJ and ESM builds both work the same, I can describe how to do that if needed) but I'm fairly confident it's a solid approach. If you wanted you could revive and validate it that'd be awesome, but yes there will be a code freeze so no rush.

deodad:
@vinliao can you run `yarn typecheck --force`? I don't think it's actually running since Turbo doesn't see any changes (see `cache hit, replaying output`). Another way to check is to actually cd into any of the leaf packages and run `yarn typecheck`

vinliao:
Heh, the first time it ran it was OK, (hence it was cached by Turbo). Now it outputs things like `src/index.ts(1,15): error TS2307: Cannot find module '@farcaster/protobufs' or its corresponding type declarations.` Still got some stuff to fix, thanks for pointing it out :)

sds:
Are you still working on this, @vinliao? No worries if not, just wanted to check before closing. Thanks!

vinliao:
Currently not, please close it :)

PR #675, opened 2023-03-18
feat: refactor events to create and broadcast in order

pfletcherhill:
## Motivation

Right now event creation and broadcasting happen separately, so an event can reserve a certain event ID but by the time that event is broadcast another event with a later ID has already been sent. Transactions with many operations (like pruning or revoking signers) could demonstrate this behavior, and we want to guarantee that events are broadcast in order (by ID).

## Change Summary

* Refactor event handler to only assign IDs to events when they're committed and broadcast them in the same method
* Lock the event handler when a transaction is being processed

## Merge Checklist

## Additional Context


apps/hubble/src/storage/stores/storeEventHandler.ts
```diff
@@ -154,7 +155,36 @@ class StoreEventHandler extends TypedEmitter<StoreEvents> {
...
+          const event = HubEvent.create({ ...args, id: eventId.value });
+          // TODO: validate event
+          events.push(event);
+          txn = putEventTransaction(txn, event);
+        }
+
+        await this._db.commit(txn);
+
+        for (const event of events) {
+          void this.broadcastEvent(event);
```
deodad, line 177: what does this `void` do given `broadcastEvent` is sync?

Issue #715, opened 2023-03-21
ethers.Wallet returning type error with EthersEip712Signer


pfletcherhill:
**What is the bug?**
Example replit: https://replit.com/@pfletcherhill/LikableTeemingNormalform#index.ts

<img width="1411" alt="Screen Shot 2023-03-21 at 2 41 39 PM" src="https://user-images.githubusercontent.com/1607180/226710442-a6d65314-6e71-4753-8a57-fff43fb90048.png">

PR #716, opened 2023-03-21
chore: use subset of ethers Signer class

deodad:
Use type-only import (e.g. `import type`) to import `Signer` from ethers. This will avoid pulling in the entire `Signer` class and its dependencies at compile time.

Add `MinimalEthersSigner` that only contains the class we care about.


deodad:
closes #715

PR #717, opened 2023-03-21
feat: add storage cache to speed up pruning

pfletcherhill:
## Motivation

Keep track of message counts per fid and store to speed up pruning.

## Change Summary

* Add `StorageCache` class that when started keeps track of the number of messages for each fid/store in memory
* Use values from storage cache when pruning to avoid having to count messages live
* Make storage engine startable and stoppable, and start validation worker and add listeners on start

## Merge Checklist

## Additional Context

### DIFF
--- a/apps/hubble/src/storage/engine/index.ts
+++ b/apps/hubble/src/storage/engine/index.ts
@@ -91,10 +91,20 @@ class Engine {
       logger.warn({ workerPath, e }, 'failed to create validation worker, falling back to main thread');
     }
 
+    this.eventHandler.on('mergeIdRegistryEvent', this.handleMergeIdRegistryEvent.bind(this));
+    this.eventHandler.on('mergeMessage', this.handleMergeMessageEvent.bind(this));
+    this.eventHandler.on('revokeMessage', this.handleRevokeMessageEvent.bind(this));
+    this.eventHandler.on('pruneMessage', this.handlePruneMessageEvent.bind(this));
+
     await this._storageCache.syncFromDb(this._db);
   }
 
   async stop(): Promise<void> {
+    this.eventHandler.off('mergeIdRegistryEvent', this.handleMergeIdRegistryEvent.bind(this));
+    this.eventHandler.off('mergeMessage', this.handleMergeMessageEvent.bind(this));
+    this.eventHandler.off('revokeMessage', this.handleRevokeMessageEvent.bind(this));
+    this.eventHandler.off('pruneMessage', this.handlePruneMessageEvent.bind(this));
+
     if (this._validationWorker) {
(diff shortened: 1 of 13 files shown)


apps/hubble/src/storage/engine/storageCache.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    this._usage = usage;
+    this._synced = true;
+    log.info('storage cache synced');
+  }
+
+  getMessageCount(fid: number, set: UserMessagePostfix): HubResult<number> {
+    if (this._synced !== true) {
+      return err(new HubError('unavailable.storage_failure', 'storage cache is not synced with db'));
```
deodad, line 53: Consider dropping a log statement here since the callers ignore the error

apps/hubble/src/storage/engine/index.ts
```diff
@@ -656,6 +675,58 @@ class Engine {
...
+    return ok(undefined);
+  }
+
+  private async handlePruneMessageEvent(event: protobufs.PruneMessageHubEvent): HubAsyncResult<void> {
+    const { message } = event.pruneMessageBody;
+
+    if (protobufs.isSignerAddMessage(message)) {
+      const revokeResult = await this.revokeMessagesBySigner(message.data.fid, message.data.signerAddBody.signer);
+      if (revokeResult.isErr()) {
+        return err(revokeResult.error);
```
deodad, line 711: it doesn't seem like this should ever get executed but consider logging since these are getting called by event emitters and errors wont get noticed

apps/hubble/src/storage/engine/storageCache.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+    const key = makeKey(fid, set);
+    return ok(this._usage.get(key) ?? 0);
+  }
+
+  processEvent(event: HubEvent): HubResult<void> {
+    if (this._synced !== true) {
+      const error = new HubError('unavailable.storage_failure', 'storage cache is not synced with db');
+      log.error({ errCode: error.errCode }, `processEvent error: ${error.message}`);
+      return err(error);
```
deodad, line 63: Maybe unnecessary but could set another bit to indicate the cache failed to process an event and will no longer be accurate


I don't think this would happen given your implementation but seems like its theoretically possible

deodad (commented):
It looks good, a few comments

I'll look over again and play around with the tests tomorrow

apps/hubble/src/storage/engine/storageCache.ts
pfletcherhill: I think the `synced` attribute is sufficient for now, but we should keep an eye on the logs

PR #754, opened 2023-03-25
docs: Use SSL or Insecure clients in examples

adityapk00:
## Motivation

Using `getHubRpcClient` can be slow because it attempts a SSL connection first, and has to wait for a timeout. Update the docs to use `getSSLHubRpcClient` or `getInsecureHubRpcClient` directly. 

## Change Summary

- Update docs

## Merge Checklist

### DIFF
--- a/.changeset/chilly-bats-kneel.md
+++ b/.changeset/chilly-bats-kneel.md
@@ -1,8 +1,8 @@
 ---
-'@farcaster/hub-nodejs': patch
-'@farcaster/protobufs': patch
-'@farcaster/utils': patch
-'@farcaster/hubble': patch
+'@farcaster/hub-nodejs': minor
+'@farcaster/protobufs': minor
+'@farcaster/utils': minor
+'@farcaster/hubble': minor
 ---
 
 Remove getHubRpcClient, use getSSLRpcClient() or getInsecureRpcClient()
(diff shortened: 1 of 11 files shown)


varunsrin (commented):
1. should we deprecate `getHubRpcClient` entirely? the slower performance and unclear security behavior seem like they might lead users to footguns. if we do this, we'll need to bump the minor version in the package. 

2. connecting to a non-SSL hub with ssl or vice versa returns the error `  error: HubError: No connection established`. Is it possible we can throw a more descriptive error here to help users understand that they are using the wrong authentication method?

adityapk00:
For 1. Yes, we probably should. I'll update this PR with deprecating that. 
For 2. Unfortunately we don't have more info here, since we don't know for sure that the wrong method was used. We can't differentiate between:
a. Connecting to SSL but server is not SSL (or vice-versa)
b. Couldn't reach the server or Could reach the server, but there wasn't anything listening on the port. 

Hence the generic message. We could add a hint here, saying something like "No connection established. Are you sure you are using SSL?" or something. Does that make sense?

.changeset/chilly-bats-kneel.md
```diff
@@ -0,0 +1,8 @@
+---
+'@farcaster/hub-nodejs': patch
+'@farcaster/protobufs': patch
+'@farcaster/utils': patch
```
varunsrin, line 4: should be minor version for hub-nodejs and utils since we're "breaking" the API

adityapk00: Oh, right. Sorry, will fix

PR #787, opened 2023-03-30
feat: implement ViemLocalEip712Signer

### LINKED ISSUE #777: feat(hub-nodejs): add ViemEip712 Signer
Allow creating an Eip712Signer with Viem.

Users who use Viem can create signers using those instances, instead of having to generate Ethers wallets. Note that ethers will still be a dependency since our package needs it for other reasons. 

No

michaelhly:
## Motivation
closes #777 

## Change Summary
Implement an Eip712Signer using Viem

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR adds a new `ViemLocalEip712Signer` class and makes several changes to the `core` package. 

### Detailed summary
- Adds `ViemLocalEip712Signer` class to `core/src/signers`
- Updates `core/package.json` to include `viem` dependency
- Updates `core/src/crypto/eip712.ts` to add a type cast for compatibility with `viem`
- Updates `core/src/signers/viemLocalEip712Signer.test.ts` to test `ViemLocalEip712Signer`
- Updates `core/src/signers/nobleEd25519Signer.test.ts` to import `ok` from `neverthrow`
- Updates `core/src/signers/viemLocalEip712Signer.ts` to implement `Eip712Signer` and add methods for signing message hashes and verification claims
- Updates `core/src/signers/testUtils.ts` to add a `testEip712Signer` function for testing `Eip712Signer` implementations
- Updates `core/src/signers/ethersV5Eip712Signer.test.ts` and `core/src/signers/ethersEip712Signer.test.ts` to use `testEip712Signer` function instead of manually testing methods.

> The following files were skipped due to too many changes: `packages/core/src/signers/ethersEip712Signer.test.ts`, `yarn.lock`

### DIFF
--- a/packages/core/src/signers/viemLocalEip712Signer.ts
+++ b/packages/core/src/signers/viemLocalEip712Signer.ts
@@ -10,7 +10,7 @@ import { HubAsyncResult, HubError } from '../errors';
 import { VerificationEthAddressClaim } from '../verifications';
 import { Eip712Signer } from './eip712Signer';
 
-export class ViemEip712Signer extends Eip712Signer {
+export class ViemLocalEip712Signer extends Eip712Signer {
   private readonly _viemLocalAccount: LocalAccount<string>;
 
   constructor(viemLocalAccount: LocalAccount<string>) {
--- a/packages/core/src/signers/viemLocalEip712Signer.ts
+++ b/packages/core/src/signers/viemLocalEip712Signer.ts
@@ -1,5 +1,6 @@
 import { ResultAsync } from 'neverthrow';
 import { LocalAccount } from 'viem/accounts';
+import { bytesToHex } from 'viem/utils';
 import { hexStringToBytes } from '../bytes';
 import {
   EIP_712_FARCASTER_MESSAGE_DATA,
@@ -43,7 +44,7 @@ export class ViemLocalEip712Signer extends Eip712Signer {
         types: { MessageData: EIP_712_FARCASTER_MESSAGE_DATA },
         primaryType: 'MessageData',
         message: {
-          hash,
+          hash: bytesToHex(hash),
         },
       }),
       (e) => new HubError('bad_request.invalid_param', e as Error)
--- a/packages/core/src/signers/viemLocalEip712Signer.ts
+++ b/packages/core/src/signers/viemLocalEip712Signer.ts
@@ -19,14 +19,6 @@ export class ViemLocalEip712Signer extends Eip712Signer {
     this._viemLocalAccount = viemLocalAccount;
   }
 
-  private _toViemCompat712Domain() {
-    const { salt, ...rest } = EIP_712_FARCASTER_DOMAIN;
-    return {
-      salt: salt as `0x${string}`,
-      ...rest,
-    };
-  }
-
   public async getSignerKey(): HubAsyncResult<Uint8Array> {
     return ResultAsync.fromPromise(
       Promise.resolve(this._viemLocalAccount.address),
@@ -37,7 +29,7 @@ export class ViemLocalEip712Signer extends Eip712Signer {
   public async signMessageHash(hash: Uint8Array): HubAsyncResult<Uint8Array> {
     const hexSignature = await ResultAsync.fromPromise(
       this._viemLocalAccount.signTypedData({
-        domain: this._toViemCompat712Domain(),
+        domain: EIP_712_FARCASTER_DOMAIN,
         types: { MessageData: EIP_712_FARCASTER_MESSAGE_DATA },
         primaryType: 'MessageData',
         message: {
@@ -52,7 +44,7 @@ export class ViemLocalEip712Signer extends Eip712Signer {
   public async signVerificationEthAddressClaim(claim: VerificationEthAddressClaim): HubAsyncResult<Uint8Array> {
     const hexSignature = await ResultAsync.fromPromise(
       this._viemLocalAccount.signTypedData({
-        domain: this._toViemCompat712Domain(),
+        domain: EIP_712_FARCASTER_DOMAIN,
         types: { VerificationClaim: EIP_712_FARCASTER_VERIFICATION_CLAIM },
         primaryType: 'VerificationClaim',
         message: {
--- a/packages/core/src/signers/viemLocalEip712Signer.ts
+++ b/packages/core/src/signers/viemLocalEip712Signer.ts
@@ -27,14 +27,11 @@ export class ViemLocalEip712Signer extends Eip712Signer {
     };
   }
 
-  private async _getSignerKey() {
-    return this._viemLocalAccount.address;
-  }
-
   public async getSignerKey(): HubAsyncResult<Uint8Array> {
-    return ResultAsync.fromPromise(this._getSignerKey(), (e) => new HubError('unknown', e as Error)).andThen(
-      hexStringToBytes
-    );
+    return ResultAsync.fromPromise(
+      Promise.resolve(this._viemLocalAccount.address),
+      (e) => new HubError('unknown', e as Error)
+    ).andThen(hexStringToBytes);
   }
 
   public async signMessageHash(hash: Uint8Array): HubAsyncResult<Uint8Array> {
--- a/.changeset/angry-fireants-melt.md
+++ b/.changeset/angry-fireants-melt.md
@@ -1,5 +1,5 @@
 ---
-'@farcaster/utils': minor
+'@farcaster/core': patch
 ---
 
 feat: add ViemLocalEip712Signer
--- a/.changeset/angry-fireants-melt.md
+++ b/.changeset/angry-fireants-melt.md
@@ -2,4 +2,4 @@
 '@farcaster/utils': minor
 ---
 
-feat: add ViemEip712Signer
+feat: add ViemLocalEip712Signer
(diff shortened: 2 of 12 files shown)


michaelhly:
Ran into:
https://github.com/wagmi-dev/viem/blob/1140b1fe189db8296e8eb62a713fd611c7057a9e/src/utils/data/size.ts#L10

```sh
      TypeError: Cannot read properties of undefined (reading 'length')
          at size (/Users/michaelhly/Projects/hubble/node_modules/viem/src/utils/data/size.ts:12:16)
          at validateData (/Users/michaelhly/Projects/hubble/node_modules/viem/src/actions/wallet/signTypedData.ts:169:22)
          at validateTypedData (/Users/michaelhly/Projects/hubble/node_modules/viem/src/actions/wallet/signTypedData.ts:187:3)
          at signTypedData (/Users/michaelhly/Projects/hubble/node_modules/viem/src/actions/wallet/signTypedData.ts:104:3)
          at Object.signTypedData (/Users/michaelhly/Projects/hubble/node_modules/viem/src/clients/decorators/wallet.ts:90:54)
          at ViemEip712Signer.signTypedData [as signVerificationEthAddressClaim] (/Users/michaelhly/Projects/hubble/packages/utils/src/signers/viemEip712Signer.ts:73:30)
          at processTicksAndRejections (node:internal/process/task_queues:95:5)
          at Object.<anonymous> (/Users/michaelhly/Projects/hubble/packages/utils/src/signers/viemEip712Signer.test.ts:70:22)
```

packages/utils/src/signers/viemEip712Signer.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+];
+
+describe('ViemEip712Signer', () => {
+  let signer: ViemEip712Signer;
+  let signerKey: Uint8Array;
+  let ethersSigner;
+
+  beforeAll(async () => {
+    const localAccount = getAccount(new Wallet(accounts[0].privateKey)) as LocalAccount;
```
michaelhly, line 27: viem currently does not have client-side signing utilities:
https://viem.sh/docs/clients/wallet.html#set-up-your-local-account

packages/utils/src/signers/viemEip712Signer.test.ts
```diff
@@ -0,0 +1,89 @@
...
+describe('ViemEip712Signer', () => {
+  const localAccount = getAccount(Wallet.createRandom()) as LocalAccount;
+  let signer: ViemEip712Signer;
+  let signerKey: Uint8Array;
+  let walletClient: WalletClient;
+
+  beforeAll(async () => {
+    walletClient = createWalletClient({
+      transport: custom({
+        // Mock RPC server behavior
```
michaelhly, line 32: viem currently does not have client-side signing utilities, so we mock the server behavior with a local account (ethers):
https://viem.sh/docs/clients/wallet.html#set-up-your-local-account

packages/utils/src/signers/viemEip712Signer.test.ts
```diff
@@ -0,0 +1,89 @@
...
+describe('ViemEip712Signer', () => {
+  const localAccount = getAccount(Wallet.createRandom()) as LocalAccount;
+  let signer: ViemEip712Signer;
+  let signerKey: Uint8Array;
+  let walletClient: WalletClient;
+
+  beforeAll(async () => {
+    walletClient = createWalletClient({
+      transport: custom({
+        // Mock RPC server responses
```
michaelhly, line 32: viem currently does not have client-side signing utilities, so we mock the server responses with a local account (which uses ethers.js):
[viem.sh/docs/clients/wallet.html#set-up-your-local-account](https://viem.sh/docs/clients/wallet.html#set-up-your-local-account)

jxom: FYI: we just released viem 0.2 (Twitter announcement soon!) which includes [private key accounts](https://viem.sh/docs/accounts/privateKey.html) (with respective signing utils).

deodad:
thanks for the PR @michaelhly! planning on trying this out tomorrow or friday. 

are you able to do the 0.2.0 upgrade [suggested](https://github.com/farcasterxyz/hub-monorepo/pull/787#discussion_r1155804930) by @jxom? would be great to replace the mock with a private key account

packages/utils/src/signers/viemEip712Signer.test.ts
deodad: @jxom our use case doesn't require interacting with Ethereum, we just need the Account portion.

Is there a way to instantiate a JSON-RPC account without using the full WalletClient? Ideally we could use the `Account` union type but have access to `signMessage` and `signTypedData` if it's either an RPC or local account.

jxom: The account itself has those methods (ie. the account initialised from `privateKeyToAccount`) so you can sign client-side. We should probably document that.

However, for JSON-RPC accounts, you’d need to set up a wallet client with a transport because you request a signature over JSON-RPC, so viem needs to know how to do that (the transport).

michaelhly:
> thanks for the PR @michaelhly! planning on trying this out tomorrow or friday.
> 
> are you able to do the 0.2.0 upgrade [suggested](https://github.com/farcasterxyz/hub-monorepo/pull/787#discussion_r1155804930) by @jxom? would be great to replace the mock with a private key account

Seems like there were some non-backwards compatible changes between 0.1.23 and 0.3.x. Currently don't have time to do the upgrade ... happy to close this PR and let someone have a shot at it!

jxom:
Happy to help here. The migration to 0.3.x really should not take long at all. We even have a migration guide: https://viem.sh/docs/migration-guide.html

michaelhly:
> Happy to help here. The migration to 0.3.x really should not take long at all. We even have a migration guide: [viem.sh/docs/migration-guide.html](https://viem.sh/docs/migration-guide.html)

I took a look, somehow these two methods don't produce the same signature:
Ethers5.7 - https://github.com/ethers-io/ethers.js/blob/v5.7/packages/signing-key/src.ts/index.ts#L54-L66
Viem0.3.2 - https://github.com/wagmi-dev/viem/blob/d19791425aac446651467c73a8b2273584892b86/src/accounts/utils/sign.ts#L20-L30

I suspect it's a padding problem ...

michaelhly:
@jxom you can uncomment this test to reproduce the signing issue:
https://github.com/michaelhly/hubble/blob/b4e40c9235c32968088e716637c906d72a81fa37/packages/core/src/signers/viemEip712Signer.test.ts#L68-L120

jxom:
Interesting. I took a test vector and ran it in an isolated viem environment and was able to recover the signing address successfully: https://github.com/wagmi-dev/viem/compare/jxom/hubble-repro?expand=1

michaelhly:
> Interesting. I took a test vector and ran it in an isolated viem environment and was able to recover the signing address successfully: [wagmi-dev/viem@`jxom`?expand=1 (compare)](https://github.com/wagmi-dev/viem/compare/jxom/hubble-repro?expand=1)
> 
> FWIW, I also compared signatures across ethers.js v6 & viem in that environment, and they were the same.
> 
> Can potentially take a look at the hub implementation when I get some time.

Thanks for this example! This was the fix:
[`aa096c5` (#787)](https://github.com/farcasterxyz/hub-monorepo/pull/787/commits/aa096c5629bad4e62031fefe9efb4f27cb74720b)

🙏 🙏 🙏

packages/core/src/signers/viemLocalEip712Signer.ts
```diff
@@ -0,0 +1,58 @@
...
+    return hexSignature.andThen((hex) => hexStringToBytes(hex));
+  }
+
+  public async signVerificationEthAddressClaim(claim: VerificationEthAddressClaim): HubAsyncResult<Uint8Array> {
+    const hexSignature = await ResultAsync.fromPromise(
+      this._viemLocalAccount.signTypedData({
+        domain: EIP_712_FARCASTER_DOMAIN,
+        types: { VerificationClaim: EIP_712_FARCASTER_VERIFICATION_CLAIM },
+        primaryType: 'VerificationClaim',
+        message: {
```
deodad, line 50: nit: could this just be `message: claim` instead of destructuring?

deodad:
Can you [sign the commits](https://github.com/farcasterxyz/hub-monorepo/blob/main/CONTRIBUTING.md#22-signing-commits) and I'll merge this

packages/core/src/signers/viemLocalEip712Signer.ts
michaelhly: Passing in just `claim` causes an issue here https://github.com/wagmi-dev/viem/blob/d19791425aac446651467c73a8b2273584892b86/src/utils/typedData.ts#L50

ran into 
```
      TypeError: Cannot read properties of undefined (reading 'length')
          at size (/Users/michaelhly/Projects/hubble/node_modules/viem/src/utils/data/size.ts:12:16)
          at validateData (/Users/michaelhly/Projects/hubble/node_modules/viem/src/utils/typedData.ts:50:22)
```

deodad: got it thanks for the context

michaelhly:
@deodad signed!

Issue #573, opened 2023-02-15
feat: @farcaster/js should run in browsers 


varunsrin:
**What is the feature you would like to implement?**

Get @farcaster/js compiling in browsers

**Why is this feature important?**

Single-page and stand alone client applications will be able to interface directly with Hubs without a server. 

**Will the protocol spec need to be updated??**

No

**Additional context**

- Can @farcaster/js just use grpc-web instead?
- Can @farcaster/hubble run a server that runs both grpc-js and grpc-web?
- Can @farcaster/js compile in browsers end to end?

pfletcherhill:
Related other tickets to consolidate:
* https://github.com/farcasterxyz/hubble/issues/424
* https://github.com/farcasterxyz/hubble/issues/497
* https://github.com/farcasterxyz/hubble/issues/473
* https://github.com/farcasterxyz/hubble/issues/494

pfletcherhill:
Supporting web browsers isn't an immediate problem, but in the medium term it lowers the barrier to entry for community members looking to make something using farcaster, which is good. I imagine some devs will want very basic access and choose to use the hub RPC for reading/writing and others will use more feature-rich APIs or build their own backends for indexing data. There's certainly an opportunity for community members to make their own indexing services and APIs on top of hubs for many use cases, similar to Alchemy's various APIs for ethereum.

We should support both grpc and grpc-web simultaneously using a proxy (I made it work with Envoy, as you mentioned, though there may be other options too). So hub-to-hub communication will continue to use gRPC directly.

varunsrin:
we should see whether farcaster/js works by default in serverless platforms like vercel. if it does, we can punt this to a later milestone

pfletcherhill:
#### High-level goal:
Create a new package called `@farcaster/hub-web` that exports the same methods as `@farcaster/hub-nodejs` except with a grpc-web client instead of a grpc one that works in browsers.

#### Work to be done:
* Create new `@farcaster/hub-web` package in the monorepo
* Generate grpc-web compatible client from the existing `rpc.proto` schema and export it from the new package. If using ts-proto, we can use the `outputClientImpl=grpc-web` option ([see documentation](https://github.com/stephenh/ts-proto)).
* Wrap the grpc-web client implementation in promises and results (see [client.ts](https://github.com/farcasterxyz/hub-monorepo/blob/main/packages/utils/src/client.ts) in utils for reference)
[...]

PR #797, opened 2023-04-03
feat: add hub-web package to only depend on grpc-web

PangZhi:
## Motivation

To address #573, it sets up a new hub-web package, add grpc-web support and envoy support. 
Applications would be able to call hub endpoints from a browser environment

## Change Summary

- add support for envoy
- expose RpcWebClient
- add new hub-web package and generate code from grpc-web

## Merge Checklist

## Additional Context

Follow ups:
- refactor protobufs/utils
- wrap responses: current responses (Promise<T>) are slightly different from hub-nodejs (Promise<HubResult<T>>)


envoy/envoy.yaml
```diff
@@ -0,0 +1,64 @@
...
+                    typed_config:
+                      '@type': type.googleapis.com/envoy.extensions.filters.http.grpc_web.v3.GrpcWeb
+                  - name: envoy.filters.http.cors
+                    typed_config:
+                      '@type': type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors
+                  - name: envoy.filters.http.router
+                    typed_config:
+                      '@type': type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+  clusters:
+    - name: greeter_service
```
sds, line 49: Do we want to call this `rpc_service`?

envoy/envoy-start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+docker run -d -v "$(pwd)"/envoy.yaml:/etc/envoy/envoy.yaml:ro \
+    --network=host envoyproxy/envoy:v1.22.0
+
+# command to run in Mac/Windows
+# docker run -d -v "$(pwd)"/envoy.yaml:/etc/envoy/envoy.yaml:ro \
+#    -p 2284:2284 -p 9901:9901 envoyproxy/envoy:v1.22.0
```
sds, line 8: What is port 9901 used for in the Mac/Windows context? Admin port? Do we actually want to expose that by default?

pfletcherhill (commented):
Overall looks great! I just added a few questions.

envoy/envoy.yaml
PangZhi: opps yes!

envoy/envoy-start.sh
PangZhi: this is admin port, we can actually remove it if necessary. In Mac/Windows we just can't use --network=host

packages/hub-web/package.json
```diff
@@ -0,0 +1,35 @@
...
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "license": "MIT",
+  "scripts": {
+    "build": "tsup --config tsup.config.ts",
+    "clean": "rimraf ./dist",
+    "protoc": "protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto  --ts_proto_out=./src/generated/ --ts_proto_opt=esModuleInterop=true,exportCommonSymbols=false,outputClientImpl=grpc-web,useOptionals=none,unrecognizedEnum=false,removeEnumPrefix=true --proto_path=../protobufs/src/schemas ../protobufs/src/schemas/*",
```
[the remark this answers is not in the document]
PangZhi: Currently we have to generate for all as we haven't refactored protobufs/utils. If we don't generate for all we have to import those from protobufs/utils and that will include dependency on rpc-js.

After we refactor protobufs/utils, we will only have to generated for rpc.proto, that will come in the following PR

packages/hub-web/src/generated/rpc.ts
```diff
@@ -0,0 +1,3153 @@
+/* eslint-disable */
+import { grpc } from '@improbable-eng/grpc-web';
+import { BrowserHeaders } from 'browser-headers';
+import Long from 'long';
+import _m0 from 'protobufjs/minimal';
+import { Observable } from 'rxjs';
```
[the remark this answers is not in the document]
PangZhi: rxjs is used for streaming by grpc. Currently we don't need to explicitly add it as dependency

envoy/README.md
```diff
@@ -0,0 +1,36 @@
+In order for farcaster to support grpc-web, we need to set up envoy alongside the server.
```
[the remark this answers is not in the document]
PangZhi: that's a good idea!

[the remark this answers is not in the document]
pfletcherhill: Now that we've moved it under apps/hubble, we should remove these files from here right?

packages/hub-web/src/generated/rpc.ts
```diff
@@ -0,0 +1,3153 @@
...
+  ): IdRegistryEventByAddressRequest {
+    const message = createBaseIdRegistryEventByAddressRequest();
+    message.address = object.address ?? new Uint8Array();
+    return message;
+  },
+};
+
+export interface HubService {
+  /** Submit Methods */
+  SubmitMessage(request: DeepPartial<Message>, metadata?: grpc.Metadata): Promise<Message>;
```
pfletcherhill, line 1892: Minor thing...the grpc methods from grpc-js start with lowercase letters. We should decide whether it's important for the method names to be formatted the same across grpc-js and grpc-web or not. My slight bias would be for them to be the same so that we can share as much documentation as possible.

pfletcherhill (approved):
Looks good. I just had two minor follow-up comments, but feel free to merge once you address those.

PR #798, opened 2023-04-03
feat: refactor protobufs and utils to remove grpc-js dependency

PangZhi:
## Motivation

Part of #573, removed grpc-js dependency from protobufs (thus utils). 

## Change Summary
- separate rpc.proto to request_response.proto and rpc.proto (only have service definition), remove grpc-js dependency from protobufs (thus utils)
- move all grpc-js code generation to hub-nodejs
- refactor hubble app to use hub-nodejs
- add protobufs & utils as dependency for hub-web and add new exports

## Merge Checklist


.changeset/honest-news-own.md
```diff
@@ -0,0 +1,10 @@
+---
+'@farcaster/protobufs': minor
```
PangZhi, line 2: ideally it's major change for protobufs and utils (interface not backward compatible). However, it seems that we haven't released major version of protobufs  & utils yet, so I made it minor.

apps/hubble/package.json
```diff
@@ -47,6 +47,7 @@
   "dependencies": {
     "@chainsafe/libp2p-gossipsub": "6.1.0",
     "@chainsafe/libp2p-noise": "^11.0.0 ",
+    "@farcaster/hub-nodejs": "^0.6.0",
```
PangZhi, line 50: What's the process for package release here? We will need to update this to the new version of hub-nodejs

pfletcherhill: We use changesets for version control, so we can run `yarn changeset version` and it will update all of these correctly.

apps/hubble/src/rpc/test/bulkService.test.ts
```diff
@@ -1,5 +1,6 @@
 import * as protobufs from '@farcaster/protobufs';
-import { Factories, getInsecureHubRpcClient, HubResult, HubRpcClient } from '@farcaster/utils';
+import { Factories, HubResult } from '@farcaster/utils';
```
pfletcherhill, line 2: What's the reason to continue importing these from `@farcaster/utils` vs from `@farcaster/hub-nodejs` since hub-nodejs also exports everything from utils?

packages/protobufs/src/schemas/request_response.proto
```diff
@@ -0,0 +1,124 @@
+syntax = "proto3";
```
pfletcherhill, line 1: What's the reason for splitting rpc.proto into two files?

packages/protobufs/package.json
```diff
@@ -18,7 +18,7 @@
   "scripts": {
     "build": "tsup --config tsup.config.ts",
     "clean": "rimraf ./dist",
-    "protoc": "protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_out=./src/generated/ --ts_proto_opt=esModuleInterop=true,exportCommonSymbols=false,outputServices=grpc-js,useOptionals=none,unrecognizedEnum=false,removeEnumPrefix=true --proto_path=./src/schemas ./src/schemas/*",
+    "protoc": "protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_out=./src/generated/ --ts_proto_opt=esModuleInterop=true,exportCommonSymbols=false,outputServices=false,useOptionals=none,unrecognizedEnum=false,removeEnumPrefix=true --proto_path=./src/schemas ./src/schemas/*",
```
pfletcherhill, line 21: Could we get rid of `@farcaster/protobufs` entirely and just add a protoc command to `@farcaster/utils` to generate the non-service protobuf classes for itself?

pfletcherhill (commented):
Exciting! I think the most important question to answer is whether we want hubble to import anything from `@farcaster/protobufs` or `@farcaster/utils` rather than getting everything from `@farcaster/hub-nodejs`. Right now it seems somewhat arbitrary which package a file picks for each import.

packages/protobufs/src/schemas/request_response.proto
PangZhi: Personally feel this is slightly cleaner pattern: share whatever can be shared.
Where hub-nodejs and hub-web are different is the service/client generation part (exactly the rpc.proto after splitting). After splitting, we will be able to share like SyncIds and Empty from protobufs

packages/protobufs/package.json
PangZhi: Yes this is exactly next step: completely merge these two into a single package. Was planning to do it in separate PR.

PangZhi:
@sds do you know how to let codecov/path and codecov/project skip generated code from grpc? I think now they're failing because of generated code.

sds:
I didn't set this tool up, so I'm not familiar with (honestly didn't even know this existed until now 🙂). 

It _looks_ like we could specify the [`ignore` option](https://docs.codecov.com/docs/codecovyml-reference#ignore) in `codecov.yml` to ignore the directories with generated files?

varunsrin:
fine to merge without code coverage requirements, theyre not meant to be blocking just a guide 

but it would be nice if we added that to the ignore file to keep the code coverage metrics useful

PangZhi:
> codecov.yml

I

> I didn't set this tool up, so I'm not familiar with it (honestly didn't even know this existed until now 🙂).
> 
> It _looks_ like we could specify the [`ignore` option](https://docs.codecov.com/docs/codecovyml-reference#ignore) in `codecov.yml` to ignore the directories with generated files?
> 
> CC @pfletcherhill @varunsrin how much do we care about code coverage requirements being met?

I think the codecov ignored the previous generated code in the protobufs package, but it's not in the codecov.yml. We should definitely fix so the metric is useful.

apps/hubble/src/console/protobufCommand.ts
```diff
@@ -1,5 +1,5 @@
-import * as protobufs from '@farcaster/protobufs';
-import { Factories } from '@farcaster/utils';
+import * as hub_nodejs from '@farcaster/hub-nodejs';
```
pfletcherhill, line 1: Nit: can you camelCase this variable?

packages/hub-web/src/index.ts
```diff
@@ -1,7 +1,5 @@
-export * from './generated/hub_event';
-export * from './generated/id_registry_event';
-export * from './generated/message';
-export * from './generated/name_registry_event';
-export * from './generated/rpc';
+export * from '@farcaster/protobufs';
+export * from '@farcaster/utils';
 
+export * from './generated/rpc';
```
pfletcherhill, line 4: Do we need to export from `./generated/request_response` too? Or does the export of that file from `@farcaster/protobufs` work with the services exported from `./generated/rpc`?

pfletcherhill (approved):
Looks good to me! One minor change and a question, otherwise ready to merge.

packages/hub-web/src/index.ts
PangZhi: I tried locally using FidRequest (one of the request_response), so the export was from @farcaster/protobufs but things work well when I fid into rpc request.

PR #799, opened 2023-04-03
fix: Prevent slow subscribers from crashing the hub

adityapk00:
## Motivation

When clients don't consume events from the subscribe() RPC fast enough, pending writes to the grpc stream can clog up memory, crashing the client. 

## Change Summary

- Write to subscribers with a timeout of 10s
- Close stream if a event subscriber is too slow. 

## Merge Checklist

## Additional Context

### DIFF
--- a/.changeset/pretty-llamas-grow.md
+++ b/.changeset/pretty-llamas-grow.md
@@ -1,0 +1,5 @@
+---
+'@farcaster/hubble': patch
+---
+
+Write events to subscribe() with a timeout to prevent slow clients from clogging up memory
--- a/apps/hubble/src/rpc/server.ts
+++ b/apps/hubble/src/rpc/server.ts
@@ -42,6 +42,7 @@ import { addressInfoFromParts } from '~/utils/p2p';
 import { RateLimiterMemory, RateLimiterAbstract } from 'rate-limiter-flexible';
 
 export type RpcUsers = Map<string, string[]>;
+export const SUBSCRIBE_TIMEOUT_MS = 10_000; // 10 seconds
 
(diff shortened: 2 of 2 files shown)


deodad:
Do you know what the high watermark is or approx how many events will be pushed into the stream before `write` returns false? Depending on how much data can be buffered, we might need a shorter or longer timeout.

The scenario I'm thinking of is:
- subscribe is called with no fromId
- hub quickly writes a large amount of data (let's say 1000 events)
- consumer starts pulling but doesn't get through so many events in 10 seconds

adityapk00:
Ran some tests, and on Linux, ~1100 HubEvents are cached in the stream before it starts returning false, after which the 10s timeout is triggered. So it should be able to handle some network issues, but prevent the hub from getting overwhelmed. 

While I was doing the tests, I realized there's another problem which is the way in which the eventListner works. There is no backpressure for the event listner, so as events are merged, they are sent to the `eventListener` in `rpcServer.subscribe()`. And if the client is working but slow, everytime the stream backs up, the rpcServer stops sending messages for a while, the client drains the stream, and then the server resumes sending. However, since the client is consuming the stream at a slower rate than messages are being merged, each merge is causing a new `eventListener` closure waiting, causing an OOM. 

I modified this a bit into a new `BufferedStreamWriter` class, that handles both the 10s timeout, and has an additional 1000 message buffer. Exceeding either will close the stream.

deodad:
You might have already seen this but I just remembered reading it some time ago, dropping it here in case it's useful
https://nodejs.org/en/docs/guides/backpressuring-in-streams

PR #822, opened 2023-04-05
chore: export getAuthMetadata & getHubRpcClient

PangZhi:
## Motivation

export getAuthMetadata & getHubRpcClient for hub-web

## Change Summary

export getAuthMetadata & getHubRpcClient for hub-web

## Merge Checklist

### DIFF
--- a/packages/hub-web/README.md
+++ b/packages/hub-web/README.md
@@ -28,7 +28,7 @@ import { getHubRpcClient } from '@farcaster/hub-web';
   // if you are testing from a node environment
   // const client = getHubRpcClient('https://testnet1.farcaster.xyz:2284', false);
 
-  // if you are testing from a braowser environment
+  // if you are testing from a browser environment
   const client = getHubRpcClient('https://testnet1.farcaster.xyz:2284');
 
   const castsResult = await client.getCastsByFid({ fid: 7884 });
--- a/packages/hub-web/README.md
+++ b/packages/hub-web/README.md
@@ -22,14 +22,14 @@ pnpm install @farcaster/hub-web
 ### Fetching Data from Hubs
 
 ```typescript
-import { getInsecureHubRpcClient } from '@farcaster/hub-web';
+import { getHubRpcClient } from '@farcaster/hub-web';
 
 (async () => {
   // if you are testing from a node environment
-  // const client = getInsecureHubRpcClient('testnet1.farcaster.xyz:2284', false);
+  // const client = getHubRpcClient('https://testnet1.farcaster.xyz:2284', false);
 
   // if you are testing from a braowser environment
-  const client = getInsecureHubRpcClient('testnet1.farcaster.xyz:2284');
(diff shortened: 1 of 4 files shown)


packages/hub-web/src/client.ts
```diff
@@ -98,3 +102,9 @@ export const getAdminRpcClient = (address: string, isBrowser = true): AdminRpcCl
 const getRpcWebClient = (address: string, isBrowser = true): GrpcWebImpl => {
   return new GrpcWebImpl(address, isBrowser ? {} : { transport: NodeHttpTransport() });
 };
+
+export const getAuthMetadata = (username: string, password: string): grpc.Metadata => {
+  const metadata = new grpc.Metadata();
+  metadata.set('authorization', `Basic ${Buffer.from(`${username}:${password}`).toString('base64')}`);
```
pfletcherhill, line 108: I don't think Buffer is available in browsers, so we should use `btoa` or another method for encoding the username and password

packages/hub-web/src/client.ts
```diff
@@ -98,3 +102,9 @@ export const getAdminRpcClient = (address: string, isBrowser = true): AdminRpcCl
 const getRpcWebClient = (address: string, isBrowser = true): GrpcWebImpl => {
```
pfletcherhill, line 102: Can you also refactor these methods to just be `getHubRpcClient` and respect whatever address the user passes in rather than prefixing the input with http or https?

packages/hub-web/src/client.ts
```diff
@@ -14,6 +14,10 @@ import { HubError, HubErrorCode, HubResult } from '@farcaster/core';
 import { Observable, of } from 'rxjs';
 import { catchError } from 'rxjs/operators';
 
+export const Code = grpc.Code;
+export const Metadata = grpc.Metadata;
+export type { Observable } from 'rxjs';
+
 const fromServiceError = (err: GrpcWebError): HubError => {
```
pfletcherhill, line 21: How about also using this PR to fix this method to return the right errCode in the HubError?

pfletcherhill (changes requested):

packages/hub-web/README.md
```diff
@@ -22,14 +22,14 @@ pnpm install @farcaster/hub-web
...
 ```typescript
-import { getInsecureHubRpcClient } from '@farcaster/hub-web';
+import { getHubRpcClient } from '@farcaster/hub-web';
 
 (async () => {
   // if you are testing from a node environment
-  // const client = getInsecureHubRpcClient('testnet1.farcaster.xyz:2284', false);
+  // const client = getHubRpcClient('https://testnet1.farcaster.xyz:2284', false);
 
   // if you are testing from a braowser environment
```
pfletcherhill, line 31: Tiny nit: "browser" vs "braowser"

pfletcherhill (approved):
Looks good!

PR #833, opened 2023-04-06
fix: handle invalid version strings

sanjayprabhu:
## Motivation

Fixes hub crashing when version is empty

## Change Summary

Version can be empty for older hub versions.

## Merge Checklist

## Additional Context

### DIFF
--- a/apps/hubble/src/hubble.ts
+++ b/apps/hubble/src/hubble.ts
@@ -48,7 +48,7 @@ import {
   p2pMultiAddrStr,
 } from '~/utils/p2p';
 import { PeriodicTestDataJobScheduler, TestUser } from '~/utils/periodicTestDataJob';
-import { isBelowMinFarcasterVersion, VersionSchedule } from '~/utils/versions';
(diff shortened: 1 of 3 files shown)


apps/hubble/src/hubble.ts
```diff
@@ -545,7 +545,8 @@ export class Hub implements HubInterface {
 
       // Ignore peers that are below the minimum supported version.
       const theirVersion = message.hubVersion;
-      if (isBelowMinFarcasterVersion(theirVersion)) {
+      const versionCheckResult = isBelowMinFarcasterVersion(theirVersion);
+      if (versionCheckResult.isErr() || (versionCheckResult.isOk() && !versionCheckResult.value)) {
```
sanjayprabhu, line 549: This is a bit awkward, I wonder if should just rename it to ensureAboveMinVersion and just return an `err` if false/error and `ok` for true. But not sure if that's abusing `HubResult`.

adityapk00: Yeah, I think that would make sense. `ensureAboutMinVersion`, and have it return HubResult<bool>, and we disconnect ifErr or ifOk() && value === false

Issue #737, opened 2023-03-23
bug(hubble): VerificationEthAddress's claim network is not verified to match the message's network


varunsrin:
**What is the bug?**
Merging a VerificaitonAddEthereumAddress does not verify that the network of the Claim matches the network of the Message, which could lead to merging a claim that has a different network than the message. 

**How can it be reproduced? (optional)**

```
import {
  FarcasterNetwork,
  hexStringToBytes,
  makeVerificationAddEthAddress,
  makeVerificationEthAddressClaim,
} from '@farcaster/hub-nodejs';

// Safety: input is known and cannot throw, so safe to unwrap in the example
const addressBytes = (await eip712Signer.getSignerKey())._unsafeUnwrap();

const fid = 1;
const blockHashHex = '0x1d3b0456c920eb503450c7efdcf9b5cf1f5184bf04e5d8ecbcead188a0d02018';
const blockHashBytes = hexStringToBytes(blockHashHex)._unsafeUnwrap();

const claimResult = makeVerificationEthAddressClaim(fid, addressBytes, FarcasterNetwork.TESTNET, blockHashBytes);

if (claimResult.isOk()) {
  const claim = claimResult.value;

  // Sign the claim
  const ethSignResult = await eip712Signer.signVerificationEthAddressClaim(claim);
  const ethSignature = ethSignResult._unsafeUnwrap();

  // Construct a Verification Add Message with the claim signature
  const verificationBody = {
    address: addressBytes,
    ethSignature,
    blockHash: blockHashBytes,
  };

  const dataOptions = {
    fid: 1,
    network: FarcasterNetwork.TESTNET,
    timestamp: farcasterTimestamp,
  };

  const verificationMessage = await makeVerificationAddEthAddress(verificationBody, dataOptions, ed25519Signer);

// This message will be accepted. 
}
```

varunsrin:
this behavior actually works correctly (no bug) , but make sure we have a test case for this

PR #836, opened 2023-04-06
chore: Add test to ensure verification and claim must have same network

sanjayprabhu:
## Motivation

Closes https://github.com/farcasterxyz/hub-monorepo/issues/737

## Change Summary

Add a test to make sure it's not possible to pass in a different network in the claim vs the verification and have it pass.

## Merge Checklist

## Additional Context

### DIFF
--- a/apps/hubble/src/storage/engine/index.test.ts
+++ b/apps/hubble/src/storage/engine/index.test.ts
@@ -1,4 +1,5 @@
 import {
+  bytesToHexString,
   bytesToUtf8String,
   CastAddMessage,
   CastId,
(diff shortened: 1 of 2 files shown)


varunsrin (approved):

apps/hubble/src/storage/engine/index.test.ts
```diff
@@ -165,6 +166,36 @@ describe('mergeMessage', () => {
         ).resolves.toEqual(ok(verificationAdd));
         expect(mergedMessages).toEqual([signerAdd, verificationAdd]);
       });
+
+      test('fails when network does not match claim network', async () => {
+        const address = (await custodySigner.getSignerKey())._unsafeUnwrap();
```
pfletcherhill, line 171: Nit: you can use `custodySignerKey` which is already defined

apps/hubble/src/storage/engine/index.test.ts
```diff
@@ -165,6 +166,36 @@ describe('mergeMessage', () => {
...
+        const blockHash = Factories.BlockHash.build();
+        const mainnetClaim = Factories.VerificationEthAddressClaim.build(
+          {
+            fid: BigInt(fid),
+            network: FarcasterNetwork.MAINNET,
+            blockHash: bytesToHexString(blockHash)._unsafeUnwrap(),
+            address: bytesToHexString(address)._unsafeUnwrap(),
+          },
+          { transient: { signer: signer } }
```
pfletcherhill, line 180: You don't need to pass signer to this factory—the transient params aren't used for anything

apps/hubble/src/storage/engine/index.test.ts
```diff
@@ -165,6 +166,36 @@ describe('mergeMessage', () => {
         ).resolves.toEqual(ok(verificationAdd));
         expect(mergedMessages).toEqual([signerAdd, verificationAdd]);
       });
+
+      test('fails when network does not match claim network', async () => {
+        const address = (await custodySigner.getSignerKey())._unsafeUnwrap();
+        const blockHash = Factories.BlockHash.build();
+        const mainnetClaim = Factories.VerificationEthAddressClaim.build(
```
pfletcherhill, line 173: Might actually be easier to use the `makeVerificationEthAddressClaim` method, because it takes bytes as inputs so you don't have to call `bytesToHexString` on multiple arguments and can pass in `fid` as a number.

apps/hubble/src/storage/engine/index.test.ts
```diff
@@ -165,6 +166,36 @@ describe('mergeMessage', () => {
...
+              fid,
+              network: FarcasterNetwork.TESTNET,
+              verificationAddEthAddressBody: { address: address, blockHash: blockHash, ethSignature: claimSignature },
+            },
+          },
+          { transient: { signer: signer, ethSigner: custodySigner } }
+        );
+        const result = await engine.mergeMessage(testnetVerificationAdd);
+        // Signature will not match because we're attempting to recover the address based on the wrong network
+        expect(result._unsafeUnwrapErr()).toEqual(
```
pfletcherhill, line 195: It's better to use this pattern: `expect(result).toEqual(err(new HubError(...))`. The reason is in the case that `result` is an instance of Ok rather than Err, the jest error will show that there wasn't a match between the two. With the way it's written now, jest will saw we called `_unsafeUnwrapErr` on something that wasn't an error but won't tell us where exactly.

pfletcherhill (approved):
Thanks for adding this! A few minor nits to consider, but feel free to merge.

apps/hubble/src/storage/engine/index.test.ts
pfletcherhill: Seems like there's a place in factories where we pass transient params to `VerificationEthAddressClaimFactory.build(...)` as well, but that's an oversight and should be fixed too

PR #838, opened 2023-04-07
fix: Make dbreset its own command

adityapk00:
## Motivation

make dbreset its own command

## Change Summary

Remove --db-reset arg
Add dbreset command

## Merge Checklist

### DIFF
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -12,7 +12,7 @@ import { APP_VERSION, Hub, HubOptions } from '~/hubble';
 import { logger } from '~/utils/logger';
 import { addressInfoFromParts, ipMultiAddrStrFromAddressInfo, parseAddress } from '~/utils/p2p';
 import { DEFAULT_RPC_CONSOLE, startConsole } from './console/console';
-import { DB_DIRECTORY } from './storage/db/rocksdb';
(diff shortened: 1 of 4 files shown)


apps/hubble/src/cli.ts
```diff
@@ -71,7 +71,6 @@ app
   .option('--admin-server-enabled', 'Enable the admin server. (default: disabled)')
   .option('--admin-server-host <host>', "The host the admin server should listen on. (default: '127.0.0.1')")
   .option('--db-name <name>', 'The name of the RocksDB instance')
-  .option('--db-reset', 'Clears the database before starting')
```
sanjayprabhu, line 74: This would be a breaking change, is it safe to completely remove?

adityapk00: I think it should be fine. The prod servers are depending on the DB_RESET_TOKEN, which still works, and I don't think anyone is relying on the --db-reset. @varunsrin Can you confirm that this is OK and it doesn't need a "breaking change" minor version upgrade?

varunsrin: a patch version bump is ok for now, i don't think anyone uses this command extensively in production yet. but good callout @sanjayprabhu i think this will change rapidly in the next few weeks, and we should be more careful about changes like this.

adityapk00: Cool, merging!

PR #841, opened 2023-04-07
fix: Hangup on peers when removing from address book

adityapk00:
## Motivation
Add call to hangup() so we actually disconnect from peers when removing from our addressbook

## Change Summary

- Add hangup() when remove from address book 
- Test

## Merge Checklist

### DIFF
--- a/apps/hubble/src/network/p2p/gossipNode.test.ts
+++ b/apps/hubble/src/network/p2p/gossipNode.test.ts
@@ -14,6 +14,7 @@ import { jestRocksDB } from '~/storage/db/jestUtils';
 import Engine from '~/storage/engine';
 import { MockHub } from '~/test/mocks';
 import SyncEngine from '../sync/syncEngine';
+import { PeerId } from '@libp2p/interface-peer-id';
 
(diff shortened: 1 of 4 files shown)


apps/hubble/src/network/p2p/gossipNode.test.ts
```diff
@@ -92,6 +93,33 @@ describe('GossipNode', () => {
...
+      let other = await node1.addressBook?.get(node2.peerId as PeerId);
+      expect(other?.length).toEqual(1);
+
+      await node1.removePeerFromAddressBook(node2.peerId as PeerId);
+
+      // Make sure the connection is closed
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+
+      other = await node1.addressBook?.get(node2.peerId as PeerId);
+      expect(other).toEqual([]);
```
sanjayprabhu, line 116: Shouldn't we check that node2 doesn't have a connection with node1 to be sure that it actually disconnected?

sanjayprabhu (approved):

Issue #860, opened 2023-04-12
bug(hubble): nodes on different networks are peering with each other


sanjayprabhu:
**What is the bug?**
We're seeing error messages like 

```
{"timestamp":1681164350000,"hash":"0xffadb4598e5cdd567cc7462ad28f3175be6ca226","fid":8928,"type":1},"source":"gossip","errCode":"bad_request.validation_failure","msg":"submitMessage error: incorrect network: 2 (expected: 1)"}
```

which means mainnet hubs are seeing testnet and vice versa.

**How can it be reproduced? (optional)**
Include steps, code samples, replits, screenshots and anything else that would be helpful to reproduce the problem. 

**Additional context (optional)**

Issue #849, opened 2023-04-10
chore: improve logs when peer has mismatched version or network


pfletcherhill:
We print `Peer is running an invalid or outdated version, ignoring` logs whenever the peer is running the wrong version or network.

#### Implementation
* Differentiate between version and network mismatches
* Include peer, network/version, and expected network/version in log message for easier browsing

PR #861, opened 2023-04-12
feat: Include network id in gossip topics to keep networks isolated

sanjayprabhu:
## Motivation

Fixes https://github.com/farcasterxyz/hub-monorepo/issues/860 and https://github.com/farcasterxyz/hub-monorepo/issues/849

## Change Summary

 - Adds the network id to all gossip topics the hubs subscribes to ensure we're only talking with peers on the same network

## Merge Checklist

## Additional Context

Tested manually. Bootstrapped the a node to testnet node and verified it's not receiving any messages from current testnet/mainnet nodes but can receive messages from another instance with the topic change.

### DIFF
--- a/apps/hubble/src/hubble.ts
+++ b/apps/hubble/src/hubble.ts
@@ -563,14 +563,7 @@ export class Hub implements HubInterface {
         return;
       }
 
-      // Ignore peers that are below the minimum supported version.
-      const theirVersion = message.hubVersion;
-      const versionCheckResult = ensureAboveMinFarcasterVersion(theirVersion);
-      if (versionCheckResult.isErr() || message.network !== this.options.network) {
-        log.warn(
-          { peerId, theirVersion, theirNetwork: message.network },
-          'Peer is running an invalid or outdated version, ignoring'
-        );
+      if (!(await this.isValidPeer(peerId, message))) {
         await this.gossipNode.removePeerFromAddressBook(peerId);
         this.syncEngine.removeContactInfoForPeerId(peerId.toString());
         return;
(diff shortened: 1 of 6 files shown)


apps/hubble/src/hubble.ts
```diff
@@ -835,6 +827,29 @@ export class Hub implements HubInterface {
 
     return ResultAsync.fromPromise(this.rocksDB.commit(txn), (e) => e as HubError);
   }
+  async isValidPeer(ourPeerId: PeerId, message: ContactInfoContent) {
```
pfletcherhill, line 830: Nit: it's a bit weird to me to have a separate method for this code. Because it's only used once and logs errors, maybe just include it where it's called above? I also don't love that we send our peer ID as a parameter just for logging purposes.

apps/hubble/src/network/p2p/gossipNode.ts
```diff
@@ -57,6 +64,12 @@ interface NodeOptions {
 export class GossipNode extends TypedEmitter<NodeEvents> {
   private _node?: Libp2p;
   private _periodicPeerCheckJob?: PeriodicPeerCheckScheduler;
+  private _network: FarcasterNetwork;
+
+  constructor(network?: FarcasterNetwork) {
+    super();
+    this._network = network ?? FarcasterNetwork.NONE;
```
pfletcherhill, line 71: It doesn't seem like we handle the NONE case any differently from any other network. Is that intended?

pfletcherhill (approved):
Looks good! Added two comments/questions to consider but otherwise seems good to merge.

apps/hubble/src/network/p2p/gossipNode.ts
sanjayprabhu: Yeah, the gossip node doesn't really care. None will only be used in tests.

PR #878, opened 2023-04-14
chore: re-organize sync engine to use submitMessage rather than mergeMessage

pfletcherhill:
## Motivation

In trying to diagnose sync behavior, I became frustrated at not having a clearer separation between sync, gossip, and storage. This PR attempts to draw a clearer distinction and use the `Hub` class as a bridge between them all.

## Change Summary

* Refactor sync to always merge messages through the hub's `submitMessage` method
* Remove sync-related methods (`forEachMessage` and `getAllMessagesBySyncIds`) from the storage engine

## Merge Checklist

## Additional Context

### DIFF
--- a/apps/hubble/src/network/sync/syncEngine.ts
+++ b/apps/hubble/src/network/sync/syncEngine.ts
@@ -19,12 +19,11 @@ import { PeerId } from '@libp2p/interface-peer-id';
 import { err, ok, Result } from 'neverthrow';
 import { TypedEmitter } from 'tiny-typed-emitter';
 import { EthEventsProvider } from '~/eth/ethEventsProvider';
-import { Hub } from '~/hubble';
+import { Hub, HubInterface } from '~/hubble';
 import { MerkleTrie, NodeMetadata } from '~/network/sync/merkleTrie';
 import { SyncId, timestampToPaddedTimestampPrefix } from '~/network/sync/syncId';
 import { TrieSnapshot } from '~/network/sync/trieNode';
 import RocksDB from '~/storage/db/rocksdb';
-import Engine from '~/storage/engine';
 import { sleepWhile } from '~/utils/crypto';
 import { logger } from '~/utils/logger';
 
@@ -61,7 +60,7 @@ type SyncStatus = {
 class SyncEngine extends TypedEmitter<SyncEvents> {
   private readonly _trie: MerkleTrie;
 
-  private readonly engine: Engine;
+  private readonly _hub: HubInterface;
   private readonly _ethEventsProvider: EthEventsProvider | undefined;
 
   private _isSyncing = false;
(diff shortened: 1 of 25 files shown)


apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -447,21 +455,13 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
...
-      );
-    }
-
-    // If there was a failed merge, log the error and move on. We'll only log one error, since they're likely all the same.
-    const failedMerge = mergeResults.find((r) => r.isErr());
-    if (failedMerge) {
-      log.warn(
-        { error: failedMerge._unsafeUnwrapErr(), errorMessage: failedMerge._unsafeUnwrapErr().message },
-        'Failed to merge message'
+        `Merged ${successCount} messages during sync with ${mergeResults.length - successCount} failures`
```
sanjayprabhu, line 464: The tradeoff here is that it makes log based metrics harder to analyze (e.g. there's no single field to filter by to create a datadog dashboard). I think we should add a new field like `at: "SyncEngine.merged_messages"` for messages that we think we'll create metrics from. 

Doesn't have to change here since we don't have any dashboards for this.

apps/hubble/src/rpc/adminServer.ts
```diff
@@ -114,31 +115,30 @@ export default class AdminServer {
...
+          let deletedCount = 0;
+          const prefixes = [
+            RootPrefix.User,
+            RootPrefix.CastsByMention,
+            RootPrefix.CastsByParent,
+            RootPrefix.ReactionsByTarget,
+          ];
+          for (const prefix of prefixes) {
+            const prefixBuffer = Buffer.from([prefix]);
+            const iterator = this.db.iteratorByPrefix(prefixBuffer);
```
sanjayprabhu, line 127: I had to write something very similar to analyze the db size. Maybe worth generalizing this something like iterateByPrefixes([prefix])  => prefix, key, value.

sanjayprabhu (approved):

apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -412,13 +419,16 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
     // Merge messages sequentially, so we can handle missing users.
     this._syncMergeQ += messages.length;
```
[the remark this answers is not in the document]
pfletcherhill: If we can confirm that the lock on `commitTransaction` is sufficient memory usage protection, then I'd be in favor of that. Aditya added this to ensure that we're not merging messages faster than the sync engine can update.

apps/hubble/src/network/sync/syncEngine.ts
pfletcherhill: I agree and think we can address this using other fields (like `method` or `component`) rather than trying to standardize the message

Issue #892, opened 2023-04-19
bug: hub does not retry when eth provider returns errors


varunsrin:

sanjayprabhu:
Noticed this is still happening. Re-opening.

Issue #898, opened 2023-04-20
bug: Sync issues v1.2


sanjayprabhu:
**What is the bug?**
This is a tracking ticket for all the sync bugs we've noticed on v1.2. Splitting this into two sections, roughly categorized as "breaking" and "efficiency". Breaking bugs are anything that can cause hubs to never converge and always report as being out of sync or otherwise interfere with the normal operation of the hub. Efficiency bugs are the rest, where impact is mainly doing excessive syncs.

The net effect here is that the hubs is constantly syncing.

**Breaking**

1. Sync trie hashes don't match even after attempting to sync.
I'm seeing a lot of logs like this:
```
{"level":30,"time":1681935767288,"pid":56629,"hostname":"hubble","component":"SyncEngine","peerId":"12D3KooWRxgeFVLTAxeVgj4QHPzuT6HaV3Hj93LNZQeRi7gAwUVk","inSync":false,"isSyncing":false,"theirMessages":1089397,"ourMessages":1089391,"peerNetwork":1,"peerVersion":"2023.3.1","msg":"SyncStatus"}
{"level":30,"time":1681935767288,"pid":56629,"hostname":"hubble","component":"SyncEngine","peerId":"12D3KooWRxgeFVLTAxeVgj4QHPzuT6HaV3Hj93LNZQeRi7gAwUVk","msg":"Diffsync: Syncing with peer"}
{"level":30,"time":1681935767288,"pid":56629,"hostname":"hubble","component":"SyncEngine","msg":"Perform sync: Start"}
{"level":30,"time":1681935767289,"pid":56629,"hostname":"hubble","component":"SyncEngine","divergencePrefix":"007","prefix":"007247656","msg":"Divergence prefix"}
```

Where we're only off by a few messages but the divergence prefix is pretty high up and we never get back in sync. Still unclear what could be causing this (maybe missing custody events as reported by @pfletcherhill?)

2. Some hubs are getting stuck 100K+ messages behind
![image](https://user-images.githubusercontent.com/75855/233441623-57c2779f-df07-4149-b019-c2451424a243.png)
This could because of the missing eth events issue, or the hubs getting into some kind of sync loop or otherwise getting into a bad state.

**Efficiency** 

1. Truncated prefixes when comparing snapshots
e.g. logs
```
[...]

PR #899, opened 2023-04-21
fix: Fix off by one error when retrying id registry event on sync

sanjayprabhu:
## Motivation

Fixes https://github.com/farcasterxyz/hub-monorepo/issues/892

And helps with https://github.com/farcasterxyz/hub-monorepo/issues/898

## Change Summary

There's an off by one error when we retry fetching a id registry event which meant we weren't actually querying the events. This fixes the issue, and adds a flag to trigger a manual resync if desired

## Merge Checklist

- [x] All [commits have been signed](../CONTRIBUTING.md#22-signing-commits)

## Additional Context

### DIFF
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -78,6 +78,7 @@ app
   .option('--admin-server-host <host>', "The host the admin server should listen on. (default: '127.0.0.1')")
   .option('--db-name <name>', 'The name of the RocksDB instance')
   .option('--rebuild-sync-trie', 'Rebuilds the sync trie before starting')
+  .option('--resync-events', 'Resyncs events from the Farcaster contracts before starting')
   .option('--commit-lock-timeout <number>', 'Commit lock timeout in milliseconds (default: 500)', parseNumber)
   .option('--commit-lock-max-pending <number>', 'Commit lock max pending jobs (default: 1000)', parseNumber)
   .option('-i, --id <filepath>', 'Path to the PeerId file')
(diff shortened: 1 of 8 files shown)


apps/hubble/src/cli.ts
```diff
@@ -78,6 +78,7 @@ app
   .option('--admin-server-host <host>', "The host the admin server should listen on. (default: '127.0.0.1')")
   .option('--db-name <name>', 'The name of the RocksDB instance')
   .option('--rebuild-sync-trie', 'Rebuilds the sync trie before starting')
+  .option('--resync-events', 'Resyncs events from the Farcaster contracts before starting')
```
pfletcherhill, line 81: Can we call this `resync-eth-events` just to draw more separation from hub events?

apps/hubble/src/eth/ethEventsProvider.ts
```diff
@@ -146,9 +146,9 @@ export class EthEventsProvider {
     return this._lastBlockNumber;
   }
 
-  public async start() {
+  public async start(resyncEvents?: boolean) {
```
pfletcherhill, line 149: We already pass in other config options to the constructor or the static `build` method. Can we pass this option there instead?

pfletcherhill (commented):
Two small comments

apps/hubble/src/cli.ts
sanjayprabhu: Addressed in https://github.com/farcasterxyz/hub-monorepo/pull/916

PR #928, opened 2023-04-29
feat: reject prunable messages on merge

pfletcherhill:
## Motivation

We've noticed that hubs sometimes get into a prune loop (since messages are only pruned at the hour mark, it's possible for hubs to sync messages that were pruned on some hubs but not others thus being treated as out of sync).

In order to prevent this scenario from happening, we don't merge any message that would be immediately prunable (timestamp earlier than the earliest message or the prune cut off)

Fixes https://github.com/farcasterxyz/hub-monorepo/issues/936

TODO:
 - [x] Handle other sets
 - [x] Fix existing tests
 - [x] Additional tests everywhere

## Change Summary

- Keep track of earliestTsHash for every set (and update it correctly on add/remove)
- On merge, validate that the message is not prunable and if so return an error

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR adds the ability to reject prunable messages on merge. It also includes changes to several files related to pruning messages.

### Detailed summary
- Added ability to reject prunable messages on merge
- Changes to several files related to pruning messages

> The following files were skipped due to too many changes: `apps/hubble/src/storage/jobs/pruneMessagesJob.test.ts`, `apps/hubble/src/network/sync/syncEnginePerf.test.ts`, `apps/hubble/src/storage/stores/storeEventHandler.ts`, `apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts`, `apps/hubble/src/storage/stores/storeEventHandler.test.ts`, `apps/hubble/src/storage/stores/reactionStore.test.ts`, `apps/hubble/src/network/sync/syncEngine.test.ts`, `apps/hubble/src/storage/stores/castStore.test.ts`, `apps/hubble/src/storage/stores/storageCache.ts`, `apps/hubble/src/storage/stores/storageCache.test.ts`

### DIFF
--- a/apps/hubble/src/storage/stores/castStore.ts
+++ b/apps/hubble/src/storage/stores/castStore.ts
@@ -282,7 +282,12 @@ class CastStore {
       .acquire(
         message.data.fid.toString(),
         async () => {
-          const prunableResult = await this._isPrunable(message);
+          const prunableResult = await this._eventHandler.isPrunable(
+            message,
+            UserPostfix.CastMessage,
+            this._pruneSizeLimit,
+            this._pruneTimeLimit
+          );
           if (prunableResult.isErr()) {
             throw prunableResult.error;
           } else if (prunableResult.value) {
@@ -602,45 +607,6 @@ class CastStore {
 
     // Delete message
     return deleteMessageTransaction(txn, message);
-  }
-
-  private async _isPrunable(message: CastAddMessage | CastRemoveMessage): HubAsyncResult<boolean> {
-    const farcasterTime = getFarcasterTime();
-    if (farcasterTime.isErr()) {
-      return err(farcasterTime.error);
-    }
-
-    // Calculate the timestamp cut-off to prune
-    const timestampToPrune = farcasterTime.value - this._pruneTimeLimit;
-
-    if (message.data.timestamp < timestampToPrune) {
-      return ok(true);
-    }
-
-    const messageCount = await this._eventHandler.getCacheMessageCount(message.data.fid, UserPostfix.CastMessage);
-    if (messageCount.isErr()) {
-      return err(messageCount.error);
-    }
-    if (messageCount.value < this._pruneSizeLimit) {
-      return ok(false);
-    }
-
-    const earliestTimestamp = await this._eventHandler.getEarliestTsHash(message.data.fid, UserPostfix.CastMessage);
-    if (earliestTimestamp.isErr()) {
-      return err(earliestTimestamp.error);
-    }
-    const tsHash = makeTsHash(message.data.timestamp, message.hash);
-    if (tsHash.isErr()) {
-      return err(tsHash.error);
-    }
-    if (earliestTimestamp.value === undefined) {
-      return ok(false);
-    }
-    if (bytesCompare(tsHash.value, earliestTimestamp.value) < 0) {
-      return ok(true);
-    } else {
-      return ok(false);
-    }
   }
 }
 
--- a/apps/hubble/src/storage/stores/castStore.ts
+++ b/apps/hubble/src/storage/stores/castStore.ts
@@ -316,7 +316,7 @@ class CastStore {
   async pruneMessages(fid: number): HubAsyncResult<number[]> {
     const commits: number[] = [];
 
-    const cachedCount = this._eventHandler.getCacheMessageCount(fid, UserPostfix.CastMessage);
+    const cachedCount = await this._eventHandler.getCacheMessageCount(fid, UserPostfix.CastMessage);
 
     // Require storage cache to be synced to prune
     if (cachedCount.isErr()) {
@@ -345,7 +345,7 @@ class CastStore {
         return ok(undefined); // Nothing left to prune
       }
 
-      const count = this._eventHandler.getCacheMessageCount(fid, UserPostfix.CastMessage);
+      const count = await this._eventHandler.getCacheMessageCount(fid, UserPostfix.CastMessage);
       if (count.isErr()) {
         return err(count.error);
       }
@@ -400,6 +400,13 @@ class CastStore {
   /* -------------------------------------------------------------------------- */
 
   private async mergeAdd(message: CastAddMessage): Promise<number> {
+    const prunableResult = await this._isPrunable(message);
+    if (prunableResult.isErr()) {
+      throw prunableResult.error;
+    } else if (prunableResult.value) {
+      throw new HubError('bad_request.pruned', 'message pruned');
+    }
+
     // Start RocksDB transaction
     let txn = this._db.transaction();
 
@@ -442,6 +449,13 @@ class CastStore {
   }
 
   private async mergeRemove(message: CastRemoveMessage): Promise<number> {
+    const prunableResult = await this._isPrunable(message);
+    if (prunableResult.isErr()) {
+      throw prunableResult.error;
+    } else if (prunableResult.value) {
+      throw new HubError('bad_request.pruned', 'message pruned');
+    }
+
     // Define cast hash for lookups
     const removeTargetHash = message.data.castRemoveBody.targetHash;
 
@@ -595,6 +609,50 @@ class CastStore {
 
     // Delete message
     return deleteMessageTransaction(txn, message);
+  }
+
+  private async _isPrunable(message: CastAddMessage | CastRemoveMessage): HubAsyncResult<boolean> {
+    // todo: will any of these error checks actually fail?
+
+    const farcasterTime = getFarcasterTime();
+    if (farcasterTime.isErr()) {
+      return err(farcasterTime.error);
+    }
+
+    // Calculate the timestamp cut-off to prune
+    const timestampToPrune = farcasterTime.value - this._pruneTimeLimit;
+
+    if (message.data.timestamp < timestampToPrune) {
+      return ok(true);
+    }
+
+    const messageCount = await this._eventHandler.getCacheMessageCount(message.data.fid, UserPostfix.CastMessage);
+    if (messageCount.isErr()) {
+      return err(messageCount.error);
+    }
+    if (messageCount.value < this._pruneSizeLimit) {
+      return ok(false);
+    }
+
+    const earliestTimestamp = await this._eventHandler.getEarliestMessageTimestamp(
+      message.data.fid,
+      UserPostfix.CastMessage
+    );
+    if (earliestTimestamp.isErr()) {
+      return err(earliestTimestamp.error);
+    }
+    const tsHash = makeTsHash(message.data.timestamp, message.hash);
+    if (tsHash.isErr()) {
+      return err(tsHash.error);
+    }
+    if (earliestTimestamp.value === undefined) {
+      return ok(false);
+    }
+    if (bytesCompare(tsHash.value, earliestTimestamp.value) < 0) {
+      return ok(true);
+    } else {
+      return ok(false);
+    }
   }
 }
 
--- a/apps/hubble/src/storage/stores/castStore.ts
+++ b/apps/hubble/src/storage/stores/castStore.ts
@@ -282,6 +282,13 @@ class CastStore {
       .acquire(
         message.data.fid.toString(),
         async () => {
+          const prunableResult = await this._isPrunable(message);
+          if (prunableResult.isErr()) {
+            throw prunableResult.error;
+          } else if (prunableResult.value) {
+            throw new HubError('bad_request.prunable', 'message would be pruned');
+          }
+
           if (isCastAddMessage(message)) {
             return this.mergeAdd(message);
           } else if (isCastRemoveMessage(message)) {
@@ -400,13 +407,6 @@ class CastStore {
   /* -------------------------------------------------------------------------- */
 
   private async mergeAdd(message: CastAddMessage): Promise<number> {
-    const prunableResult = await this._isPrunable(message);
-    if (prunableResult.isErr()) {
-      throw prunableResult.error;
-    } else if (prunableResult.value) {
-      throw new HubError('bad_request.pruned', 'message pruned');
-    }
-
     // Start RocksDB transaction
     let txn = this._db.transaction();
 
@@ -449,13 +449,6 @@ class CastStore {
   }
 
   private async mergeRemove(message: CastRemoveMessage): Promise<number> {
-    const prunableResult = await this._isPrunable(message);
-    if (prunableResult.isErr()) {
-      throw prunableResult.error;
-    } else if (prunableResult.value) {
-      throw new HubError('bad_request.pruned', 'message pruned');
-    }
-
     // Define cast hash for lookups
     const removeTargetHash = message.data.castRemoveBody.targetHash;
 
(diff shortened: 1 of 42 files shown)


apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -596,6 +600,35 @@ class CastStore {
     // Delete message
     return deleteMessageTransaction(txn, message);
   }
+
+  private async _isPrunable(message: Message) {
```
pfletcherhill, line 604: Nit: restrict this to `message: CastAddMessage | CastRemoveMessage`

apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -596,6 +600,35 @@ class CastStore {
...
+    const messageCount = this._eventHandler.getCacheMessageCount(message.data.fid, UserPostfix.CastMessage);
+    // todo: will any of these error checks actually fail?
+    if (messageCount.isErr()) {
+      throw messageCount.error;
```
pfletcherhill, line 608: What if we update the `getCacheMessageCount` method to compute the value live if it's undefined like we're doing with the earliest tshash? I don't think we should throw in this case.

apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -596,6 +600,35 @@ class CastStore {
...
+    }
+    if (this._pruneSizeLimit == 0 || messageCount < this._pruneSizeLimit) {
```
pfletcherhill, line 610: If pruneSizeLimit is 0, I think that would mean the store doesn't accept any messages so it should always be prunable. We should probably validate pruneSizeLimit when we set it that it's above 0.

apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -596,6 +600,35 @@ class CastStore {
...
+      return false;
+    }
+
+    const farcasterTime = getFarcasterTime();
```
pfletcherhill, line 614: We should put the timestamp checks above the message count and earliest tshash checks. Right now I think we'll incorrectly say messages with a timestamp that is too old are not prunable if the size of the set is smaller than pruneSizeLimit.

apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -596,6 +600,35 @@ class CastStore {
...
+    }
+
+    // Calculate the timestamp cut-off to prune
+    const timestampToPrune = farcasterTime.value - this._pruneTimeLimit;
+
+    if (message.data.timestamp < timestampToPrune) {
+      return true;
+    }
+
+    const earliestTimestamp = this._eventHandler.getEarliestMessageTimestamp(message.data.fid, UserPostfix.CastMessage);
```
pfletcherhill, line 626: We want to check the ordering of all the cast messages for an fid, which depends on the tshash, not just the timestamp. If two messages have the same timestamp, we look at their hash to determine ordering.

apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -596,6 +600,35 @@ class CastStore {
...
+    if (earliestTimestamp.isErr()) {
+      throw earliestTimestamp.error;
+    }
+    return message.data.timestamp < earliestTimestamp.value;
```
pfletcherhill, line 630: Once we return the tshash, we'll do `bytesCompare` rather than <

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -20,11 +25,15 @@ const makeKey = (fid: number, set: UserMessagePostfix): string => {
...
   private _synced = false;
 
-  constructor(usage?: Map<string, number>) {
+  constructor(db: RocksDB, usage?: Map<string, number>) {
     this._usage = usage ?? new Map();
+    this._earliest = new Map();
+    this._db = db;
   }
 
   async syncFromDb(db: RocksDB): Promise<void> {
```
pfletcherhill, line 39: Since we're passing db into the constructor, we should remove it here

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -20,11 +25,15 @@ const makeKey = (fid: number, set: UserMessagePostfix): string => {
 const log = logger.child({ component: 'StorageCache' });
 
 export class StorageCache {
+  private _db: RocksDB;
   private _usage: Map<string, number>;
+  private _earliest: Map<string, number>;
```
pfletcherhill, line 30: I think these maps should be called `counts` and `earliestTsHashes` for extra clarity

pfletcherhill: Not sure why I called it usage in the first place...

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -20,11 +25,15 @@ const makeKey = (fid: number, set: UserMessagePostfix): string => {
...
   private _synced = false;
```
pfletcherhill, line 31: If we change the behavior for getting the message count to compute it on the fly like we do with earliest tshash, we can get rid of this synced flag. We should still call `syncFromDb` when we start the hub, but we shouldn't depend on being synced to give correct responses.

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -35,11 +44,15 @@ export class StorageCache {
...
     const iterator = db.iteratorByPrefix(prefix);
 
-    for await (const [key] of iterator) {
+    for await (const [key, value] of iterator) {
       const postfix = (key as Buffer).readUint8(1 + FID_BYTES);
       if (postfix < UserMessagePostfixMax) {
         const usageKey = (key as Buffer).subarray(1, 1 + FID_BYTES + 1).toString('hex');
         const count = usage.get(usageKey) ?? 0;
+        if (this._earliest.get(usageKey) === undefined) {
+          const message = Message.decode(new Uint8Array(value as Buffer));
```
pfletcherhill, line 53: We should store the tshash here which is retrievable from the rocksdb key. I think this should be it: `(key as Buffer).subarray(1 + FID_BYTES + 1)`

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -59,6 +72,21 @@ export class StorageCache {
     return ok(this._usage.get(key) ?? 0);
   }
 
+  async getEarliestMessageTimestamp(fid: number, set: UserMessagePostfix): HubResult<number> {
```
pfletcherhill, line 75: `getEarliestTsHash`

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -59,6 +72,21 @@ export class StorageCache {
...
+    // sync on initial load
+    const key = makeKey(fid, set);
+    if (this._earliest.get(key) === undefined) {
```
pfletcherhill, line 78: I think we should check whether we have any messages for that fid and store as well (using `getMessageCount` at the top here) so we don't keep creating iterators for stores we know won't have an earliest ts hash

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -59,6 +72,21 @@ export class StorageCache {
...
+      const pruneIterator = getMessagesPruneIterator(this._db, fid, set);
+      // iterate once or twice?
+      const nextMessage = await ResultAsync.fromPromise(getNextMessageFromIterator(pruneIterator), () => undefined);
```
pfletcherhill, line 81: We should just use `next` rather than this wrapper around next, because we only need to get the tshash from the next key rather than decoding the whole message and getting a value from it

apps/hubble/src/storage/stores/castStore.ts
sanjayprabhu: Done.

apps/hubble/src/storage/stores/castStore.ts
```diff
@@ -596,6 +610,50 @@ class CastStore {
     // Delete message
     return deleteMessageTransaction(txn, message);
   }
+
+  private async _isPrunable(message: CastAddMessage | CastRemoveMessage): HubAsyncResult<boolean> {
+    // todo: will any of these error checks actually fail?
```
sanjayprabhu, line 615: There's so much branching logic in this code, mainly for error handling. I'm pretty sure none of these will actually ever be hit because we've passed all the message validations by this point, but there's no clean way to avoid it.

apps/hubble/src/storage/stores/storageCache.ts
sanjayprabhu: Done, will remove the flag.

apps/hubble/src/storage/stores/storageCache.ts
sanjayprabhu: Done

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -20,43 +21,81 @@ const makeKey = (fid: number, set: UserMessagePostfix): string => {
...
 
-  getMessageCount(fid: number, set: UserMessagePostfix): HubResult<number> {
-    if (this._synced !== true) {
-      const error = new HubError('unavailable.storage_failure', 'storage cache is not synced with db');
-      log.warn({ errCode: error.errCode }, `getMessageCount error: ${error.message}`);
-      return err(error);
+  async getMessageCount(fid: number, set: UserMessagePostfix): HubAsyncResult<number> {
+    const key = makeKey(fid, set);
+    if (this._counts.get(key) === undefined) {
+      const iterator = getMessagesPruneIterator(this._db, fid, set);
```
sanjayprabhu, line 64: Should probably just rename this to `getMessagesForFidBySet` or something like that.

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -80,6 +108,7 @@ export class StorageCache {
 
   private addMessage(message: Message): void {
     if (message.data !== undefined) {
+      this.resetEarliestTimestamp(message);
```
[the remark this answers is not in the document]
sanjayprabhu: I was trying to reduce duplication, but you're right, this is clearer. Fixed.

apps/hubble/src/storage/stores/storageCache.ts
sanjayprabhu: Will rename

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -20,43 +21,81 @@ const makeKey = (fid: number, set: UserMessagePostfix): string => {
...
+      const pruneIterator = getMessagesPruneIterator(this._db, fid, set);
+      const [firstKey] = await pruneIterator.next();
+      await pruneIterator.end();
+      if (firstKey === undefined) {
+        // todo: this might be normal (empty sets)
+        return err(new HubError('unavailable.storage_failure', 'could not read earliest message from db'));
+      }
+
+      const tsHash = firstKey.subarray(1 + FID_BYTES + 1);
+      const lookupKey = firstKey.subarray(1, 1 + FID_BYTES + 1).toString('hex');
```
pfletcherhill, line 93: Any reason not to re-use `key` here?

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -20,43 +21,81 @@ const makeKey = (fid: number, set: UserMessagePostfix): string => {
...
+      const iterator = getMessagesPruneIterator(this._db, fid, set);
+      for await (const [,] of iterator) {
+        const count = this._counts.get(key) ?? 0;
+        this._counts.set(key, count + 1);
+      }
     }
+    return ok(this._counts.get(key) ?? 0);
+  }
+
+  async getEarliestMessageTimestamp(fid: number, set: UserMessagePostfix): HubAsyncResult<Uint8Array | undefined> {
```
pfletcherhill, line 73: change to `getEarliestTsHash`

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -83,8 +122,19 @@ export class StorageCache {
...
+      this._counts.set(key, count + 1);
+
+      const tsHashResult = makeTsHash(message.data.timestamp, message.hash);
+      if (tsHashResult.isOk()) {
+        const currentEarliest = this._earliestTsHashes.get(key);
+        // todo: check direction on the compare
+        if (currentEarliest === undefined || bytesCompare(currentEarliest, tsHashResult.value) > 0) {
+          this._earliestTsHashes.set(key, tsHashResult.value);
+        }
+      } else {
```
pfletcherhill, line 135: Maybe check for `tshashResult.isErr()` above so you can return the error above rather than wrapping everything following in a conditional

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -93,11 +143,22 @@ export class StorageCache {
...
         log.error(`error: ${set} store message count is already at 0 for fid ${fid}`);
       } else {
-        this._usage.set(key, count - 1);
+        this._counts.set(key, count - 1);
+      }
+
+      const tsHashResult = makeTsHash(message.data.timestamp, message.hash);
+      if (tsHashResult.isOk()) {
+        const currentEarliest = this._earliestTsHashes.get(key);
+        // todo: Should this be an exact match? Is there an ordering issue here when two messages have the same timestamp?
```
pfletcherhill, line 156: Two messages will not have the same tsHash

apps/hubble/src/storage/stores/storeEventHandler.ts
```diff
@@ -149,15 +149,19 @@ class StoreEventHandler extends TypedEmitter<StoreEvents> {
...
-    this._storageCache = new StorageCache();
+    this._storageCache = new StorageCache(this._db);
   }
 
-  getCacheMessageCount(fid: number, set: UserMessagePostfix): HubResult<number> {
+  async getCacheMessageCount(fid: number, set: UserMessagePostfix): HubAsyncResult<number> {
     return this._storageCache.getMessageCount(fid, set);
   }
 
+  async getEarliestMessageTimestamp(fid: number, set: UserMessagePostfix): HubAsyncResult<Uint8Array | undefined> {
```
pfletcherhill, line 159: `getEarliestTsHash`

pfletcherhill (commented):
Looking good! Added some more comments

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -20,43 +21,81 @@ const makeKey = (fid: number, set: UserMessagePostfix): string => {
...
+      }
     }
+    return ok(this._counts.get(key) ?? 0);
+  }
+
+  async getEarliestMessageTimestamp(fid: number, set: UserMessagePostfix): HubAsyncResult<Uint8Array | undefined> {
+    // sync on initial load
     const key = makeKey(fid, set);
-    return ok(this._usage.get(key) ?? 0);
+    const messageCount = await this.getMessageCount(fid, set);
```
pfletcherhill, line 76: I think it's probably okay (and I know I suggested it), but I'm a bit weirded out by the interdependency of these two methods. In the case where the cache is not synced, we're iterating through all messages to count them and then iterating over the first one to get its key. I know the idea here is to quickly catch the case where there aren't any messages so we know the first tshash will be undefined, but can you think of any other ways of doing it?

sanjayprabhu: It should only happen for the first message in a set for new fids (we sync the rest from the db on startup, and will set it to 0 if there are no messages). That also means there's likely not much data to scan.

The other approach is to have some kind of sentinel value here to denote there are no messages and therefore don't scan again. That would have the same performance as the above and remove the dependency.

PR #959, opened 2023-05-11
fix(eth): add stronger retry logic to eth events provider

### LINKED ISSUE #892: bug: hub does not retry when eth provider returns errors

CassOnMars:
## Motivation

The ethEventsProvider has some amount of retry logic baked in, but there are problems associated with getting events in on the contracts' subscribers – the RPC provider gets invoked internally by the contract proxies’s subscriber polling, which means any attempt at retry-wrapping inside the ethEventsProvider will only catch the cases where we’re directly invoking the provider. Additionally, the existing retry logic in ethEventsProvider is not accounting for caching considerations in how ethers handles identical requests in rapid succession.

Fixes #892 

## Change Summary

A mixin has been added, called RetryProvider, which wraps around either the JsonRpcProvider class or the AbstractProvider class and overrides relevant calls for `_perform`, and specifically `getLogs` and `getLatestBlock` for AbstractProvider (which seems to be the particular calls that are not fully wrapped inside the contract's subscriber).

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR adds stronger retry logic to the `ethEventsProvider` in the `Hubble` app. Here are the changes:

### Detailed summary
- Add `RetryProvider` class to handle retries for `JsonRpcProvider` calls.
- Add `MockFaultyRPCProvider` class to simulate a faulty RPC provider.
- Update `EthEventsProvider` to use `RetryProvider` and handle retries when calling `JsonRpcProvider` methods.

> The following files were skipped due to too many changes: `apps/hubble/src/eth/ethEventsProvider.ts`, `apps/hubble/src/eth/ethEventsProvider.test.ts`


apps/hubble/src/eth/ethEventsProvider.ts
```diff
@@ -292,33 +294,20 @@ export class EthEventsProvider {
...
 
       // Fetch our batch of blocks
-      let batchIdEvents = await ResultAsync.fromPromise(
+      const batchIdEvents = await this.executeCallAsPromiseWithRetry(
         // Safety: queryFilter will always return EventLog as long as the ABI is valid
-        this._idRegistryContract.queryFilter(typeString, Number(nextFromBlock), Number(nextToBlock)) as Promise<
-          EventLog[]
-        >,
+        () =>
+          this._idRegistryContract.queryFilter(typeString, Number(nextFromBlock), Number(nextToBlock)) as Promise<
```
sanjayprabhu, line 300: Maybe I'm missing something, but why do we need both `executeCallAsPromiseWithRetry` and the RetryProvider? Isn't the retry provider going to call `getLogsWithRetry` within `queryFilter`?

apps/hubble/src/eth/ethEventsProvider.ts
```diff
@@ -642,4 +618,44 @@ export class EthEventsProvider {
...
+          case 'OFFCHAIN_FAULT':
+            if (attempt !== 3) {
+              const logger = log.child({ event: { rpcError: err.code } });
+              logger.warn(`ethRPCError: RPC returned response ${err.message}, retrying (attempt ${attempt}).`);
+
+              // delay is required because ethers caches results from promises for 250ms
+              await new Promise<void>((resolve) =>
+                setTimeout(() => {
+                  resolve();
+                }, this._ethersPromiseCacheDelayMS + 1)
```
sanjayprabhu, line 648: We've noticed that we're failing because we're hitting rate limits. Not sure if a 251ms delay will actually fix that issue. May need a longer delay or exponential backoff. Doesn't have to happen in this PR, but just wanted to note that in case this doesn't fully solve the problem.

apps/hubble/src/eth/ethEventsProvider.ts
CassOnMars: They handle different types of failures – The retryprovider handles HTTP-code based failures, which can be external to JSON-RPC. The executeCallAsPromiseWithRetry call cares about the specific RPC error codes

apps/hubble/src/eth/ethEventsProvider.ts
CassOnMars: This isn't performing exponential backoff, just a delay to bypass ethers' hard-coded cache window. I could add exponential backoff if desired

CassOnMars: exponential backoff added

sanjayprabhu (approved):

Issue #933, opened 2023-05-01
chore: create a system to track network propagation delay


varunsrin:
**What is the feature you would like to implement?**

A system that tracks how quickly messages propagate across hubs. The ideal, but impractical, system would have a god's eye view that observes new casts entering any hub and then measures how long it takes for that cast to show up on every other hub. 

**Why is this feature important?**

Our goal is for the Hub network to be as real-time as possible. Being able to measure this in an objective way will make it easier to improve over time. 

**Will the protocol spec need to be updated??**
No

akshaan:
Here's the approach I'm thinking of taking for this, would love to get some feedback @sanjayprabhu @sagar-a16z :

New proto schemas:
```
syntax = "proto3";

enum NetworkLatencyMessageType {
  NETWORK_LATENCY_MESSAGE_NONE = 0;
  NETWORK_LATENCY_MESSAGE_PING = 1; // Ping sent from hub measuring network latency
  NETWORK_LATENCY_MESSAGE_ACK = 2; // Acknowledgement sent from hub receiving a ping message
}

message NetworkLatencyMessage {
    NetworkLatencyMessageType messageType = 1;
    oneof body {
        PingMessageBody ping_message = 2;
        AckMessageBody ack_message = 3;
    }
}

message PingMessageBody {
    string sender_peer_id = 1;
    uint32 ping_timestamp = 2;
}

message AckMessageBody {
    string sender_peer_id = 1;
    uint32 ping_timestamp = 2;
    uint32 ack_timestamp = 3;
}
```

Procedure: 
- When enabled (via CLI flag), a hub will send a `PingMessageBody` message with its peerId and the current timestamp to the gossip network on startup. The hub will also hold the current timestamp in memory.
- Hubs that receive the ping message will respond with an `AckMessageBody` with the sender's peerId and the current timestamp on the gossip network.
- When the sending hub receives an ACK message, it checks that the message peerId matches it's peerId and the message `ping_timestamp` field matches the value in memory. 
- When the sending hub's in-memory state has more than N records, it writes a log line with `ack_timestamp` and `ping_timestamp` for each of the records. This will be consumed by Datadog, where we can plot the mean and quantiles of network latency (i.e. `ack_timestamp - ping_timestamp`)
- After logging, the sending hub repeats the process by sending a new ping message to the network.

Questions:
[...]

PR #970, opened 2023-05-16
feat: create a system to track network propagation delay 

akshaan:
## Motivation

This change adds an default-off flag that enables sending diagnostic messages on the gossip network (primary topic) to estimate network latency metrics. The metrics captures are:
- Time taken for a message to reach a given peer
- Time taken for a message to cover 50/75/90/99% of the network

#933

## Change Summary

A ping message (`PingMessageBody`) is sent as part of `GossipMessage`s periodically (every 5 minutes by default). Hubs respond to this message with an acknowledgement (`AckMessageBody`). The acknowledgements are collected by the hub that issued the ping messages and logged with aggregated metrics.

## Merge Checklist

## Additional Context

---

### PR Focus
This PR focuses on adding network latency metrics to the gossip network and enabling gossip network tracing and metrics.

### Detailed summary
- Added `NetworkLatencyMessage` and related message bodies to the `GossipMessage` protobuf schema.
- Added `GossipMetrics` enum value to `UserNameProof` in `types.ts`.
- Added `GossipMetricsRecorder` class to record network latency metrics in `gossipMetricsRecorder.ts`.
- Updated `GossipNode` constructor to accept a `db` parameter and enable network latency messages based on `networkLatencyMessagesEnabled` parameter.
- Updated `GossipNode` start and stop methods to start and stop the `GossipMetricsRecorder`.
- Updated `GossipNode` publish method to record message receipt and merge time.
- Updated `GossipNode` recordMessageReceipt, recordMessageMerge, and recordLatencyAckMessageReceipt methods to call corresponding methods in `GossipMetricsRecorder`.
- Updated `cli.ts` to add `--gossip-metrics-enabled` option to enable gossip network tracing and metrics.
- Updated `docker-compose-multinode.yml` to add `--gossip-metrics-enabled` option to the CLI command for both hub instances.
- Updated `gossipNetwork.test.ts` to import `jestRocksDB` and pass it to the `GossipNode` constructor.
- Updated `gossipNode.test.ts` to import `jestRocksDB` and pass it to the `GossipNode` constructor.
- Updated `hubble.ts` to import `AckMessageBody` and `NetworkLatencyMessage` from `@farcaster/hub-nodejs` and `GOSSIP_PROTOCOL_VERSION` from `protocol.js`.
- Updated `hubble.ts` to add `gossipMetricsEnabled` option to the `Hub` constructor and pass it to the `GossipNode` constructor.

> The following files were skipped due to too many changes: `apps/hubble/src/hubble.ts`, `apps/hubble/src/network/p2p/gossipMetricsRecorder.test.ts`, `apps/hubble/src/network/p2p/gossipMetricsRecorder.ts`, `packages/core/src/protobufs/generated/gossip.ts`


apps/hubble/src/hubble.ts
```diff
@@ -535,6 +538,14 @@ export class Hub implements HubInterface {
         await this.handleContactInfo(peerIdResult.value, gossipMessage.contactInfoContent);
       }
       return ok(undefined);
+    } else if (gossipMessage.networkLatencyMessage) {
+      if (peerIdResult.isOk()) {
+        await this.gossipNode.handleNetworkLatencyMessage(
+          peerIdResult._unsafeUnwrap(),
```
sanjayprabhu, line 544: We can just call `.value` and I don't think we need the `isOk` check above as well, since we've already checked for `isErr` above

protobufs/schemas/gossip.proto
```diff
@@ -24,11 +24,30 @@ message ContactInfoContent {
   FarcasterNetwork network = 6;
 }
 
+message PingMessageBody {
+    bytes ping_origin_peer_id = 1;
+    uint32 ping_timestamp = 2;
+}
+
+message AckMessageBody {
+    bytes ping_origin_peer_id = 1;
```
sanjayprabhu, line 33: I think we should also include the acking peer id.

apps/hubble/src/network/p2p/gossipNode.ts
```diff
@@ -209,12 +230,63 @@ export class GossipNode extends TypedEmitter<NodeEvents> {
     return this.publish(gossipMessage);
   }
 
+  /** Serializes and sends a network latency ping message */
+  async gossipNetworkLatencyPing(): Promise<HubResult<PublishResult>[]> {
+    const pingMessage = PingMessageBody.create({
+      pingOriginPeerId: this.peerId?.toBytes() ?? new Uint8Array(),
```
sanjayprabhu, line 236: We should always have a peer id by this point. Better to fail if we don't than default to empty.

apps/hubble/src/network/p2p/gossipNode.ts
```diff
@@ -209,12 +230,63 @@ export class GossipNode extends TypedEmitter<NodeEvents> {
...
+        topics: [this.primaryTopic()],
+        peerId: this.peerId?.toBytes() ?? new Uint8Array(),
+        version: GOSSIP_PROTOCOL_VERSION,
+      });
+      return this.publish(gossipMessage);
+    } else if (message.ackMessage) {
+      log.info('handling gossip ack message');
+      const peerIdMatchesOrigin = this.peerId?.equals(message.ackMessage.pingOriginPeerId) ?? false;
+      if (peerIdMatchesOrigin) {
+        this._networkLatencyMetrics?.logMetrics(senderPeerId, message);
```
sanjayprabhu, line 277: I don't think senderPeerId will always be the same as the peer that generated the ack message, so we should be storing the peerId of who created the ack into the message and using that instead

apps/hubble/src/network/p2p/gossipNode.ts
```diff
@@ -236,7 +308,7 @@ export class GossipNode extends TypedEmitter<NodeEvents> {
         })
       );
 
-      log.debug({ identity: this.identity, results }, 'Published to gossip peers');
+      log.info({ identity: this.identity, results }, 'Published to gossip peers');
```
sanjayprabhu, line 311: Let's keep these to debug

apps/hubble/src/network/p2p/gossipNode.ts
```diff
@@ -209,12 +230,63 @@ export class GossipNode extends TypedEmitter<NodeEvents> {
...
+    return [err(new HubError('unavailable', { message: 'invalid message data in NetworkLatencyMessage' }))];
+  }
+
   /** Publishes a Gossip Message to the network */
   async publish(message: GossipMessage): Promise<HubResult<PublishResult>[]> {
     const topics = message.topics;
     const encodedMessage = GossipNode.encodeMessage(message);
 
-    log.debug({ identity: this.identity }, `Publishing message to topics: ${topics}`);
+    log.info({ identity: this.identity }, `Publishing message to topics: ${topics}`);
```
sanjayprabhu, line 289: debug

apps/hubble/src/network/p2p/gossipNode.ts
```diff
@@ -209,12 +230,63 @@ export class GossipNode extends TypedEmitter<NodeEvents> {
...
+    return this.publish(gossipMessage);
+  }
+
+  async handleNetworkLatencyMessage(
+    senderPeerId: PeerId,
+    message: NetworkLatencyMessage
+  ): Promise<HubResult<PublishResult | undefined>[]> {
+    if (message.pingMessage) {
+      // Respond to ping message with an ack message
+      log.info('handling gossip ping message');
```
sanjayprabhu, line 257: Should just be `debug`, could get noisy

apps/hubble/src/network/p2p/networkLatencyMetrics.ts
```diff
@@ -0,0 +1,91 @@
...
+  public logMetrics(senderPeerId: PeerId, message: NetworkLatencyMessage) {
+    if (message.ackMessage) {
+      const ackMessage = message.ackMessage;
+      // Log ack latency for peer
+      log.info(
+        {
+          receivingHubPeerId: senderPeerId.toString(),
+          latencyMilliseconds: ackMessage.ackTimestamp - ackMessage.pingTimestamp,
+        },
+        'gossip network latency metrics'
```
sanjayprabhu, line 43: prefer to keep these simple, like `LatencyMetrics` so it's easier to get an exact match on the message.

apps/hubble/src/network/p2p/networkLatencyMetrics.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  private logNetworkCoverage(senderPeerId: PeerId, ackMessage: AckMessageBody) {
+    // Add peerId to recent peerIds
+    this._recentPeerIds.set(senderPeerId.toString(), Date.now());
+
+    // Compute coverage metrics
+    const metricsKey = `${ackMessage.pingOriginPeerId}_${ackMessage.pingTimestamp}`;
+    const currentMetrics = this._metrics.get(metricsKey);
+    const newNumAcks = (this._metrics.get(metricsKey)?.numAcks ?? 0) + 1;
+    const coverageProportion = newNumAcks / this._recentPeerIds.size;
```
sanjayprabhu, line 73: Hmm, restarting hubs is going to mess with this. Since we deploy fairly frequently, this data could be noisy. We should come up with an approach that deals with that. One option could be to wait an hour or so (or x number of ping messages) before we start logging metrics.

apps/hubble/src/network/p2p/networkLatencyMetrics.ts
```diff
@@ -0,0 +1,91 @@
+import { PeerId } from '@libp2p/interface-peer-id';
+import { AckMessageBody, NetworkLatencyMessage } from '@farcaster/hub-nodejs';
+import { logger } from '~/utils/logger';
+
+const RECENT_PEER_TTL_MILLISECONDS = 5 * 60 * 3600 * 1000;
+const METRICS_TTL_MILLISECONDS = 5 * 60 * 3600 * 1000;
```
sanjayprabhu, line 6: Is this meant to be 5 hours? Is there an extra `* 60`? I would add a comment that explains the intended TTL in text

apps/hubble/src/network/p2p/periodicLatencyPing.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Result } from 'neverthrow';
+
+const log = logger.child({
+  component: 'PeriodicLatencyPing',
+});
+
+type SchedulerStatus = 'started' | 'stopped';
+
+// Every 5 minutes
+const DEFAULT_PERIODIC_LATENCY_PING_CRON = '*/5 * * * *';
```
sanjayprabhu, line 13: As network gets bigger having every hub send the ping on the exact same minute could get problematic. We should add a jitter.

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
```diff
@@ -0,0 +1,340 @@
...
+  lastAckTimestamp: number;
+};
+
+type PeerMessageMetrics = {
+  messageCount: number;
+};
+
+interface StringMap {
+  [key: string]: number;
+}
```
sanjayprabhu, line 37: Why not use the existing `Map<string, number>`?

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+  }
+
+  private getUpdatedCoverage(
+    peerId: PeerId,
+    time: number,
+    currentCoverage?: NetworkCoverageTimes
+  ): NetworkCoverageTimes {
+    const seenPeerIds: StringMap = Object.assign({}, currentCoverage?.seenPeerIds ?? {});
+    const coverageMap: StringMap = Object.assign({}, currentCoverage?.coverageMap ?? {});
```
sanjayprabhu, line 305: If we use the map, this will just be `new Map(currentCoverage?.seenPeerIds)`

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+  }
+
+  private computePeerLatencyAndCoverageMetrics(ackMessage: AckMessageBody, ackOriginPeerId: PeerId) {
+    // Compute peer-level latency metrics
+    const key = `${ackOriginPeerId.toString()}_${ackMessage.pingTimestamp}`;
+    const currentLatencyMetrics = this._metrics.peerLatencyMetrics[key.toString()];
+    const updatedLatencyMetrics: PeerLatencyMetrics = {
+      numAcks: (currentLatencyMetrics?.numAcks ?? 0) + 1,
+      lastAckTimestamp: currentLatencyMetrics?.lastAckTimestamp ?? ackMessage.ackTimestamp,
```
sanjayprabhu, line 173: Shouldn't we prefer `askMessage.ackTimestamp` over the existing `lastAckTimestamp`?

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  private getUpdatedCoverage(
+    peerId: PeerId,
+    time: number,
+    currentCoverage?: NetworkCoverageTimes
+  ): NetworkCoverageTimes {
+    const seenPeerIds: StringMap = Object.assign({}, currentCoverage?.seenPeerIds ?? {});
+    const coverageMap: StringMap = Object.assign({}, currentCoverage?.coverageMap ?? {});
+    if (seenPeerIds[peerId.toString()]) {
```
sanjayprabhu, line 306: If we do this check first, saves copying the objects.

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
```diff
@@ -0,0 +1,340 @@
...
+        coverageMap: coverageMap,
+      };
+    }
+  }
+
+  private async readMetricsFromDb(): Promise<GossipMetrics | undefined> {
+    const key = Buffer.from(METRICS_DB_KEY);
+    const value = await this._db?.get(key);
+    if (value) {
+      return GossipMetrics.fromBuffer(value);
```
sanjayprabhu, line 330: I wonder if we should call `expireMetrics` after we load from the db in case the hub was down for a long time or didn't successfully write to db when exiting.

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+    return undefined;
+  }
+
+  private async writeMetricsToDB() {
+    const key = Buffer.from(METRICS_DB_KEY);
```
sanjayprabhu, line 336: Create a new entry under `RootPrefix` and use that instead of a string key.

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
akshaan: I tried using `Map`s but it's not easy to JSON-deserialize maps from the database without coercing the deserialzed object to a `Map`. This makes it hard to read / write to RocksDB, so I decided to use an object instead

sanjayprabhu: Ah, I see. Makes sense. Definitely worth a comment here. Another option worth considering is to create a protobuf for this. I'm not sure it's worth it since the metrics are not really critical to the hub, but take a look at `HubState` and how it's used.

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
```diff
@@ -0,0 +1,340 @@
...
+  ) {
+    this.recentPeerIds = recentPeerIds ?? {};
+    this.peerLatencyMetrics = peerLatencyMetrics ?? {};
+    this.peerMessageMetrics = peerMessageMetrics ?? {};
+    this.globalMetrics = globalMetrics ?? { networkCoverage: {}, messageMergeTime: { sum: 0, numElements: 0 } };
+  }
+
+  static fromBuffer(buffer: Buffer): GossipMetrics {
+    const obj = JSON.parse(buffer.toString());
+    return new GossipMetrics(obj.recentPeerIds, obj.peerLatencyMetrics, obj.peerMessageMetrics, obj.globalMetrics);
```
sanjayprabhu, line 69: We should gracefully handle parsing errors in case the db entry is corrupted for some reason.

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
akshaan: We want to record the first ack from a given peer, so in the case that we receive multiple acks from the same peer, we'll retain the timestamp from the first one (i.e. `currentLatencyMetrics?.lastAckTimestamp`)

apps/hubble/src/network/p2p/gossipMetricsRecorder.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+const db = jestRocksDB('network.p2p.gossipMetricsRecorder.test');
+
+describe('Gossip metrics recorder accumulates metrics from messages', () => {
+  afterEach(async () => {
+    jest.resetAllMocks();
+    await db.del(Buffer.from([RootPrefix.GossipMetrics]));
+  });
+
+  test('', async () => {
```
sanjayprabhu, line 18: Needs a name

apps/hubble/src/network/p2p/gossipMetricsRecorder.ts
sanjayprabhu: Ah I see, should we call this firstAckTimestamp?

sanjayprabhu (approved):
Should be good to merge after fixing conflicts.

PR #973, opened 2023-05-17
feat: Links 

CassOnMars:
## Motivation

Strawman PR (will persist as draft PR) following https://github.com/farcasterxyz/protocol/discussions/85 FIP

## Change Summary

- Adds support for Links as per the current FIP

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR adds support for Links to the FarCaster Hub NodeJS. Notable changes include:

- Added support for Links
- Added `LinkRequest`, `LinksByFidRequest`, `LinksByTargetRequest` to `rpc.proto`
- Added `LinkAddData`, `LinkAddMessage`, `LinkRemoveData`, `LinkRemoveMessage` to `types.ts`
- Updated `UserPostfix` and `MessageType` enums to include `LinkMessage`, `MESSAGE_TYPE_LINK_ADD` and `MESSAGE_TYPE_LINK_REMOVE`
- Added `LinkBody` to `message.proto`
- Updated `isLinkAddData`, `isLinkAddMessage`, `isLinkRemoveData`, `isLinkRemoveMessage` in `typeguards.ts`
- Updated `PrunableMessage` and `UserMessagePostfix` enums in `types.ts`
- Updated `PruneMessagesJobScheduler` in `pruneMessagesJob.test.ts`
- Added `ensureAboveTargetFarcasterVersion` to `versions.ts` and updated `versions.test.ts`

> The following files were skipped due to too many changes: `packages/core/src/builders.ts`, `apps/hubble/src/rpc/server.ts`, `packages/core/src/factories.ts`, `packages/core/src/validations.ts`, `apps/hubble/src/storage/engine/index.ts`, `packages/hub-nodejs/src/generated/rpc.ts`, `packages/hub-nodejs/src/generated/message.ts`, `packages/core/src/protobufs/generated/message.ts`, `apps/hubble/src/rpc/test/linkService.test.ts`, `apps/hubble/src/storage/engine/index.test.ts`, `packages/hub-nodejs/src/generated/request_response.ts`, `packages/core/src/protobufs/generated/request_response.ts`, `apps/hubble/src/storage/stores/linkStore.ts`, `apps/hubble/src/storage/stores/linkStore.test.ts`


sds (approved):
Seems straightforward to me. I wish GitHub could auto-detect files that are auto-generated (to make it more obvious which files are worth reviewing by hand), but it looks like we could manually specify them with `.gitattributes`?

https://thoughtbot.com/blog/github-diff-supression

apps/hubble/src/storage/stores/linkStore.ts
```diff
@@ -0,0 +1,713 @@
...
+  MessagesPage,
+  PAGE_SIZE_MAX,
+  PageOptions,
+  StorePruneOptions,
+} from '~/storage/stores/types';
+import { logger } from '~/utils/logger';
+
+const PRUNE_SIZE_LIMIT_DEFAULT = 2_500;
+
+const makeTargetKey = (target: number | string): Buffer => {
```
varunsrin, line 45: Links currently only have an fid target, so this should always be a number.

apps/hubble/src/storage/stores/linkStore.ts
```diff
@@ -0,0 +1,713 @@
...
+/**
+ * Generates a unique key used to store a LinkAdd message key in the LinksAdd Set index
+ *
+ * @param fid farcaster id of the user who created the link
+ * @param type type of link created
+ * @param targetId id of the object being reacted to
+ *
+ * @returns RocksDB key of the form <RootPrefix>:<fid>:<UserPostfix>:<targetKey?>:<type?>
+ */
+const makeLinkAddsKey = (fid: number, type?: string, target?: number | string): Buffer => {
```
varunsrin, line 58: target should always be a number, which lets us simplify this interface

apps/hubble/src/storage/stores/linkStore.ts
```diff
@@ -0,0 +1,713 @@
...
+/**
+ * Generates a unique key used to store a LinkRemove message key in the LinksRemove Set index
+ *
+ * @param fid farcaster id of the user who created the link
+ * @param type type of link created
+ * @param targetId id of the object being reacted to
+ *
+ * @returns RocksDB key of the form <RootPrefix>:<fid>:<UserPostfix>:<targetKey?>:<type?>
+ */
+const makeLinkRemovesKey = (fid: number, type?: string, target?: number | string): Buffer => {
```
varunsrin, line 84: same comment re: target

apps/hubble/src/storage/stores/linkStore.ts
```diff
@@ -0,0 +1,713 @@
...
+    const linkRemovesKey = makeLinkRemovesKey(fid, type, target);
+    const linkMessageKey = await this._db.get(linkRemovesKey);
+
+    return getMessage(this._db, fid, UserPostfix.LinkMessage, linkMessageKey);
+  }
+
+  /** Finds all LinkAdd Messages by iterating through the prefixes */
+  async getLinkAddsByFid(
+    fid: number,
+    type?: string,
```
varunsrin, line 200: suggestion: is it worth sanitizing the string here and throwing an error in case an upstream caller provides an invalid value like a string that is empty or too long?

this applies to all other `type?` parameters in functions below

protobufs/schemas/message.proto
```diff
@@ -155,3 +158,11 @@ message VerificationRemoveBody {
   bytes address = 1; // Address of the Verification to remove
 }
 
+/** Adds or removes a Link */
+message LinkBody {
+  string type = 1; // Type of link, <= 8 characters
+  optional uint32 displayTimestamp = 2; // User-defined timestamp to preserve original time
```
varunsrin, line 164: nit: `User-defined timestamp that preserves original timestamp when message.data.timestamp needs to be updated for compaction`

packages/core/src/validations.ts
```diff
@@ -404,6 +409,15 @@ export const validateCastRemoveBody = (body: protobufs.CastRemoveBody): HubResul
   return validateMessageHash(body.targetHash).map(() => body);
 };
 
+export const validateLinkType = (type: string): HubResult<string> => {
+  const typeBuffer = Buffer.from(type);
```
varunsrin, line 413: this check is *slightly* different from what we do in the store, where we check `type.length === 0 || typeBuffer.length > 8`

assuming the types are valid (i.e. a string is passed in) these should be equivalent i think. but if some other kind of object makes it down here past the type checks, then they might deviate in behavior.

varunsrin (commented):
Overall looks good, the one point of feedback is that some methods assume that target is either a number or a string, while the protobuf only permits a number. 

This leads to some excess logic that can be eliminated, and potentially erroneous behavior if some part of the hub passed in a string in an unexpected place.

apps/hubble/src/storage/db/types.ts
```diff
@@ -27,8 +27,8 @@ export enum RootPrefix {
   CastsByParent = 2,
   /* Used to index casts by mention */
   CastsByMention = 3,
-  /* Used to index amps by fid */
-  AmpsByUser = 4,
```
sanjayprabhu, line 31: Just to confirm, this was never actually used on mainnet right?

apps/hubble/src/storage/stores/linkStore.ts
```diff
@@ -0,0 +1,713 @@
...
+      throw new HubError('bad_request.validation_failure', 'target is missing');
+    }
+
+    const type = message.data.linkBody.type;
+    if (!type) {
+      throw new HubError('bad_request.validation_failure', 'type is missing');
+    } else {
+      const typeBuffer = Buffer.from(type);
+      if (typeBuffer.length === 0 || typeBuffer.length > 8) {
+        throw new HubError('bad_request.validation_failure', 'type must be 1-8 bytes');
```
sanjayprabhu, line 634: Why are we converting to buffer? Is it to avoid the null check?

apps/hubble/src/storage/engine/index.ts
```diff
@@ -177,6 +182,9 @@ class Engine {
     const setPostfix = typeToSetPostfix(message.data!.type);
 
     switch (setPostfix) {
+      case UserPostfix.LinkMessage: {
+        return ResultAsync.fromPromise(this._linkStore.merge(message), (e) => e as HubError);
```
sanjayprabhu, line 186: Hmm, this is going to break sync backwards compatibility as mentioned in the `Release` section of the FIP. We should check for the hubble protocol version and ignore the links until the next one. Will need to make an exception for tests. Let me think if there's a cleaner way to do this in the future.

sanjayprabhu (commented):
We should also add a changeset with a minor version bump.

apps/hubble/src/storage/db/types.ts
varunsrin: correct, we never shipped amps

apps/hubble/src/storage/stores/linkStore.ts
CassOnMars: to get exact byte count

apps/hubble/src/storage/engine/index.ts
```diff
@@ -177,6 +183,14 @@ class Engine {
     const setPostfix = typeToSetPostfix(message.data!.type);
 
     switch (setPostfix) {
+      case UserPostfix.LinkMessage: {
+        const versionCheck = ensureAboveTargetFarcasterVersion('2023.4.19');
+        if (versionCheck.isErr()) {
```
sanjayprabhu, line 188: Nice, thanks. I don't think we need this check when reading them. The write is more important. But it doesn't hurt.

sanjayprabhu (approved):
LGTM

PR #997, opened 2023-05-26
chore(hubble): use viem to interact with ethereum

deodad:
## Motivation

See [why viem](https://viem.sh/docs/introduction.html).

## Change Summary

- Use viem instead of ethers to interact with ethereum.
- Configured much more conservative polling intervals to avoid hitting rate limits with RPC providers
  - Contract events are polled for every 36s, since events are only processed with 6+ confirmed blocks this should be sufficient to ensure we always have the events that can be processed 
  - Blocks are polled for every 6s

## Merge Checklist

## Additional Context

The existing test coverage of EthEventsProvider relied heavily on mocking ethers. Since this no longer applied, we needed a new approach. Rather than attempting to mock the viem library, which doesn't lend itself to easily be mocked, I opted to setup anvil so that we can test against a local ethereum node.

---

## PR-Codex overview
This PR updates various dependencies and introduces Viem for interacting with Ethereum. Notable changes include:

- Updating viem to version 1.1.4
- Adding ViemLocalEip712Signer to exports
- Adding Anvil proxy for parallelizing test suite
- Adding WatchBlockNumber and WatchContractEvent classes
- Removing RetryProvider and JsonRpcProvider
- Using Viem for interacting with Ethereum in various files

> The following files were skipped due to too many changes: `apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts`, `apps/hubble/src/test/mocks.ts`, `apps/hubble/src/test/utils.ts`, `yarn.lock`, `apps/hubble/src/eth/ethEventsProvider.test.ts`, `apps/hubble/src/eth/ethEventsProvider.ts`, `apps/hubble/src/eth/abis.ts`


manan19:
Can we remove `ethers` as part of this change?

Issue #1003, opened 2023-06-01
bug(hubble): Server can hang when subscribing client is slow


sanjayprabhu:
**What is the bug?**
When there is a slow subscriber, it's possible to get into an infinite loop if writing to stream fails while we're draining:

https://github.com/farcasterxyz/hub-monorepo/blob/13415ead7160c1f1dfc8e1f410e41d4db3a0b3a6/apps/hubble/src/rpc/bufferedStreamWriter.ts#L65

this causes the server to become unresponsive.

**How can it be reproduced? (optional)**
Include steps, code samples, replits, screenshots and anything else that would be helpful to reproduce the problem. 

**Additional context (optional)**

PR #1004, opened 2023-06-01
fix: server no longer hangs due to slow subscribers

sanjayprabhu:
## Motivation

Fixes https://github.com/farcasterxyz/hub-monorepo/issues/1003

## Change Summary

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR fixes a server hang issue caused by slow subscribers and adds a new constant to the `BufferedStreamWriter` class.

### Detailed summary
- Fixes server hanging due to slow subscribers in certain conditions.
- Adds `SLOW_CLIENT_GRACE_PERIOD_MS` constant to `BufferedStreamWriter` class.
- Changes `bufferedStreamWriter` import in `server.ts` to include `STREAM_MESSAGE_BUFFER_SIZE` and `SLOW_CLIENT_GRACE_PERIOD_MS` constants.
- Adds new `StreamMessage` type in `bufferedStreamWriter.test.ts`.
- Adds new `reset` method to `MockStream` class in `bufferedStreamWriter.test.ts`.
- Adds new test case to `bufferedStreamWriter.test.ts` to handle write failing when draining.


apps/hubble/src/rpc/bufferedStreamWriter.ts
```diff
@@ -62,7 +62,11 @@ export class BufferedStreamWriter {
     while (this.dataWaitingForDrain.length > 0) {
       const message = this.dataWaitingForDrain.shift();
 
-      this.writeToStream(message);
+      const res = this.writeToStream(message);
+      // If the write fails, break out of the loop, otherwise we'll end up in an infinite loop
+      if (res.isErr() || res.value === false) {
```
deodad, line 67: in the `res.value === false` case, will the result be that `message` gets re-ordered since it was pulled from the from the front of `dataWaitingForDrain` and then re-pushed to the end of `dataWaitingForDrain` by `writeToStream`?

deodad: I think there might be another race condition that could have messages written out of order when `this.streamIsBackedUp = false;` after a drain and then an external caller calls `writeToStream` before `sendWaitingData` is able to empty our buffer:
- writeToStream is called externally with message M1
- message M1 is buffered
- stream is drained
- this.streamIsBackedUp = false;
- writeToStream is called externally with message M2
- message M2 is written to the stream
- this.sendWaitingData(); executes
- message M1 is written to the stream

apps/hubble/src/rpc/bufferedStreamWriter.ts
```diff
@@ -47,22 +57,26 @@ export class BufferedStreamWriter {
       // We'll wait only for 10 seconds before destroying the stream
       const timeout = setTimeout(() => this.destroyStream(), STREAM_DRAIN_TIMEOUT_MS);
 
-      this.stream.once('drain', () => {
+      this.stream.once('drain', async () => {
         this.streamIsBackedUp = false;
```
deodad, line 61: I think there may be a race condition where events get written to the stream out-of-order after a drain happens and `this.streamIsBackedUp = false;`. An external call to `writeToStream` can happen before existing data is drained via `sendWaitingData`

apps/hubble/src/rpc/bufferedStreamWriter.ts
sanjayprabhu: Yup, good catch. Fixed now and added a test to make sure message ordering is preserved.

apps/hubble/src/rpc/bufferedStreamWriter.ts
sanjayprabhu: I converted this class back to fully synchronous and fixed the above bug, I don't think this is possible anymore now.

Issue #883, opened 2023-04-18
bug(hubble): db size grows too fast during sync 


sanjayprabhu:
**What is the bug?**
Sometimes when syncing a mainnet node from scratch the DB size grows to > 5gb, when it should only <1gb. This fixes itself on restart, but we've seen hubs crash because they took up all available disk space.

**How can it be reproduced? (optional)**
Start a fresh mainnet node and observe size of the db during sync.

**Additional context (optional)**
Suspect this is due to rocksdb compaction not kicking in.

sanjayprabhu:
Still see this happening. Reopening

varunsrin:
closing, since this should be resolved as of 1.3.3

PR #1011, opened 2023-06-08
bug(hubble): db size grows too fast during sync

akshaan:
## Motivation

#883 

## Change Summary

- Updates RocksDB iterator usage so that iterators are always closed, regardless of whether they reach their end during traversal or not
- Adds a system to track RocksDB iterators and log warnings if they are left open for too long

In many cases, we traverse iterators without the `for await` pattern, so they are not closed automatically. In these cases, we close the iterators only when they do not reach their end (i.e. `.next()` did not return an error). This leaves the iterators open in the case where they did reach their end (see [here](https://github.com/Level/abstract-leveldown/blob/2f836a89616657e50428254abcb90516a70c2f4e/README.md?plain=1#LL354C1-L354C1)). Open iterators pin the RocksDB LSM tree at the time of creation, preventing deletions + compactions from clearing pinned SST files. These SST files then collect causing hubs to overrun disk capacity.

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR adds a check for open iterators in RocksDB and clears them after a set time limit. It also adds a new constant for the maximum time an iterator can be open.

### Detailed summary
- Added a check for open iterators in RocksDB
- Clears open iterators after a set time limit
- Added a new constant for the maximum time an iterator can be open

### DIFF
--- a/apps/hubble/src/storage/db/rocksdb.ts
+++ b/apps/hubble/src/storage/db/rocksdb.ts
@@ -89,7 +89,11 @@ class RocksDB {
   /** This set and cron are used to check whether iterators are open
    * (i.e. iterator.end) has not been called for MAX_DB_ITERATOR_OPEN_MILLISECONDS
    */
-  private _openIterators: Set<{ iterator: Iterator; openTimestamp: number }>;
+  private _openIterators: Set<{
+    iterator: Iterator;
+    openTimestamp: number;
+    options: AbstractRocksDB.IteratorOptions | undefined;
+  }>;
   private _iteratorCheckTimer?: NodeJS.Timer;
 
   constructor(name?: string) {
@@ -103,7 +107,8 @@ class RocksDB {
         } else {
           if (now - entry.openTimestamp >= MAX_DB_ITERATOR_OPEN_MILLISECONDS) {
             log.warn(
-              `RocksDB iterator is open (.end was not called) for more than ${MAX_DB_ITERATOR_OPEN_MILLISECONDS} ms`
+              `RocksDB iterator with options ${entry.options} open (.end was not called)
+                for more than ${MAX_DB_ITERATOR_OPEN_MILLISECONDS} ms`
             );
           }
           return [entry];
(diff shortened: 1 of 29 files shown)


sanjayprabhu (commented):
Great catch. We should also have some way to prevent this from happening again or detect it at least.

We have an iterator class. Maybe we can add some logging there to detect iterators that have been open for too long.

apps/hubble/src/storage/db/rocksdb.ts
```diff
@@ -71,8 +86,31 @@ class RocksDB {
...
+    this._openIterators = new Set();
+    this._iteratorCheckTimer = setInterval(() => {
+      const now = Date.now();
+      const openIterators = [...this._openIterators].flatMap((entry) => {
+        if (!entry.iterator.isOpen) {
+          return [];
+        } else {
+          if (now - entry.openTimestamp >= MAX_DB_ITERATOR_OPEN_MILLISECONDS) {
+            log.warn(
+              `RocksDB iterator is open (.end was not called) for more than ${MAX_DB_ITERATOR_OPEN_MILLISECONDS} ms`
```
sanjayprabhu, line 106: Maybe worth logging the options so we can try to figure out which iterators are being left open.

apps/hubble/src/storage/db/rocksdb.ts
```diff
@@ -71,8 +86,36 @@ class RocksDB {
...
+    this._iteratorCheckTimer = setInterval(() => {
+      const now = Date.now();
+      const openIterators = [...this._openIterators].flatMap((entry) => {
+        if (!entry.iterator.isOpen) {
+          return [];
+        } else {
+          if (now - entry.openTimestamp >= MAX_DB_ITERATOR_OPEN_MILLISECONDS) {
+            log.warn(
+              `RocksDB iterator with options ${entry.options} open
+                for more than ${MAX_DB_ITERATOR_OPEN_MILLISECONDS} ms`
```
sanjayprabhu, line 111: What does this line look like? Just curious if it has enough information for us to track down where the iterator is coming from.

sanjayprabhu (approved):
Nice catch. Not sure if this will completely fix the issue, there might still be other weird cases but it should definitely help a lot.

apps/hubble/src/storage/db/rocksdb.ts
akshaan: This line will have these options in it:
```
  gt?: K | undefined;
  gte?: K | undefined;
  lt?: K | undefined;
  lte?: K | undefined;
  reverse?: boolean | undefined;
  limit?: number | undefined;
  keys?: boolean | undefined;
  values?: boolean | undefined;
  keyAsBuffer?: boolean | undefined;
  valueAsBuffer?: boolean | undefined;
```

I don't think this is enough to track down iterators exactly, since most uses of the time we use iterators with default options. Not sure how best to pin point specific iterators without adding some kind of name or id to each instance

PR #1107, opened 2023-07-12
fix: Improve open iterator logging

adityapk00:
## Motivation

Improve logging for iterators so we know exactly where the iterators are open

## Change Summary

- Fix issue where hubevents that too long didn't close the iterator
- Improve iterator logging with id, stacktrace
- Don't overlap multiple DB compactions

## Merge Checklist

## Additional Context

---

### Focus of this PR:
Improve logging for open iterators and add a timeout for hub events stream.

### Detailed summary:
- Added a new private property `_isCompacting` to track if the database is currently being compacted.
- Modified the `compactDbIfRequired` method to check if the database is already being compacted before initiating a compaction.
- Added logging statements for starting and completing database compaction.
- Added a new private property `_openIteratorId` to assign unique IDs to open iterators.
- Modified the `iterator` method to store the stack trace and ID of each open iterator in the `_openIterators` set.
- Added a new timeout for the hub events stream to prevent situations where the client is not reading the stream.
- Added logging statements for stream timeout and stream memory usage.

### DIFF
--- a/apps/hubble/src/network/sync/syncEngine.ts
+++ b/apps/hubble/src/network/sync/syncEngine.ts
@@ -104,6 +104,7 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
 
   // Number of messages since last compaction
   private _messagesSinceLastCompaction = 0;
+  private _isCompacting = false;
 
   constructor(hub: HubInterface, rocksDb: RocksDB, ethEventsProvider?: EthEventsProvider, profileSync = false) {
     super();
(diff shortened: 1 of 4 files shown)


apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -722,11 +723,15 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
 
   public async compactDbIfRequired(messagesLength: number): Promise<boolean> {
     this._messagesSinceLastCompaction += messagesLength;
-    if (this.shouldCompactDb) {
+    if (this.shouldCompactDb && !this._isCompacting) {
+      this._isCompacting = true;
```
sanjayprabhu, line 727: I think we can actually remove this compacting logic (or make it very infrequent). This was added before we found out about the iterators and I think that's the real issue. Anyway, doesn't have to happen in this PR, but just FYI.

apps/hubble/src/rpc/server.ts
```diff
@@ -952,6 +952,22 @@ export default class Server {
             return;
           }
 
+          // We'll set a timeout of 1 hour, after which we'll stop writing to the stream
+          // This is to prevent a situation where we're writing to the stream, but the client
+          // is not reading it.
+          const timeout = setTimeout(async () => {
```
sanjayprabhu, line 958: I think we already have a more aggressive check in BufferedStreamWriter? Wouldn't that close the stream if no one read from it within ~60 seconds

adityapk00: There's something strange with gRPC timeouts and streams. We had to create the BufferedStreamWriter because the gRPC timeouts / buffer management wasn't handling it properly. 

I want to run this in prod for a while to see which timeouts are being missed. And we'll be able to co-relate it with the iterator stack trace to see what exactly is not working. 

I'd like to keep this timeout in the PR as well (and if this is not the problem, I'll remove it)

PR #1113, opened 2023-07-12
feat: Add hash to links table in postgres replicator

zachterrell57:
## Motivation

Right now every table has a `hash` column, so this is just making the `links` table consistent

## Change Summary

Added property, added migration

## Merge Checklist

### DIFF
--- a/packages/hub-nodejs/examples/replicate-data-postgres/migrations/002_add_hash_to_links.ts
+++ b/packages/hub-nodejs/examples/replicate-data-postgres/migrations/002_add_hash_to_links.ts
@@ -1,0 +1,12 @@
+import { Kysely, sql } from 'kysely';
+
+export const up = async (db: Kysely<any>) => {
+  await db.schema
+    .alterTable('links')
+    .addColumn('hash', sql`bytea`, (col) => col.notNull())
+    .execute();
+};
+
+export const down = async (db: Kysely<any>) => {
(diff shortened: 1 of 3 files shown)


packages/hub-nodejs/examples/replicate-data-postgres/migrations/002_add_hash_to_links.ts
```diff
@@ -0,0 +1,13 @@
+import { Kysely, sql } from "kysely";
+import { Database } from "../db";
+
+export const up = async (db: Kysely<Database>) => {
+  await db.schema
+    .alterTable("links")
+    .addColumn("hash", sql`bytea`, (col) => col.notNull())
```
sds, line 7: This won't work for users with existing DBs, but it's quick enough to start over that it seems fine. Would be work to backfill.

sds (approved):
Thanks!

packages/hub-nodejs/examples/replicate-data-postgres/migrations/002_add_hash_to_links.ts
manan19: @sds Running an app on top of this db now, are you suggesting starting over for links table only or for the entire db?

sds: Yes, sorry for the hassle. When we run this locally it takes about an hour to start from scratch.

Issue #742, opened 2023-03-23
feat: support fallback RPC providers


deodad:
**What is the feature you would like to implement?**
Allow for multiple backend services to specified and connected to for retrieving Ethereum events.

**Why is this feature important?**
This ensures that if a backend has become out-of-sync, or if it has been compromised that its responses are dropped in favor of responses that match the majority.

**Will the protocol spec need to be updated??**
No

**How should this feature be built? (optional)**
Viem has a [fallback transport](https://viem.sh/docs/clients/transports/fallback.html) that allows multiple RPC urls to be specified and will automatically rank them based on stability and latency. If a request fails on one transport (say due to rate limiting), additional transports will be tried.

Hub runners would need to be able to specify multiple RPC urls when running a hub.

sanatshah:
Looking at picking this up, does it still need to be implemented?

deodad:
yes

maurerbot:
where in the code is are you initializing viem providers?

deodad:
@sanatshah looks like @stephancill beat you to it

therealharpaljadeja:
is this still an issue to resolved? @deodad now that you mention @stephancill

deodad:
hey @therealharpaljadeja this was implemented in #1128, the issue didn't get closed but closing now

PR #1128, opened 2023-07-13
feat: support fallback RPC providers

stephancill:
CLI now supports comma separated RPC URLs to fall back on (--eth-rpc-url, --eth-mainnet-rpc-url, --l2-rpc-url)

## Motivation

"This ensures that if a backend has become out-of-sync, or if it has been compromised that its responses are dropped in favor of responses that match the majority." #742.

## Change Summary

Implements backwards-compatible fallback URLs using viem's `fallback` transport, as suggested by #742.
Makes use of existing --eth-rpc-url, --eth-mainnet-rpc-url, --l2-rpc-url arguments, but splits by ',' to support multiple URLs to fall back on if others are not responding.

## Merge Checklist

## Additional Context

Ranking can also be enabled. This would periodically make additional calls to each RPC URL every 10 seconds to determine its ranking based on [latency and stability](https://viem.sh/docs/clients/transports/fallback.html#transport-ranking). Perhaps this configuration could be exposed to the user via CLI?

---

## PR-Codex overview
This PR focuses on adding support for fallback RPC providers in the Hubble application. 

### Detailed summary
- Added `fallback` function to the `viem` library to handle fallback RPC providers.
- Modified `L2EventsProvider` and `EthEventsProvider` classes to support multiple RPC URLs.
- Updated the `Hub` class to use the `rankRpcs` option and fallback transports for RPC providers.

### DIFF
--- a/.changeset/tasty-countries-jump.md
+++ b/.changeset/tasty-countries-jump.md
@@ -1,0 +1,5 @@
+---
+"@farcaster/hubble": minor
+---
+
+feat: support fallback RPC providers
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -52,9 +52,10 @@ app.name("hub").description("Farcaster Hub").version(APP_VERSION);
 app
   .command("start")
   .description("Start a Hub")
-  .option("-e, --eth-rpc-url <url>", "RPC URL of a Goerli Ethereum Node")
-  .option("-l, --l2-rpc-url <url>", "RPC URL of a Goerli Optimism Node")
-  .option("-m, --eth-mainnet-rpc-url <url>", "RPC URL of a mainnet Ethereum Node")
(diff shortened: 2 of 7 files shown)


deodad:
Thanks @stephancill!

deodad:
> Ranking can also be enabled. This would periodically make additional calls to each RPC URL every 10 seconds to determine its ranking based on [latency and stability](https://viem.sh/docs/clients/transports/fallback.html#transport-ranking). Perhaps this configuration could be exposed to the user via CLI?

I'm supportive

stephancill:
@deodad do you think there should be a single --rank-rpc option that applies to all RPCs or should there be one for each kind of RPC?

deodad:
> @deodad do you think there should be a single --rank-rpc option that applies to all RPCs or should there be one for each kind of RPC?

Starting with a single option at the CLI level to turn on automatic ranking globally is good. A more sophisticated approach would allow a config file where users could do fine grained configuration of urls, ranking parameters per chains, but let's take the simple win first. Thank you!

deodad:
This is awesome thank you!

PR #1165, opened 2023-07-19
feat: Adds support for storage/logging if would be pruned

CassOnMars:
## Motivation

## Change Summary

Adds storage rent support to pruning logic, only warns about pruning if storage rent has not been purchased.

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR focuses on adding support for logging would-be-pruned events and handling extra storage if purchased in the `@farcaster/hubble` package. 

### Detailed summary
- Adds support for logging would-be-pruned events
- Handles extra storage if purchased

> The following files were skipped due to too many changes: `apps/hubble/src/storage/stores/store.test.ts`

### DIFF
--- a/apps/hubble/src/storage/stores/storageCache.ts
+++ b/apps/hubble/src/storage/stores/storageCache.ts
@@ -21,6 +21,7 @@ import {
   RentRegistryEvent,
   StorageRegistryEventType,
 } from "@farcaster/core";
+import { getNextRentRegistryEventFromIterator, getRentRegistryEventsIterator } from "../db/storageRegistryEvent.js";
 
 const makeKey = (fid: number, set: UserMessagePostfix): string => {
   return Buffer.concat([makeFidKey(fid), Buffer.from([set])]).toString("hex");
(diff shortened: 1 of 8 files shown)


apps/hubble/src/storage/stores/store.ts
```diff
@@ -282,7 +294,7 @@ export abstract class Store<TAdd extends Message, TRemove extends Message> {
         }
 
         if (
-          count.value <= this._pruneSizeLimit &&
+          count.value <= this._pruneSizeLimit * unitsMultiplier &&
```
adityapk00, line 297: The same change also needs to be made in storeEventHandler.isPruneable() (Or in the calling fn at store.merge())

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -78,6 +116,22 @@ export class StorageCache {
...
+      return ok(0);
+    }
+
+    const time = getFarcasterTime();
+
+    if (time.isErr()) {
+      return err(time.error);
+    }
+
+    return ok(slots.filter((s) => s.from <= time.value && s.to >= time.value).reduce((accum, s) => accum + s.units, 0));
```
sanjayprabhu, line 132: This could be an expensive check to run on every message merge (within `isPrunable`). We might want to cache this by fid and then bust the cache when the rent expires. We could schedule the job to run a few seconds after the rent expires when we process the rent event. For example see `UpdateNameRegistryEventExpiryJobWorker`.

CassOnMars: Let me know if this approach works better – the contract does not provide a method to determine whether expiry has occurred in the same way as the name registry, so instead I have it collapsing the cached info into a single item per fid, and refresh individually when the cache invalidation time has passed.

sanjayprabhu: Nice yeah, that should work. Worth adding a test to make sure the invalidation happens as we expect. Apart from that, lgtm

Issue #1166, opened 2023-07-19
bug(examples/postgres-replicator): Casts table is missing image url embeds


manan19:
**What is the bug?**
The casts table is missing the image URL. Assuming the URL is supposed to be part of the embeds because it was a screenshot pasted into the desktop Merkle client.

Ran the following query. Attaching screenshot results.
`select * from casts where fid = 191 order by timestamp desc limit 5`
![image](https://github.com/farcasterxyz/hub-monorepo/assets/377624/2708cbca-b7ce-4da6-9dca-4fe70b0c38b8)

vs. relevant casts with screenshots - https://warpcast.com/manan/0xe9ae9d, https://warpcast.com/manan/0xc06f71

**How can it be reproduced? (optional)**
Don't know yet. Maybe other Postgres replicator users can try to confirm if they're seeing the same issue.

davidfurlong:
The example uses the `embedsDeprecated` property; it should be updated to use the new `embedCastIds` and `embedCastUrls`, with `embedsDeprecated` being a fallback for `embedCastUrls`

manan19:
This was fixed in #1174

PR #1174, opened 2023-07-21
fix: unique hash on links, index embeds

manan19:
## Motivation
Fixes 2 bugs in the indexer
1. https://github.com/farcasterxyz/hub-monorepo/issues/1166.
2. links table grew to 110M rows in the absence of uniqueness constraint.

## Change Summary

- Removes the added migration from #1113 because everyone would have to start over anyway
- Added uniqueness constraints to the `links` table.
- Indexes `embeds` in addition to `embedsDeprecated`
- Bumped the hub-nodejs version being used

_**Note - These changes are breaking and are not compatible with existing postgres instances with indexed data. Starting over is the easiest path to fixing these issues.**_

## Merge Checklist

---

## PR-Codex overview
This PR focuses on updating dependencies and making changes related to the `embeds` property in the `hubReplicator.ts` file.

### Detailed summary
- Updated `@farcaster/hub-nodejs` dependency from version `0.8.4` to `0.9.1`.
- Updated `@farcaster/core` dependency from version `0.10.2` to `0.11.1`.
- Modified the `embeds` property in the `db.ts` file to accept an array of objects.
- Added a new import for `Embed` in the `hubReplicator.ts` file.
- Modified the `embeds` property in the `hubReplicator.ts` file to handle both the new and deprecated formats.
- Added a new `onConflict` clause in the `HubReplicator` class in the `hubReplicator.ts` file.
- Modified the `addColumn` method in the `001_initial_migration.ts` file to use the `jsonb` type for the `embeds` column.
- Added a new `hash` column in the `002_add_hash_to_links.ts` migration file.
- Added unique constraints for the `hash` and `fid`, `targetFid`, `type` columns in the `links` table in the `001_initial_migration.ts` file.
- Updated the `@farcaster/hub-nodejs` dependency in the `yarn.lock` file.


packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
```diff
@@ -407,7 +407,8 @@ export class HubReplicator {
           parentHash: message.data.castAddBody.parentCastId?.hash,
           parentFid: message.data.castAddBody.parentCastId?.fid,
           parentUrl: message.data.castAddBody.parentUrl,
-          embeds: message.data.castAddBody.embedsDeprecated,
+          embedsDeprecated: message.data.castAddBody.embedsDeprecated,
+          embeds: message.data.castAddBody.embeds,
```
manan19, line 411: image

embeds is getting stored as an array of url encoded strings instead of an array of objects. Any idea what might be wrong here?

packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
```diff
@@ -631,6 +632,10 @@ export class HubReplicator {
           displayTimestamp: farcasterTimeToDate(message.data.linkBody.displayTimestamp),
         })),
       )
+      // Do nothing on conflict since nothing should have changed if hash is the same.
+      .onConflict((oc) => oc.columns(["hash"]).doNothing())
+      // Do nothing on conflict since nothing should have changed if the fid, targetFid & type are the same.
+      .onConflict((oc) => oc.columns(["fid", "targetFid", "type"]).doNothing())
```
sds, line 638: You aren't able to have multiple `ON CONFLICT` clauses in an insert statement. Since `hash` is computed from `fid` + `targetFid` + `type`, just need to use `hash` as the conflict column.

sds (commented):
Thanks for this. It's unclear why you're getting strings stored—would need to dig further, but it's not obvious to me from this PR why that would be happening.

davidfurlong:
Also this PR is missing the code changes here: https://github.com/farcasterxyz/hub-monorepo/blob/083762e5d0834fb33712992b0632ed4c4c78f720/packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts#L410

Here's what I'm doing in my repo

```ts
         ...
         parentUrl: message.data.castAddBody.parentUrl,
         embedUrls: [
            ...(message.data.castAddBody.embedsDeprecated || []),
            ...message.data.castAddBody.embeds
              .filter((x) => x.url)
              .map((x) => x.url),
          ],
          embedCastIds: message.data.castAddBody.embeds
            .filter((x) => x.castId)
            .map((x) => x.castId),
          mentions: message.data.castAddBody.mentions,
          ...
```

manan19:
@davidfurlong The concern with flattening is that it will become application specific. 

e.g. consider an application that wanted to maintain the order of the embeds so that they can generate a story based on images, videos & casts. The schema with `embedUrls` and `embedCastIds` being independent arrays does not maintain the order of embeds inserted.

davidfurlong:
> @davidfurlong The concern with flattening is that it will become application specific.
> 
> e.g. consider an application that wanted to maintain the order of the embeds so that they can generate a story based on images, videos & casts. The schema with `embedUrls` and `embedCastIds` being independent arrays does not maintain the order of embeds inserted.

@manan19 I don't think the current `embed` object's property order is intended to be consistent or user defined; typically object properties are considered unordered, or else it'd be an array imo. I might be wrong

davidfurlong:
Ok I've looked at it and you're right

manan19:
@davidfurlong Re: not inheriting the schema for `embedsDeprecated`

Is your suggestion to create `Embed[]` with URL strings from the `embedsDeprecated string[]` and then inserted that into the `embeds` (jsonb) column?

davidfurlong:
yep

manan19:
@davidfurlong @sds Updated to address both comments.

sds (approved):

sds:
Thanks!

PR #1223, opened 2023-08-06
fix: resolves conflicts for links_fid_target_fid_type_unique

### LINKED ISSUE #1221: bug(examples/replicate-data-postgres): Duplicate Key Violation in onLinkAdd function
https://github.com/farcasterxyz/hub-monorepo/blob/65a4faff6656dba276468f92bd9efde49e2dbcca/packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts#L639

### Description:
I encountered an error in the onLinkAdd function. The error message suggests that there is a violation of a unique constraint, specifically "links_fid_target_fid_type_unique," when inserting new records into the "links" table.

Postgres will fail, breaking the Promise. This will cause the application to fail and exit with status code 1. 

### Error Message:

```console
db-postgres-1  | 2023-08-05 12:51:33.130 UTC [36] ERROR:  duplicate key value violates unique constraint "links_fid_target_fid_type_unique"
db-postgres-1  | 2023-08-05 12:51:33.130 UTC [36] DETAIL:  Key (fid, target_fid, type)=(976, 251, follow) already exists.
db-postgres-1  | 2023-08-05 12:51:33.130 UTC [36] STATEMENT:  insert into "links" ("hash", "timestamp", "target_fid", "type", "fid") values ($1, $2, $3, $4, $5) on conflict ("hash") do nothing
```

...

```console
PostgresError: duplicate key value violates unique constraint "links_fid_target_fid_type_unique"
```

### Expected Behavior:
The `onLinkAdd` function should gracefully handle conflicts by retaining the existing record in the "links" table when encountering duplicate entries for the "hash" column. In case of a conflict, the function should take no action and proceed with inserting the next record without throwing an error or modifying the existing data.

### Proposed Solution:

1. Modify the onLinkAdd function to gracefully log out the failure.
2. `try { } catch { } finally { }`, however none of the other functions are written this way.

### Environment Details:
Using docker-compose.yml provided.

Node.js version: 20
PostgreSQL version: 15
Operating System: macOS M1/ARM

### Steps to Reproduce:

1. Deploy the provided Docker Compose configuration.
3. Trigger the `onLinkAdd` function with a set of messages that include duplicate values for "hash," "fid," "target_fid," and "type" columns.
4. Observe the logs for the Postgres container to encounter the duplicate key violation error.

manan19:
## Motivation

Fixes #1221 

## Change Summary

Add a way to handle conflicts for this constraint links_fid_target_fid_type_unique

## Merge Checklist

---

### Focus of the PR:
This PR focuses on updating the `onConflict` clause in the `hubReplicator.ts` file to include additional columns and update their values if a conflict occurs.

### Detailed summary:
- Updated the `onConflict` clause in the `hubReplicator.ts` file.
- Added columns `fid`, `targetFid`, and `type` to the `onConflict` clause.
- Added an update set for the columns `updatedAt`, `deletedAt`, `hash`, `timestamp`, and `displayTimestamp`.
- Added a condition to the `where` clause to check if `links.timestamp` is less than `excluded.timestamp`.
- Executed the updated `onConflict` clause.

### DIFF
--- a/packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
+++ b/packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
@@ -637,6 +637,8 @@ export class HubReplicator {
       )
       // Do nothing on conflict since nothing should have changed if hash is the same.
       .onConflict((oc) => oc.columns(["hash"]).doNothing())
+      // Do nothing on conflict since nothing should have changed if the fid, targetFid & type are the same.
+      .onConflict((oc) => oc.columns(["fid", "targetFid", "type"]).doNothing())
       .execute();
 
     for (const message of messages) {
--- a/packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
+++ b/packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
@@ -636,7 +636,18 @@ export class HubReplicator {
         })),
       )
       // Do nothing on conflict since nothing should have changed if hash is the same.
-      .onConflict((oc) => oc.columns(["hash"]).doNothing())
+      .onConflict((oc) =>
+        oc
+          .columns(["fid", "targetFid", "type"])
+          .doUpdateSet(({ ref }) => ({
+            updatedAt: new Date(),
+            deletedAt: null,
+            hash: ref("excluded.hash"),
+            timestamp: ref("excluded.timestamp"),
+            displayTimestamp: ref("excluded.displayTimestamp"),
+          }))
+          .where(({ cmpr, ref }) => cmpr("links.timestamp", "<", ref("excluded.timestamp"))),
+      )
       .execute();
 
     for (const message of messages) {
@@ -649,12 +660,13 @@ export class HubReplicator {
   private async onLinkRemove(messages: LinkRemoveMessage[]) {
     for (const message of messages) {
       await this.db
-        .deleteFrom("links")
+        .updateTable("links")
         .where("fid", "=", message.data.fid)
         // type assertion due to a problem with the type definitions. This field is infact required and present in all valid messages
         // rome-ignore lint/style/noNonNullAssertion: legacy code, avoid using ignore for new code
         .where("targetFid", "=", message.data.linkBody.targetFid!)
         .where("type", "=", message.data.linkBody.type)
+        .set({ deletedAt: farcasterTimeToDate(message.data.timestamp) })
         .execute();
 
       // TODO: Execute any cleanup side effects to remove the cast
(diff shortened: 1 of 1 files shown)


hellno:
I've dropped the "links" table, ran the migrations again and now running your branch → works 🙏️️ 

generally I've preferred to get a full copy and keep links that have been deleted. but happy with the fix so I can run a sync again

packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
```diff
@@ -649,13 +649,12 @@ export class HubReplicator {
   private async onLinkRemove(messages: LinkRemoveMessage[]) {
     for (const message of messages) {
       await this.db
-        .updateTable("links")
+        .deleteFrom("links")
```
sds, line 652: This would be the only time we're hard-deleting a record (`deleteFrom`) instead of soft-deleting it. Maybe that's the right tradeoff, but when I look at light 639 above, I'm realizing that I perhaps made the [wrong suggestion](https://github.com/farcasterxyz/hub-monorepo/pull/1174#discussion_r1270994800) to use `hash` as the conflict key instead of `[fid, targetFid, type]`. I was forgetting the fact that hash would also change if the timestamp changed, and since link compaction results in updated timestamps, this would result in no conflicts detected if we used `hash` as the conflict key.

So I would change:
```ts
.onConflict((oc) => oc.columns(["hash"]).doNothing())
```

...to:

```ts
.onConflict((oc) => 
  oc
    .columns(["fid", "targetFid", "type"])
    .doUpdateSet(({ ref }) => ({ 
      updatedAt: new Date(),
      hash: ref("excluded.hash"), 
      timestamp: ref("excluded.timestamp"),
      displayTimestamp: ref("excluded.displayTimestamp"),
    }))
    .where(({ cmpr, ref }) => cmpr("links.timestamp", ">", ref("excluded.timestamp")))
)
```

In other words: updating the row if a "newer" link of the same type was inserted; otherwise ignoring it.

This way you still have data on when the link was removed, rather than losing that information via hard deletes.

sds (commented):
Thanks for the PR! Sorry for the hassle on this. I think I see a way we can work around the issue.

packages/hub-nodejs/examples/replicate-data-postgres/hubReplicator.ts
manan19: @sds If I were to remove the hard delete, should the `doUpdateSet` also nullify the `deletedAt` timestamp?

sds: @manan19 Good call—yes, it should reset it to `null` since we're adding.

sds (approved):

sds:
Thanks!

christopherwxyz:
Thank you, everyone!

PR #1235, opened 2023-08-08
fix: Move rust code to apps/hubble, @farcaster/core to PureJS

adityapk00:
## Motivation

Move all the Rust code to hubble, where it will be built as a part of the app/in docker. Make the @farcaster/core package PureJS so that it can be published to npm. This way, we keep all the perf gains of rust in hubble, but don't get into compatibility issues of the natively-build Rust code in a published library.

## Change Summary

- Remove all rust code from @farcaster/core, making it PureJS
- Move all rust code to apps/hubble. Pass in native implementations of blake3 and ed25519 when running hubble.

## Merge Checklist

## Additional Context

---

### Focus of this PR:
Move the Rust code from the `core` package to the `hubble` app, making the `core` package PureJS.

### Detailed summary:
- Move Rust code from `core/src/addon` to `hubble/src/addon`
- Update `rome.json` in both `core` and `hubble` packages
- Update imports and exports in various files
- Add bridge functions between Rust and TypeScript
- Update dependencies and scripts in `hubble/package.json` and `core/package.json`
- Update validation methods in `core/src/validations.ts`


apps/hubble/src/storage/engine/validation.worker.ts
```diff
@@ -1,11 +1,12 @@
 import { validations } from "@farcaster/hub-nodejs";
+import { nativeValidationMethods } from "../../rustfunctions.js";
 import { parentPort } from "worker_threads";
 
 // Wait for messages from the main thread and validate them, posting the result back
 parentPort?.on("message", (data) => {
   (async () => {
     const { id, message } = data;
-    const result = await validations.validateMessage(message);
+    const result = await validations.validateMessage(message, nativeValidationMethods);
```
sanjayprabhu, line 9: We should also do this on `Engine.validateMessage` (the non worker code path) for tests. 

Also, worth ensuring we have test coverage of the native functions at the hubble level (if we just `return true` in `nativeEd25519Verify` does anything fail right now? I suspect nothing will)

adityapk00: Good catch on the test path. Fixed it. 

Although,  the `rustfunctions.test.ts` tests the methods. (i.e., if we return true in `nativeEd25519Verify`, this test fails) The test compares the native functions' output to the PureJS implementation.

sanjayprabhu (approved):

PR #1253, opened 2023-08-11
fix: Check index keys are > UserMessagePostfixMax

adityapk00:
## Motivation

Index keys must use postfix that is > UserMessagePostfixMax

## Merge Checklist

---

## PR-Codex overview
This PR focuses on fixing an issue related to index keys in the `Store` class. 

### Detailed summary
- Added a check to ensure that index keys are greater than `UserMessagePostfixMax`.
- Modified the `makeAddKey` and `makeRemoveKey` methods in the `Store` class to include additional key components.
- Added checks in the `Store` class to prevent the use of message keys for the Adds and Removes indices.


apps/hubble/src/storage/stores/store.ts
```diff
@@ -605,6 +605,19 @@ export abstract class Store<TAdd extends Message, TRemove extends Message> {
     // Puts the message into the TAdds Set index
     // rome-ignore lint/suspicious/noExplicitAny: legacy code, avoid using ignore for new code
     const addsKey = this.makeAddKey(message as any);
+
+    // Ensure that the Adds key is using a Postfix > UserMessagePostfixMax
+    const keypostfix = addsKey.readUint8(1 + FID_BYTES);
+    if (keypostfix <= UserMessagePostfixMax) {
```
sanjayprabhu, line 611: Yeah, this would work. Two things though:
1. Should we only enable this in the test env?
2. Theoretically this can also happen in `buildSecondaryIndices` and the same approach won't work

But overall I like this

adityapk00: I'm a bit hesitant to add test code inside regular prod code. It's a bit weird, no? I wonder if we can write a rome linter rule instead that does some type of static check? Let me look into it

PR #1297, opened 2023-08-29
feat: Support onchain events and username proofs in replicator

sanjayprabhu:
## Motivation

Update the replicator to handle L2 migration

## Change Summary

## Merge Checklist

## Additional Context

### DIFF
--- a/packages/hub-nodejs/examples/replicate-data-postgres/docker-compose.yml
+++ b/packages/hub-nodejs/examples/replicate-data-postgres/docker-compose.yml
@@ -2,7 +2,13 @@ version: '3.9'
 
 services:
   app:
-    image: 'node:20.2-alpine'
+    build:
+      context: .
+      # Need to install python3 due to ffi requirements in core. Once we release a new version of core removing the dep,
+      # we can remove this.
+      dockerfile_inline: |
+        FROM node:20.2-alpine
+        RUN apk add --no-cache libc6-compat python3 make g++ linux-headers
     restart: unless-stopped
     command: ["sh", "-c", "yarn install && exec yarn start"]
     init: true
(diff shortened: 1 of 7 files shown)


packages/hub-nodejs/examples/replicate-data-postgres/docker-compose.yml
```diff
@@ -2,7 +2,13 @@ version: '3.9'
 
 services:
   app:
-    image: 'node:20.2-alpine'
+    build:
+      context: .
+      # Need to install python3 due to ffi requirements in core. Once we release a new version of core removing the dep,
+      # we can remove this.
+      dockerfile_inline: |
```
sds, line 9: TIL about `dockerfile_inline`—convenient!

packages/hub-nodejs/examples/replicate-data-postgres/migrations/002_mainnet.ts
```diff
@@ -0,0 +1,63 @@
...
+  await db.schema.alterTable("signers").dropConstraint("signers_hash_unique").execute();
+
+  await db.schema.alterTable("signers").dropConstraint("signers_hash_foreign").execute();
+
+  await db.schema.alterTable("signers").addUniqueConstraint("signers_fid_signer_unique", ["fid", "signer"]).execute();
+
+  await db.schema
+    .alterTable("signers")
+    .alterColumn("hash", (c) => c.dropNotNull())
+    .alterColumn("custodyAddress", (c) => c.dropNotNull())
```
sds, line 16: Should we just drop the columns instead? Seems like we're trying to support both old and new world here, but that might be difficult. We've broken things before and it seems fine to expect people to just clear their DBs and start from scratch.

Then again, if the intent is to have this working throughout the course of the migration, what is done here probably makes sense. Just hard to reason about.

sds (approved):
Awesome, thank you for this.

If we wanted to drop support for the old world, that might make this a bit easier, but since you've already done the work, might as well keep what you have here.

packages/hub-nodejs/examples/replicate-data-postgres/migrations/002_mainnet.ts
sanjayprabhu: Yes, I didn't want to break people's existing setup. Once migration is complete, and we've deleted the code in the hub, we can remove it here as well.

PR #1506, opened 2023-10-11
chore: Calculate earliestTimestamp and numMessages on demand instead of at startup

adityapk00:
## Change Summary

- Move calculation of earliest Ts hash and num Messages per fid + set to on-demand instead of at startup, which is blocking

## Merge Checklist

---

## PR-Codex overview
This PR focuses on improving the performance of message count calculations during startup. 

### Detailed summary
- Made message counts on-demand to speed up startup
- Added prepopulation of message counts in the background
- Changed `processEvent` method to return a promise


apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -63,32 +62,6 @@ export class StorageCache {
 
     const progressBar = addProgressBar("Syncing storage cache", totalFids * 2);
 
-    let lastFid = 0;
-    const prefix = Buffer.from([RootPrefix.User]);
-    await this._db.forEachIteratorByPrefix(
```
sanjayprabhu, line 68: I think this would be bad, it would really kill p95 merge performance. We need to look this up on every merge to determine whether a message is prunable. 

One thing we could do is make this non-blocking. Fill up the cache in the background and lazy lookup in the meantime.

sanjayprabhu: What kind of user is it? i.e. Is this 80ms for a user that has hit their limits or just the median user. Also, this would be per fid per set.

Also, this would be per fid per set. Maybe since we're iterating the fid anyway, we should do it for all sets for that fid.

> The problem of doing it non-blocking would be that we'd actually have to block the merges until this value is available, which would block startup (i.e., what we have now)

Hmm, I'm not sure I follow? There are 3 options:
1. Block startup and calculate this for every fid. Merges cannot complete until this is done for every fid (behavior on main)
2. Don't block startup, merge for every fid/set is blocked until we calculate this value on first message (this PR)
3. Same as 2, but also run a background job to fill the cache for all fids. (slow initial performance, but after cache is full in a few minutes, same perf as 1)

I guess there's also a fourth option, which is persist this into the db so we don't have to re-calculate it each time. Now that we have a common Store class, this might be doable. But downside is if we make a mistake could lead to weird sync bugs.

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -112,23 +85,26 @@ export class StorageCache {
...
-    this._counts = usage;
+    this._counts = new Map();
     this._earliestTsHashes = new Map();
+
     log.info({ timeTakenMs: Date.now() - start }, "storage cache synced");
   }
 
   async getMessageCount(fid: number, set: UserMessagePostfix): HubAsyncResult<number> {
     const key = makeKey(fid, set);
     if (this._counts.get(key) === undefined) {
```
sanjayprabhu, line 96: Hmm, take a look at `addMessage` and `removeMessage` below, it may theoretically be possible for them to be hit before this is called for the first time which could end up setting this key. We might need to call this to populate the count right before we modify the cache.

adityapk00: Ah, I misunderstood what you were saying. (3) above is totally doable. I'll fix the `addMessage` and `removeMessage` too

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -112,23 +85,59 @@ export class StorageCache {
...
+            await sleep(10); // Sleep to allow other threads to run
+
+            prevFid = fid;
+            prevPostfix = postfix;
+          }
+        }
+      },
+      { values: false },
+      1 * 60 * 60 * 1000, // 1 hour
+    );
```
sanjayprabhu, line 120: We should log when this is complete

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -202,7 +211,7 @@ export class StorageCache {
     }
   }
 
-  processEvent(event: HubEvent): HubResult<void> {
+  async processEvent(event: HubEvent): HubAsyncResult<void> {
```
sanjayprabhu, line 214: We need to await in the caller of this.

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -224,12 +233,12 @@ export class StorageCache {
...
   }
 
-  private addMessage(message: Message): void {
+  private async addMessage(message: Message): Promise<void> {
     if (message.data !== undefined) {
       const set = typeToSetPostfix(message.data.type);
       const fid = message.data.fid;
       const key = makeKey(fid, set);
-      const count = this._counts.get(key) ?? 0;
+      const count = this._counts.get(key) ?? (await this.getMessageCount(fid, set)).unwrapOr(0);
```
sanjayprabhu, line 241: Do we need a write lock? I think it's theoretically possible for the prepopulate to fill the count while awaiting here. But not sure if it's possible for the values to be different. But even if it's not possible for the values to be different now, may not be safe to rely on that being the case.

Ultimately, the costs of getting this wrong by a few items here or there is not huge, except for users at their limit, then we'll potentially prune early or prune too late, causing hubs to go out of sync.

adityapk00: There is a possibility of a race condition, but we don't need a lock. On line 143, we can recheck if another thread populated it before updating, preventing us from overwriting values

apps/hubble/src/storage/stores/storageCache.ts
adityapk00: The caller is the event listener, which doesn't need to be awaited. Each event listner is in a `setTimeout`, where it doesn't wait for the listner's handling to finish, so awaiting doesn't do anything.

apps/hubble/src/storage/stores/storageCache.ts
adityapk00: Done

apps/hubble/src/storage/stores/storageCache.ts
sanjayprabhu: Wouldn't we broadcast the event and release the commit lock here before this processing is complete?

https://github.com/farcasterxyz/hub-monorepo/blob/8cd6b59f076ff733e0c35036563011505853c8c8/apps/hubble/src/storage/stores/storeEventHandler.ts#L334

adityapk00: right, we don't `await` here (and do `void`) because we don't want to block downstream processing of event handlers, and release the commit lock as soon as possible, since it blocks merging. 

Event handlers are expected to handle overlapping events, which we do by doing essentially an atomic update on line 143

sanjayprabhu: No, I actually meant the caller, but TIL.

PR #1508, opened 2023-10-11
fix: Accept raw databytes if using alternate protobuf encoding

adityapk00:
## Motivation

We are currently dependent on ts-proto protobuf encoding, because hashes are computed on the ts-proto's protobuf encoding of message.data

As an alternative to that, we now allow users to submit a raw byte stream of data_bytes (instead of the Message.data field). If this field is provided, we'll use that to compute the hash and the signature. 

## Change Summary

- Add dataBytes to message protobuf
- When saving to disk, strip out data field if dataBytes is present, and re-add it when reading to save space on disk
- when submitMessage, make sure that the data field is populated from the dataBytes field. 

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR introduces the ability to sign raw message data bytes to support Rust and Golang. 

### Detailed summary
- Added support for signing raw message data bytes in Rust example
- Updated Cargo.toml with new dependencies for Rust example
- Updated message.proto to include optional `data_bytes` field
- Updated MessageData interface and generated message.ts to include `dataBytes` field
- Updated various files to use `messageDecode` function instead of directly decoding messages
- Updated validation tests to handle `dataBytes` field
- Updated documentation to include `data_bytes` field in Messages.md and messages.md

> The following files were skipped due to too many changes: `packages/hub-web/src/generated/message.ts`, `packages/hub-nodejs/src/generated/message.ts`, `packages/core/src/protobufs/generated/message.ts`, `packages/core/src/validations.ts`, `packages/hub-web/examples/rust-submitmessage/src/main.rs`, `apps/hubble/www/docs/docs/httpapi/submitmessage.md`, `apps/hubble/src/storage/db/message.ts`, `apps/hubble/src/storage/engine/messageDataBytes.test.ts`, `packages/hub-web/examples/rust-submitmessage/src/username_proof.rs`, `packages/hub-web/examples/rust-submitmessage/Cargo.lock`, `packages/hub-web/examples/rust-submitmessage/src/message.rs`


apps/hubble/src/storage/engine/messageDataBytes.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      const newBytes = Buffer.concat([bytes.slice(0, index), Buffer.from(newVarintBytes), bytes.slice(varintEndIndex)]);
+
+      // Decode and verify
+      // const decodedMsg = MessageData.decode(newBytes);
+      // console.log("Decoded message: ", decodedMsg);
+
+      return newBytes;
+    };
+
+    test("varint encoding", async () => {
```
sanjayprabhu, line 208: Nice

apps/hubble/www/docs/docs/httpapi/submitmessage.md
```diff
@@ -82,4 +82,90 @@ try {
 } catch (e) {
     // handle errors...
 }
+```
+
+## Using with Rust, Go or other programing languages
+Messages need to be signed with a ED25519 signer belonging to the FID. If you are using a different programming language than Typescript, you can manually construct the `MessageData` object and serialize it to the `data_bytes` field of the message. Then, use the `data_bytes` to compute the `hash` and `signature`. Please see the [`rust-submitmessage` example](https://github.com/farcasterxyz/hub-monorepo/tree/main/packages/hub-web/examples) for more details
```
sanjayprabhu, line 88: We should link to `hub-nodejs/examples` since web will go away soon.

packages/core/src/validations.ts
```diff
@@ -126,16 +126,30 @@ export const validateMessage = async (
...
 
-  const dataBytes = protobufs.MessageData.encode(data).finish();
-  if (message.hashScheme === protobufs.HashScheme.BLAKE3) {
-    const computedHash = validationMethods.blake3_20(dataBytes);
+  // Computed from the data_bytes if set, otherwise from the data
+  let computedHash;
+
+  // 2. If the data_bytes are set, we'll validate signature against that
+  if (message.dataBytes && message.dataBytes.length > 0) {
+    if (message.dataBytes.length > 1024) {
```
sanjayprabhu, line 141: Hmm, I was thinking about this and maybe 1mb is too generous. With the default limits, for 5$ you could still cause hubs store 5gb of casts+ 2.5gb of links and 2.5gb of reactions. 10gb total. that seems like a lot. 

Maybe we need a smarter check here, could we re-encode and verify that the the re-encoded length is not off by more than 25 bytes or something? That should be more than sufficient to account for issues like the varint encoding.

apps/hubble/www/docs/docs/httpapi/submitmessage.md
adityapk00: Is `hub-web` going away? I thought we're just going to pivot it to using the HTTP API examples (and remove the grpc-web stuff)

packages/core/src/validations.ts
sanjayprabhu: Oh, 🤦. That's right, 10mb is not too bad.

apps/hubble/www/docs/docs/httpapi/submitmessage.md
sanjayprabhu: There's no reason to have it around once grpc-web is gone? You could get everything you wanted from `core` + http api if you wanted to use it in a web context.

adityapk00: Ok, that makes sense. In that case, I have to move a lot of stuff off `hub-web`, so I'll do it in a different PR

sanjayprabhu (approved):
This could potentially introduce new messages that can't be validated by existing hubs, but I don't think it's a huge issue. Don't expect anyone will use it that fast.

PR #1530, opened 2023-10-18
fix: Remove PubSub based PeerDiscovery

sagar-a16z:
## Motivation

Hubble instances were running out of Memory due to excessive duplicated traffic on Gossip.

## Change Summary

Revoved PubSub based PeerDiscovery in favor of the existing ContactInfos being shared. This reduces amount of network traffic by over 10x while still allowing nodes to be discovered. 

The way ContactInfos are added to the Peer Store's AddressBook suffices as a discovery mechism inside of libp2p. 

## Merge Checklist

## Additional Context

The only outstanding consideration with this change is that now, Hubs only publish their own ContactInfos once every 4 hours, or on every new Connection. This means that a new Hub might not see the rest of the network immediately. We could reduce the 4 hour interval to something more reasonable once we test this change out.

---

## PR-Codex overview
This PR focuses on adding a timestamp field to the `ContactInfoContent` message in the `gossip.ts` file. 

### Detailed summary
- Added `timestamp` field to `ContactInfoContent` message in `gossip.ts`

> The following files were skipped due to too many changes: `apps/hubble/src/network/p2p/gossipNode.ts`, `apps/hubble/src/network/p2p/gossipNodeWorker.ts`, `apps/hubble/src/hubble.ts`


apps/hubble/src/hubble.ts
```diff
@@ -1003,6 +1039,12 @@ export class Hub implements HubInterface {
         log.error({ error: syncResult.error, peerId }, "Failed to sync with new peer");
       }
     }
+
+    // if the contact info doesn't include a timestamp, consider it invalid but allow the peer to stay in the address book
+    // TODO remove this once all peers are updated past 1.6.4
```
sagar-a16z, line 1044: Just calling this out as a workaround to still accept peer contacts but to not forward them if they don't include a timestamp.

apps/hubble/src/network/p2p/gossipNodeWorker.ts
```diff
@@ -212,7 +220,7 @@ export class LibP2PNode {
         }
       }
     }
-    return msgIdFnStrictSign(message);
+    return msgIdFnStrictNoSign(message);
```
sanjayprabhu, line 223: Why the NoSign version? Performance?

sagar-a16z: This is misleading. I've pulled this function in from libp2p. The NoSign MsgId just hashes the Message.data field. The Signed MsgId function from earlier instead uses a Key + randomly generated Seq number as the MsgId.

So tldr; This just uses a sha256 hash of the Gossip Message.

sanjayprabhu (approved):

PR #1545, opened 2023-10-24
feat: Add ContactInfoContent signing + config toggle for strict validation + StrictNoSign

CassOnMars:
## Motivation

StrictSign support on hubs is using javascript methods for signature creation and validation, which are both very slow compared to our rust implementation and make up a large portion of CPU time. To support use of StrictNoSign, we need signatures added to ContactInfoContent to verify provenance, and also need to allow network config based toggling of the features.

## Change Summary

This PR updates ContactInfoContent messages to include signature payloads following the message signing convention used for other Farcaster messages, and utilizes native methods. It also adds configuration support both at the hub options and network configuration level for contactinfo strict validation and StrictNoSign.

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR focuses on adding support for contact info content signing and introducing a new strictNoSign option. 

### Detailed summary
- Adds support for contact info content signing and strictNoSign option.
- Updates several files related to contact info content and signature handling.
- Includes changes to network configuration and validation methods.
- Adds a new native function for signing message hashes.
- Updates tests for native signing and verification methods.
- Makes changes to the libp2p gossipNode class for message handling.
- Includes changes to the libp2p node worker and sync engine.

> The following files were skipped due to too many changes: `packages/core/src/validations.ts`, `apps/hubble/src/network/utils/networkConfig.test.ts`, `packages/core/src/protobufs/generated/gossip.ts`, `apps/hubble/src/hubble.ts`, `packages/hub-web/src/generated/rpc.ts`


apps/hubble/src/hubble.ts
```diff
@@ -979,14 +1028,70 @@ export class Hub implements HubInterface {
...
+        content.signer,
+        nativeValidationMethods,
+      );
+      if (result.isErr()) {
+        log.debug({ message: content, error: result.error }, "signature verification failed for contact info");
+        return false;
+      }
+
+      if (!result.value) {
+        log.debug({ message: content }, "signature verification failed for contact info");
```
sanjayprabhu, line 1077: These signature verification failure logs should be `info` or `warn`

apps/hubble/src/hubble.ts
```diff
@@ -979,14 +1028,70 @@ export class Hub implements HubInterface {
...
+      let bytes: Uint8Array;
+      if (message.dataBytes) {
+        bytes = message.dataBytes;
+      } else {
+        bytes = ContactInfoContent.encode(message).finish();
+      }
+
+      const pubKey = unmarshalPublicKey(peerId.publicKey);
+      if (Buffer.compare(pubKey.marshal(), content.signer) !== 0) {
+        log.debug({ message: content }, "signer mismatch for contact info");
```
sanjayprabhu, line 1055: This should probably should not be debug. We'd want to know if this was happening.

apps/hubble/src/network/utils/networkConfig.test.ts
```diff
@@ -113,9 +123,11 @@ describe("networkConfig", () => {
       keyRegistryAddress: undefined,
       idRegistryAddress: undefined,
       allowlistedImmunePeers: undefined,
+      strictContactInfoValidation: undefined,
+      strictNoSign: undefined,
```
sanjayprabhu, line 127: Maybe time to extract this to a common function. Always annoying to have to modify all these test just to add an extra param.

apps/hubble/src/hubble.ts
```diff
@@ -710,6 +734,8 @@ export class Hub implements HubInterface {
       this.deniedPeerIds = deniedPeerIds;
 
       this.allowlistedImmunePeers = allowlistedImmunePeers;
+      this.strictContactInfoValidation = !!strictContactInfoValidation;
+      this.strictNoSign = !!strictNoSign;
```
sanjayprabhu, line 738: We'd need to restart the GossipNode node if this changed right?

apps/hubble/src/hubble.ts
```diff
@@ -979,14 +1028,70 @@ export class Hub implements HubInterface {
...
+      gossipAddress: content.gossipAddressContactInfo,
+      rpcAddress: content.rpcAddressContactInfo,
+      excludedHashes: content.excludedHashes,
+      count: content.count,
+      hubVersion: content.hubVersion,
+      network: content.network,
+      appVersion: content.appVersion,
+      timestamp: content.timestamp,
+      dataBytes: content.dataBytes,
+    });
```
sanjayprabhu, line 1044: `ContactInfoContent.decode(ContactInfoContent.encode(content).finish())` is more future proofs for when new fields get added. Maybe also worth pulling the dataBytes check out here for clarity.

apps/hubble/src/addon/src/lib.rs
```diff
@@ -1,8 +1,29 @@
 use std::convert::TryInto;
 
-use ed25519_dalek::{Signature, VerifyingKey};
+use ed25519_dalek::{Signature, Signer, SigningKey, EXPANDED_SECRET_KEY_LENGTH, VerifyingKey};
 use neon::{prelude::*, types::buffer::TypedArray};
 
+fn ed25519_sign_message_hash(mut cx: FunctionContext) -> JsResult<JsBuffer> {
```
sanjayprabhu, line 6: Not strictly required since we gossip the contact info so infrequently, but nice to have.

apps/hubble/src/hubble.ts
CassOnMars: Because we don't have a envelope to this, we have to explicitly extract the values that would exist pre-signed, but agreed we need a way to future proof this better.

apps/hubble/src/hubble.ts
CassOnMars: correct, will follow up

apps/hubble/src/hubble.ts
sanjayprabhu: We could introduce a `body` field just like Message and duplicate the fields on there since we're changing it anyway. We can deprecate the outer fields in 1.8

CassOnMars: sounds good, happy to proceed with this

apps/hubble/src/storage/jobs/updateNetworkConfigJob.ts
```diff
@@ -41,12 +41,18 @@ export class UpdateNetworkConfigJobScheduler {
...
+    const { shouldRestart, shouldExit } = this._hub.applyNetworkConfig(networkConfig.value);
     if (shouldExit) {
       log.error({}, "Network config exit signal");
       process.kill(process.pid, "SIGQUIT");
     }
 
+    if (shouldRestart) {
+      log.info({}, "Network config restart signal");
+      await this._hub.stop();
+      await this._hub.start();
```
sanjayprabhu, line 53: Could we just restart the gossipNode instead of the whole hub?

apps/hubble/src/storage/jobs/updateNetworkConfigJob.ts
```diff
@@ -41,12 +41,18 @@ export class UpdateNetworkConfigJobScheduler {
...
-    const shouldExit = this._hub.applyNetworkConfig(networkConfig.value);
+    const { shouldRestart, shouldExit } = this._hub.applyNetworkConfig(networkConfig.value);
     if (shouldExit) {
       log.error({}, "Network config exit signal");
       process.kill(process.pid, "SIGQUIT");
     }
 
+    if (shouldRestart) {
+      log.info({}, "Network config restart signal");
+      await this._hub.stop();
```
sanjayprabhu, line 52: `GossipNode.stop` calls terminate on the worker. https://github.com/farcasterxyz/hub-monorepo/blob/b47c65bb74b127eded35bf66f79a5c6b1f5ab5af/apps/hubble/src/network/p2p/gossipNode.ts#L292

I think we'll need to modify `start` to re-create the worker. Otherwise it won't start back up.

sanjayprabhu (commented):
Worth testing on testnet or locally with two hub instances.

apps/hubble/src/storage/jobs/updateNetworkConfigJob.ts
CassOnMars: There are calls being made to the worker which could behave unpredictably if we only restarted the gossipNode

apps/hubble/src/storage/jobs/updateNetworkConfigJob.ts
CassOnMars: good call, lmk if this approach handles this better

CassOnMars: https://github.com/farcasterxyz/hub-monorepo/pull/1545/commits/23e315d32d82760e4d9b3e2245dcc07b2410f25e

sanjayprabhu (approved):

PR #1677, opened 2024-02-07
fix: write out cached blocks while processing L2 events

adityapk00:
## Motivation

Write out batches of L2 blocks as they are processed instead of waiting for the end. 

## Change Summary

- Write out blocks in batches as they are processed instead of caching everything for the end, which might result in millions of events in RAM. 

## Merge Checklist

---

## PR-Codex overview
This PR focuses on fixing the process of handling L2 blocks in batches. 

### Detailed summary
- Fixed a bug where the process would exit if another hub process with a different process number was running.
- Changed the order of starting contract watchers and syncing historical events.
- Modified the `connectAndSyncHistoricalEvents` method to return the highest block that was synced.
- Added a new method `writeCachedBlocks` to write out all the cached blocks.
- Updated the `handleNewBlock` method to call `writeCachedBlocks` and handle new blocks.

### DIFF
--- a/apps/hubble/src/eth/l2EventsProvider.ts
+++ b/apps/hubble/src/eth/l2EventsProvider.ts
@@ -703,7 +703,7 @@ export class L2EventsProvider {
 
       // If there are more batches, write out all the cached blocks first
       if (i < numOfRuns - 1) {
-        this.writeCachedBlocks(nextToBlock);
+        await this.writeCachedBlocks(nextToBlock);
       }
     }
 
--- a/apps/hubble/src/eth/l2EventsProvider.ts
+++ b/apps/hubble/src/eth/l2EventsProvider.ts
@@ -177,12 +177,15 @@ export class L2EventsProvider {
 
   public async start() {
     // Connect to L2 RPC
-    await this.connectAndSyncHistoricalEvents();
 
-    this._watchBlockNumber?.start();
+    // Start the contract watchers first, so we cache events while we sync historical events
     this._watchStorageContractEvents?.start();
     this._watchIdRegistryV2ContractEvents?.start();
     this._watchKeyRegistryV2ContractEvents?.start();
+
+    await this.connectAndSyncHistoricalEvents();
+
+    this._watchBlockNumber?.start();
   }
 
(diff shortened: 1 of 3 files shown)


apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -700,6 +700,11 @@ export class L2EventsProvider {
       await this.processStorageEvents(await storageLogsPromise);
       await this.processIdRegistryV2Events(await idV2LogsPromise);
       await this.processKeyRegistryEventsV2(await keyV2LogsPromise);
+
+      // If there are more batches, write out all the cached blocks first
+      if (i < numOfRuns - 1) {
+        await this.writeCachedBlocks(nextToBlock);
```
sanjayprabhu, line 706: Should we use `latestBlock` here instead of `nextToBlock`?

sanjayprabhu: or just `toBlock` rather. I think blocks lower than the numConfirmations within a batch wouldn't be written out?

adityapk00: Ok, fixed it and started the watchers first. Can you please re-review?

sanjayprabhu (approved):
Nice, I think that should work.

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -799,12 +786,21 @@ export class L2EventsProvider {
...
-    progressBar?.update(totalBlocks);
-    progressBar?.stop();
+    this._isHandlingBlock = true;
+    log.info({ blockNumber }, `new block: ${blockNumber}`);
+    statsd().increment("l2events.blocks");
+
+    await this.writeCachedBlocks(blockNumber);
 
     // Update the last synced block if all the historical events have been synced
     if (this._isHistoricalSyncDone) {
```
sanjayprabhu, line 806: Maybe we should always write the blocknumber to the db now that we're writing the events in batches. Otherwise on restart, we'll re-sync from the beginning.

PR #1680, opened 2024-02-07
fix: prefer peers with more messages to sync with

sanjayprabhu:
## Motivation

When there are a lot of peers on the network that are behind, choosing a random peer is detrimental. Prefer to sync with peers that have more messages than us. If there are no such peers, pick a random peer.

## Change Summary

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR focuses on improving the peer selection process in the Hubble app. It prioritizes peers with more messages to sync with.

### Detailed summary
- Updated the cron schedule to run every 2 hours instead of every 4 hours.
- Modified the logic for selecting a random peer from the current list.
- Added a preference for peers with more messages to sync with.
- Added logging statements to indicate the selection criteria for the random peer.

### DIFF
--- a/apps/hubble/src/network/sync/syncEngine.ts
+++ b/apps/hubble/src/network/sync/syncEngine.ts
@@ -486,9 +486,10 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
       // Prefer hubs that have more messages than us, if no such hub is available, pick a random one
       const snapshotResult = await this.getSnapshot();
       if (snapshotResult.isOk()) {
-        peers = Array.from(this.currentHubPeerContacts.values()).filter(
-          (p) => p.contactInfo.count > snapshotResult.value.numMessages,
-        );
+        // Use a buffer of 5% of our messages so the peer with the highest message count does not get picked
+        // disproportionately
+        const messageThreshold = snapshotResult.value.numMessages * 0.95;
+        peers = Array.from(this.currentHubPeerContacts.values()).filter((p) => p.contactInfo.count > messageThreshold);
       }
 
       if (peers.length === 0) {
--- a/apps/hubble/src/network/sync/syncEngine.ts
+++ b/apps/hubble/src/network/sync/syncEngine.ts
@@ -481,14 +481,31 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
 
     // If we don't have a peer contact, get a random one from the current list
     if (!peerContact) {
-      // Pick a random key
-      const randomPeer = Array.from(this.currentHubPeerContacts.keys())[
-        Math.floor(Math.random() * this.currentHubPeerContacts.size)
-      ] as string;
+      let peers: PeerContact[] = [];
 
-      const c = this.currentHubPeerContacts.get(randomPeer);
-      peerContact = c?.contactInfo;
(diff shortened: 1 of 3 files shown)


apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -481,14 +481,31 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
...
+          { peersCount: this.currentHubPeerContacts.size, eligiblePeers: peers.length },
+          `Diffsync: Choosing random peer among ${peers.length} peers with fewer messages`,
+        );
+      } else {
+        log.info(
+          { peersCount: this.currentHubPeerContacts.size, eligiblePeers: peers.length },
+          `Diffsync: Choosing random peer among ${peers.length} peers with more messages`,
+        );
+      }
+      const randomPeer = peers[Math.floor(Math.random() * peers.length)];
```
sanjayprabhu, line 506: Hmm, this might need to be a weighted average, otherwise the peer with most messages could get a disproportionate number of sync requests? But should be ok for now.

apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -481,14 +481,31 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
...
+      let peers: PeerContact[] = [];
 
-      const c = this.currentHubPeerContacts.get(randomPeer);
-      peerContact = c?.contactInfo;
-      peerId = c?.peerId;
+      // Prefer hubs that have more messages than us, if no such hub is available, pick a random one
+      const snapshotResult = await this.getSnapshot();
+      if (snapshotResult.isOk()) {
+        peers = Array.from(this.currentHubPeerContacts.values()).filter(
+          (p) => p.contactInfo.count > snapshotResult.value.numMessages,
```
adityapk00, line 490: Maybe do within 5% of our messages? That way the biggest peer doesn't get disproportionate weight

apps/hubble/src/storage/jobs/gossipContactInfoJob.ts
```diff
@@ -20,7 +20,7 @@ export class GossipContactInfoJobScheduler {
 
   start(cronSchedule?: string) {
     const randomMinute = Math.floor(Math.random() * 60);
-    const defaultSchedule = `${randomMinute} */4 * * *`; // Every 4 hours at a random minute
+    const defaultSchedule = `${randomMinute} */2 * * *`; // Every 2 hours at a random minute
```
adityapk00, line 23: There's a 30 min timeout for syncs, so every 2 hours, 1.5 hours there won't be a sync running. Just making sure this is intentional, right? We'll bump it back up when things are stable?

sanjayprabhu: Yes, that's right.

I want to make another change that sync the last 2 hours worth when we receive a new contact info, with that I think we'll have decent coverage and maybe we don't need to increase sync frequency, but we'll see.

apps/hubble/src/network/sync/syncEngine.ts
sanjayprabhu: Good idea. Done.

PR #1688, opened 2024-02-09
fix: Stop old messages from being broadcast on gossip

adityapk00:
## Motivation

We're setting asyncValidation: true

With this setting true, libp2p doesn't forwarded immediately. They are forwarded only after the message has been merged, which can be delayed a lot on slow hubs. And then, after the message is merged it is broadcasted again to the next fanout peers and so on, adding up the delays at each step of the fanout. Essentially, each message is broadcast twice, and only when each hub reports it as reportValid(false) does the propogation stop.

This PR rejects messages older than seenTTL (even while we attempt to merge them) so that older messages don't flood the network

## Merge Checklist

---

## PR-Codex overview
This PR focuses on fixing a bug related to gossip message propagation. 

### Detailed summary
- Added a `timestamp` field to the `GossipMessage` protobuf schema.
- Modified the `handleGossipMessage` function to check the timestamp of incoming gossip messages.
- Added logic to report gossip messages as invalid if they are older than a certain cutoff time.
- Updated the `LibP2PNode` class to include the timestamp when publishing gossip messages.

### DIFF
--- a/apps/hubble/src/hubble.ts
+++ b/apps/hubble/src/hubble.ts
@@ -1006,6 +1006,18 @@ export class Hub implements HubInterface {
   /* -------------------------------------------------------------------------- */
 
   private async handleGossipMessage(gossipMessage: GossipMessage, source: PeerId, msgId: string): HubAsyncResult<void> {
+    let reportedAsInvalid = false;
+    if (gossipMessage.timestamp) {
+      // If message is older than seenTTL, we will try to merge it, but report it as invalid so it doesn't
+      // propogate across the network
+      const cutOffTime = getFarcasterTime().unwrapOr(0) - GOSSIP_SEEN_TTL;
+
+      if (gossipMessage.timestamp < cutOffTime) {
+        this.gossipNode.reportValid(msgId, peerIdFromString(source.toString()).toBytes(), false);
+        reportedAsInvalid = true;
+      }
+    }
+
     const peerIdResult = Result.fromThrowable(
       () => peerIdFromBytes(gossipMessage.peerId ?? new Uint8Array([])),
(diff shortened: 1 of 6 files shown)


apps/hubble/src/hubble.ts
```diff
@@ -1017,6 +1017,13 @@ export class Hub implements HubInterface {
     if (gossipMessage.message) {
       const message = gossipMessage.message;
 
+      // If message is older than seenTTL, we will try to merge it, but report it as invalid so it doesn't
+      // propogate across the network
+      const cutOffTime = getFarcasterTime().unwrapOr(0) - GOSSIP_SEEN_TTL;
+      if (message.data?.timestamp && message.data.timestamp < cutOffTime) {
```
sanjayprabhu, line 1023: I think this could have unintended consequences. Message timestamp is set by the clients when it's first created. Could be significant delays before it even reaches the hubs (e.g. backed queues backed up)


We should use a new field that's set to the current timestamp when the message is first gossip (after RPC submit)

adityapk00: Ok, added a new `timestamp` to the gossip.proto

apps/hubble/src/hubble.ts
```diff
@@ -1045,7 +1059,9 @@ export class Hub implements HubInterface {
           },
           "Received bad gossip message from peer",
         );
-        this.gossipNode.reportValid(msgId, peerIdFromString(source.toString()).toBytes(), false);
+        if (!reportedAsInvalid) {
```
sagar-a16z, line 1062: Should add a metric so we can track how many invalid messages were detected?

PR #1746, opened 2024-02-25
fix: update hello-world index.ts

PhatStraw:
## Motivation

Dev time should be kept minimum and this example overcomplicates the startup process. 

## Change Summary

Using a Private Key to Initiate the account rather than having to create a new wallet completely, or exposing a personal wallets mnemonic phrase.

We've also removed unused variables, which can also lead to confusion.

## Merge Checklist

- [x ] PR title adheres to the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) standard
- [ ] All [commits have been signed](https://github.com/farcasterxyz/hub-monorepo/blob/main/CONTRIBUTING.md#22-signing-commits)

## Additional Context

### DIFF
--- a/packages/hub-nodejs/examples/hello-world/index.ts
+++ b/packages/hub-nodejs/examples/hello-world/index.ts
@@ -18,7 +18,7 @@ import {
   ViemLocalEip712Signer,
   makeCastRemove,
 } from "@farcaster/hub-nodejs";
-import { mnemonicToAccount, toAccount } from "viem/accounts";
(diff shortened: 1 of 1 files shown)


sds:
Thanks for suggesting ways to improve our examples!

In this case, we're going to politely decline the change. Since this is example code, we're not optimizing for dev time, but instead are trying to show a "typical" example. Since the net difference is 3 lines we'll leave the example as-is. Thanks!

PR #1748, opened 2024-02-26
Move Reactions Store and RocksDB to rust

adityapk00:
## Motivation

Merging messages into the CRDT stores is slow, and to speed it up, we'll move the stores and the DB itself into rust. 

## Change Summary

- Move DB into Rust (instead of abstract-leveldown)
- Move the reactions store into Rust
- RocksDB is upgraded from 6.17 to 8.10
- Add statsd metrics to track merge latencies specific to rust vs nodejs

## Merge Checklist

## Additional Context
There's a README in `src/addon/` that describes the rust objects used

---

## PR-Codex overview
This PR focuses on moving the DB and Reactions store to Rust for performance improvement. 

### Detailed summary
- Added Rust modules for DB operations
- Updated storage configurations
- Improved logging functionality
- Implemented Rust functions for hashing
- Refactored DB iterators
- Added Rust client for statsd
- Updated Cargo.toml dependencies and build.rs
- Reorganized store modules

> The following files were skipped due to too many changes: `apps/hubble/package.json`, `apps/hubble/src/network/sync/trieNode.ts`, `apps/hubble/src/storage/jobs/revokeMessagesBySignerJob.ts`, `apps/hubble/src/storage/stores/castStore.test.ts`, `Dockerfile.hubble`, `apps/hubble/src/storage/db/hubState.ts`, `apps/hubble/src/storage/db/migrations/2.fnameproof.ts`, `apps/hubble/src/storage/stores/usernameProofStore.ts`, `apps/hubble/src/storage/stores/store.test.ts`, `apps/hubble/src/storage/stores/castStore.ts`, `apps/hubble/src/storage/db/nameRegistryEvent.ts`, `apps/hubble/src/storage/db/migrations/1.usernameproof.ts`, `apps/hubble/src/storage/jobs/pruneMessagesJob.ts`, `apps/hubble/src/storage/db/migrations/5.fnameSyncIds.ts`, `.github/workflows/ci.yml`, `apps/hubble/src/network/sync/syncEngine.ts`, `apps/hubble/src/addon/src/statsd.rs`, `apps/hubble/src/storage/stores/storageCache.ts`, `apps/hubble/src/storage/stores/linkStore.ts`, `apps/hubble/src/storage/stores/onChainEventStore.ts`, `apps/hubble/src/storage/db/onChainEvent.ts`, `apps/hubble/src/storage/stores/reactionStore.test.ts`, `apps/hubble/src/storage/engine/index.ts`, `apps/hubble/src/network/sync/merkleTrie.ts`, `apps/hubble/src/profile/profile.ts`, `apps/hubble/src/rpc/server.ts`, `apps/hubble/src/rustfunctions.test.ts`, `apps/hubble/src/addon/README.md`, `apps/hubble/src/addon/src/lib.rs`, `apps/hubble/src/addon/src/store/store_event_handler.rs`, `apps/hubble/src/storage/jobs/validateOrRevokeMessagesJob.ts`, `apps/hubble/src/storage/stores/storeEventHandler.test.ts`, `apps/hubble/src/addon/src/logger.rs`, `apps/hubble/src/hubble.ts`, `yarn.lock`, `apps/hubble/src/storage/db/message.ts`, `apps/hubble/src/storage/stores/verificationStore.ts`, `apps/hubble/src/addon/src/store/utils.rs`, `apps/hubble/src/storage/stores/storeEventHandler.ts`, `apps/hubble/src/storage/stores/store.ts`, `apps/hubble/src/addon/src/store/message.rs`, `apps/hubble/src/rustfunctions.ts`, `apps/hubble/src/storage/db/rocksdb.ts`, `apps/hubble/src/addon/src/store/reaction_store.rs`, `apps/hubble/src/storage/stores/reactionStore.ts`, `apps/hubble/src/addon/src/db/rocksdb.rs`, `apps/hubble/src/storage/db/rocksdb.test.ts`, `apps/hubble/src/addon/src/store/store.rs`, `apps/hubble/src/addon/Cargo.lock`, `apps/hubble/src/addon/src/proto/protobufs.rs`


apps/hubble/src/addon/src/proto/protobufs.rs
```diff
@@ -0,0 +1,2044 @@
+#[allow(clippy::derive_partial_eq_without_eq)]
```
sanjayprabhu, line 1: Is this auto generated? We should hook it up to `yarn protoc` if not, so it's always in sync with the other autogenerated files.

apps/hubble/src/addon/README.md
```diff
@@ -0,0 +1,43 @@
...
+Currently, the reaction store is hosted in Rust. It contains the code for merging CRDT messages.
+
+### Note about ownership
+All rust objects need to be "owned" by someone so that rust can manage its lifecycle. For rust objects like the `RocksDb` and the `ReactionStore`, we create them as `JsBox<Arc<T>>`. `JsBox` is like Rust's `Box`, except it is owned by the Javascript pointer that is returned. That is, 
+- The DB object is owned by the Javascript object that is returned
+- When the Javascript object goes out of scope and is gc'd, it is `drop()`-ed in Rust. 
+
+Since the Rust objects are `Arc<T>` inside a `JsBox`, we can clone them and keep them around in the rust code as we please, since the Javascript code will continue to own one `Arc<T>`, making sure that it lasts for the lifetime of the program.
+
+### Note about itearators
```
Wazzymandias, line 17:
```suggestion
### Note about iterators
```

apps/hubble/src/addon/README.md
```diff
@@ -0,0 +1,43 @@
...
+Rust code needs to be memory-safe, which means that we can't pass around iterators like we do in Javascript. This is because the `iterator` reference is valid for only as long as the `db` is valid, and the reference is dropped right after the iterator is finished.
+
+This specifically means that we need to use iterators as callbacks. The way the iterators are set up is:
+- Call the `forEachIteartor` method with your callback (Either in JS or Rust)
```
Wazzymandias, line 21:
```suggestion
- Call the `forEachIterator` method with your callback (Either in JS or Rust)
```

apps/hubble/src/addon/src/proto/protobufs.rs
adityapk00: yes, auto generated in `build.rs`. I'll look into `yarn protoc` generating rust files

apps/hubble/src/addon/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 license = "ISC"
 edition = "2018"
 exclude = ["index.node"]
```
Wazzymandias, line 6: it's a little opaque on what this index.node file is - could we add small note in readme about how we're generating this binary? 

will be easier to grok once cargo stable finally allows passing in an output directory

apps/hubble/src/addon/src/db/rocksdb.rs
```diff
@@ -0,0 +1,730 @@
...
+impl RocksDB {
+    pub fn new(path: &str) -> Result<RocksDB, HubError> {
+        Ok(RocksDB {
+            db: RwLock::new(None),
+            path: path.to_string(),
+        })
+    }
+
+    pub fn open(&self) -> Result<(), HubError> {
+        let mut db_lock = self.db.write().unwrap();
```
Wazzymandias, line 66: I think we should avoid lock poisoning panic here - would prefer if we could match on the error and bubble it up as HubError

Wazzymandias: we could just add a todo for now to reduce scope - i could work on this in followup PR

apps/hubble/src/storage/stores/reactionStore.bench.test.ts
```diff
@@ -0,0 +1,107 @@
+import { Factories, HubError, ReactionAddMessage, ReactionRemoveMessage, ReactionType } from "@farcaster/hub-nodejs";
+import StoreEventHandler from "./storeEventHandler.js";
+import { jestRocksDB } from "../db/jestUtils.js";
+import ReactionStore from "./reactionStore.js";
+import { Result, ResultAsync } from "neverthrow";
+
+xdescribe("ReactionStoreProxy", () => {
```
sanjayprabhu, line 7: Do we need this?

apps/hubble/src/storage/stores/reactionStore.ts
```diff
@@ -1,259 +1,247 @@
...
+      throw new HubError("bad_request.prunable", "message would be pruned");
+    }
 
-  override _isAddType = isReactionAddMessage;
-  override _isRemoveType = isReactionRemoveMessage;
-  override _addMessageType = MessageType.REACTION_ADD;
-  override _removeMessageType = MessageType.REACTION_REMOVE;
+    // Encode the message to bytes
+    const messageBytes = Message.encode(message).finish();
+    const result = await ResultAsync.fromPromise(merge(this.rustReactionStore, messageBytes), rustErrorToHubError);
```
sanjayprabhu, line 87: Might be good to somehow make it easy to identify `merge` is calling into rust. Either a prefix (`rsMerge`?), or put it into a namespace (`rustAPI.merge`?).

apps/hubble/src/storage/stores/reactionStore.ts
```diff
@@ -1,259 +1,247 @@
...
+    } else if (prunableResult.value) {
+      throw new HubError("bad_request.prunable", "message would be pruned");
+    }
 
-  override _isAddType = isReactionAddMessage;
-  override _isRemoveType = isReactionRemoveMessage;
-  override _addMessageType = MessageType.REACTION_ADD;
-  override _removeMessageType = MessageType.REACTION_REMOVE;
+    // Encode the message to bytes
+    const messageBytes = Message.encode(message).finish();
```
sanjayprabhu, line 86: `dataBytes` only messages work fine?

apps/hubble/src/storage/stores/storeEventHandler.ts
```diff
@@ -387,6 +386,12 @@ class StoreEventHandler extends TypedEmitter<StoreEvents> {
       });
   }
 
+  async processRustCommitedTransaction(event: HubEvent): HubAsyncResult<void> {
```
sanjayprabhu, line 389: Hmm, we don't need the commit lock anymore? I guess once we move event id generation into rust, it's probably ok? Is there anything else we need to worry about?

apps/hubble/src/storage/stores/reactionStore.ts
```diff
@@ -1,259 +1,247 @@
...
+    return undefined;
   }
 
-  override makeRemoveKey(msg: ReactionRemoveMessage) {
-    return makeReactionRemovesKey(
-      msg.data.fid,
-      msg.data.reactionBody.type,
-      msg.data.reactionBody.targetUrl || msg.data.reactionBody.targetCastId,
+  async merge(message: Message): Promise<number> {
+    const prunableResult = await this._eventHandler.isPrunable(
```
sanjayprabhu, line 72: Are we losing any accidental consistency guarantees now that the merge lock doesn't exist? e.g. looking up conflicts involve multiple db reads, would it be possible now for race conditions between two messages that might conflict with each other?

sanjayprabhu (commented):
Lefts some comments on the JS code, working on the rust code.

apps/hubble/src/addon/src/proto/protobufs.rs
adityapk00: I set it up so that the protos are compiled if they change, so we don't have to do `yarn protoc` for the rust code. (i.e., build.rs is checked on compile)

apps/hubble/src/addon/src/db/rocksdb.rs
adityapk00: The way neon works, panics in rust code are thrown as normal JS errors, which can then be caught in the JS code normally. Additionally, a panic in rust only unwinds the currently executing rust call (i.e., it doesn't crash the program), and it can be handled in the JS code just like if a normal exception had been thrown, and then the program can continue and call into rust again. 

So we can use unwrap() everywhere, and it will get caught as a normal error in JS

apps/hubble/src/storage/stores/reactionStore.bench.test.ts
adityapk00: I'll remove it before I merge, just using it for some testing

apps/hubble/src/storage/stores/storeEventHandler.ts
adityapk00: The commit lock is now in Rust, where the idGeneration code is locked by mutex

apps/hubble/src/storage/stores/reactionStore.ts
adityapk00: Because the rust code is not async (i.e., it is sync), it can't be interrupted while merging, effectively making the merges happen one-after-the-other, so there won't be race conditions. Even if we do multithreading in the future, we'll have one thread per store, so still each store will be conflict free. 

Additionally, there is an `fid_lock` in the rust merge code, even though we strictly don't need it for safety in case we do some funky optimisations in the future. 

The one thing that's different is the isPrunable is checked outside the lock, which might possibly cause some messages to be merged which would have failed, but they'll simply get pruned when the prune job runs every hour.

apps/hubble/src/storage/stores/reactionStore.ts
adityapk00: yes, added test to verify

apps/hubble/src/storage/stores/reactionStore.ts
adityapk00: Good idea. Changed everything in `rustFunctions.ts` to `rsMethodName`

apps/hubble/src/storage/stores/storeEventHandler.ts
```diff
@@ -388,7 +366,7 @@ class StoreEventHandler extends TypedEmitter<StoreEvents> {
 
   async processRustCommitedTransaction(event: HubEvent): HubAsyncResult<void> {
     void this._storageCache.processEvent(event);
-    void this.broadcastEvent(event);
+    this.broadcastEvent(event);
```
sanjayprabhu, line 354: On second thought maybe this `void` means that the emitted events are handled async? I'm not exactly sure how this works. Don't want this to cause us to block merges because we're writing to a network stream in `BufferedStreamWriter`

apps/hubble/src/rustfunctions.ts
```diff
@@ -29,21 +29,21 @@ export class RustMessagesPage {
 }
 
 // Use this function in TypeScript to call the rust code.
-export function nativeBlake3Hash20(data: Uint8Array): Uint8Array {
+export function rsBlake3Hash20(data: Uint8Array): Uint8Array {
```
sanjayprabhu, line 32: Nice, thank you.

apps/hubble/src/addon/src/store/store.rs
```diff
@@ -624,6 +628,11 @@ impl Store {
                     }
                 };
 
+                if message.data.is_none() {
+                    // This shouldn't happen, but if it does, skip it
+                    return Ok(false); // Continue the iteration
```
sanjayprabhu, line 620: Should we log? 

And actually how would logging work? I don't want this PR to keep getting bigger, so if it's not easy, we can skip for now.

apps/hubble/src/hubble.ts
```diff
@@ -786,18 +786,33 @@ export class Hub implements HubInterface {
           if (dbFiles.isErr() || dbFiles.value.length === 0) {
             log.info({ dbLocation }, "DB is empty, fetching snapshot from S3");
 
-            const response = await axios.get(`https://download.farcaster.xyz/${this.getSnapshotFolder()}/latest.json`);
-            const { key } = response.data;
+            let prevVersion = 0;
+            let latestSnapshotKey;
+            do {
+              const response = await axios.get(
+                `https://download.farcaster.xyz/${this.getSnapshotFolder(prevVersion)}/latest.json`,
```
sanjayprabhu, line 793: Nice

sanjayprabhu (approved):
Need to spend some more time with the rust code, but generally looks good. The plan is not to merge and run on one of our prod hubs for a while, right?

We should keep track of what needs to happen in a follow up PR. Off the top of my head:

 - logging
 - statsd
 - rust tests

adityapk00:
Yes, the plan for deploying this is:
- Deploy to hoyt from `rustcrdt` (not merge to main)
- Track stats on Prod for a while
- If something goes wrong, `rm -rf` hoyt's DB and have it rebuild from snapshot

Things still left in the PR (copying from Sanjay's comment above):
- Logging - stdout is already connected in Rust, so we just need to add the logging classes and log in the datadog (JSON) format
- statsd -There's a nice wrapper just like we have for nodejs. We just need to wire it up
- Start adding rust specific tests
- Throttle pruning

One thing is after this PR, the bottleneck is adding to the sync trie. I noticed that now that pruning is in Rust, it's faster than the merkle trie ops, so the pruning creates a queue in the merkle trie lock. Fine for now, but maybe we should throttle pruning

adityapk00:
Current performance is extremely good. The rust store merges ~3x to 5x faster. Additionally, it is not susceptible to spikes in latency (probably because of GC)

image

PR #1770, opened 2024-03-04
fix: replicator Unicode error

Wazzymandias:
## Motivation

- Solana verifications were causing unicode errors in postgres: `unsupported Unicode escape sequence`
- This is because we store certain attributes as JSONB in postgres
- Postgres does not allow null bytes in JSONB
- Solana addresses, claim signatures, and block hashes, due to their cryptographic nature, can be considered arbitrary binary data which can yield byte arrays that have zero values in them
- Ethereum verifications were not getting this error because their values are hex encoded 

## Change Summary

- Convert Solana addresses and block hashes to base58 encoding prior to replicator insertion
- Convert both Ethereum and Solana claim signatures to hex encoding prior to replicator insertion
- Add TODO to create separate replicator process for Solana

## Merge Checklist

## Additional Context

---

## PR-Codex overview
The focus of this PR is to fix a unicode error in the replicator and enforce hex signatures for verification messages.

### Detailed summary
- Fixed unicode error in replicator
- Added constraint for hex verification message claim signatures
- Updated Dockerfile versions
- Improved error handling in worker.ts
- Added utility functions for hex encoding
- Updated message processing logic for verification messages

> The following files were skipped due to too many changes: `apps/replicator/src/processors/index.ts`


apps/replicator/docker-compose.yml
```diff
@@ -3,6 +3,9 @@ version: '3.9'
 services:
   replicator:
     image: farcasterxyz/replicator:latest
+    build:
+        context: ../../
+        dockerfile: Dockerfile.replicator
```
sds, line 8: Does `build` conflict with `image`? Or am I misremembering?

apps/replicator/src/processors/index.ts
```diff
@@ -155,13 +155,19 @@ export async function processMessage(
...
+      case MessageType.VERIFICATION_ADD_ETH_ADDRESS: {
         // TODO: add support for multi-protoocl verification
         if (!isVerificationAddAddressMessage(message))
           throw new AssertionError(`Invalid VerificationAddEthAddressMessage: ${message}`);
         log.debug(`Processing VerificationAddEthAddressMessage ${hash} (fid ${fid})`, { fid, hash });
+        const claimSignature = message.data.verificationAddAddressBody.claimSignature;
+        message.data.verificationAddAddressBody.claimSignature =
+          claimSignature.length > 0 && claimSignature[claimSignature.length - 1] === 0
+            ? claimSignature.slice(0, claimSignature.length - 1)
+            : claimSignature;
```
sds, line 167: This is an example where Ruby's [`chomp`](https://ruby-doc.org/3.3.0/String.html#method-i-chomp) would be a nice addition to the JS standard library.

apps/replicator/src/processors/index.ts
```diff
@@ -155,13 +155,19 @@ export async function processMessage(
         log.debug(`Processing LinkRemoveMessage ${hash} (fid ${fid})`, { fid, hash });
         await processLinkRemove(message, operation, trx);
         break;
-      case MessageType.VERIFICATION_ADD_ETH_ADDRESS:
+      case MessageType.VERIFICATION_ADD_ETH_ADDRESS: {
```
sanjayprabhu, line 158: stray parens?

sds (approved):
Do we understand why Solana adds a null byte to the signature? Is there a good reason for such behavior?

apps/replicator/src/processors/index.ts
Wazzymandias: nah linter complained

I added new variable which it said could leak in the rest of the switch statement so it made me wrap in braces

apps/replicator/docker-compose.yml
Wazzymandias: AFAICT if I do `docker compose up` it'll use the image but only use build when I do `docker compose up --build`

sds: Neat!

Wazzymandias:
> Do we understand why Solana adds a null byte to the signature? Is there a good reason for such behavior?

AFAICT it seems to be standard for some reason for solana signatures. I thought it would be a null terminating byte, but there are null bytes at seemingly random positions of the signature:

```
 index 16 has null byte 64

      at hasNull (index.test.ts:56:17)

  console.log
    verificationAddAddressBody.claimSignature has null byte
 ```
    


    `    index 46 has null byte 63`
    
    
    
    We're inserting the values directly into the database. Neynar isn't seeing this issue because they iterate through the messages and convert the bytes to hex, which resolves this error

File: apps/replicator/src/processors/index.ts [hunk 4]
```diff
@@ -155,13 +159,61 @@ export async function processMessage(
...
+            if (addressBase58.isErr()) {
+              throw new AssertionError(`Invalid address: ${addressBase58.error}`);
+            }
+
+            const addressBytes = base58ToBytes(addressBase58.value);
+            if (addressBytes.isErr()) {
+              throw new AssertionError(`Invalid address: ${addressBytes.error}`);
+            }
+
+            address = addressBytes.value;
```
sanjayprabhu, line 196: This gets hex encoded again?

I think it'll be simpler to add a `processVerificationAddSolAddress`

Wazzymandias: > This gets hex encoded again?

No, it's only for solana address and gets base58 encoded

sanjayprabhu (approved):
Thank you

PR #1827, opened 2024-03-19
perf: Throttle prune job

adityapk00:
## Motivation

The prune job is running too fast, causing spikes in merge latency. This PR throttles the job

## Change Summary

  // We prune at the rate of 1000 fids per second. If we are running ahead of schedule, we sleep to catch up
  // We do this so that:
  // 1. Each fid is pruned at the same time at every hub, reducing thrash
  // 2. Don't overload the DB or Sync Trie, causing spikes in latency

## Merge Checklist

---

## PR-Codex overview
The focus of this PR is to improve the performance of the prune job in the `@farcaster/hubble` module.

### Detailed summary
- Throttled the prune job to avoid overloading the system
- Removed unnecessary parameter in `PruneMessagesJobScheduler` constructor
- Added logic to sleep and catch up if pruning is ahead of schedule

### DIFF
--- a/apps/hubble/src/storage/jobs/pruneMessagesJob.ts
+++ b/apps/hubble/src/storage/jobs/pruneMessagesJob.ts
@@ -91,12 +91,12 @@ export class PruneMessagesJobScheduler {
         // 2. Don't overload the DB or Sync Trie, causing spikes in latency
         if (fid % 100 === 0) {
           // See if we are running ahead of schedule
-          const scheduledTime = TIME_SCHEDULED_PER_FID_MS * fid;
-          const elapsedTime = Date.now() - start;
-          if (scheduledTime > elapsedTime) {
-            const sleepTime = scheduledTime - elapsedTime;
+          const allotedTimeMs = TIME_SCHEDULED_PER_FID_MS * fid;
+          const elapsedTimeMs = Date.now() - start;
+          if (allotedTimeMs > elapsedTimeMs) {
+            const sleepTimeMs = allotedTimeMs - elapsedTimeMs;
             // Sleep for the remaining time
-            await sleep(sleepTime);
+            await sleep(sleepTimeMs);
           }
         }
       }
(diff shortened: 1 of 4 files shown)


apps/hubble/src/storage/jobs/pruneMessagesJob.ts
```diff
@@ -82,13 +84,21 @@ export class PruneMessagesJobScheduler {
...
-      // Sleep for a bit avoid overloading the Merkle Trie
-      const syncTrieQSize = this._getSyncTrieQSizeFn();
-      if (syncTrieQSize > 10_000) {
-        log.info({ syncTrieQSize }, "sync trie Q is large, sleeping for 30s");
-        await sleep(30_000);
+        // We prune at the rate of 1000 fids per second. If we are running ahead of schedule, we sleep to catch up
+        // We do this so that:
+        // 1. Each fid is pruned at the same time at every hub, reducing thrash
+        // 2. Don't overload the DB or Sync Trie, causing spikes in latency
+        if (fid % 100 === 0) {
```
Wazzymandias, line 92: question: This _may_ sleep every 100 fids - is that intended?

adityapk00: Yes, it should pause every 100 fids to throttle

Wazzymandias (approved):

apps/hubble/src/storage/jobs/pruneMessagesJob.ts
```diff
@@ -8,6 +8,10 @@ import { sleep } from "../../utils/crypto.js";
 
 export const DEFAULT_PRUNE_MESSAGES_JOB_CRON = "0 */2 * * *"; // Every two hours
 
+// How much time to allocate to pruning each fid.
+// 1000 fids per second = 1 fid per ms. 500k fids will take under 10 minutes
+const TIME_SCHEDULED_PER_FID_MS = 1;
```
sanjayprabhu, line 13: Do you know what's the current rate? Wonder if even this is too aggressive, especially for lower spec'd machines. But we can tune as we see the impact

Wazzymandias:
It could be nice to have to collect statsd metric on # processed FIDs vs total # fids to make sure we don't fall behind on registration spikes. But that probably isn't necessary

sanjayprabhu (approved):

apps/hubble/src/storage/jobs/pruneMessagesJob.ts
adityapk00: Currently (on the prod hubs), the whole thing takes ~6 minutes. This PR will spread it out over ~9 minutes, with breaks every 100 fids

PR #1833, opened 2024-03-20
feat(hubble): catchup sync with s3 snapshot

Wazzymandias:
## Motivation

- If there is a large gap in message count for a hub compared to the latest snapshot, it is faster to sync from s3 snapshot which is publicly available online
- **NOTE**: Catchup sync using snapshot WILL RESET THE DATABASE

## Change Summary

- Add `--catchup-sync-with-snapshot` and `--catchup-sync-snapshot-message-limit <number>` 
  - By default, `--catchup-sync-with-snapshot` is disabled, and will be used if message difference is greater than `3_000_000` 
- Add logic for using s3 snapshot for "catch up" sync. Snapshot metadata is used to determine whether snapshot should be used. If metadata JSON is missing attribute for message count, an error is returned 
- Fix small bug where HTTP API was reporting incorrect sync status
- **NOTE**: Catchup sync using snapshot **WILL RESET THE DATABASE**

## Merge Checklist

## Additional Context

Confirmed upload and download working:

```
{"level":30,"time":1710988900433,"pid":18800,"hostname":"mbp.local","component":"Hub","downloadedSize":1073751179,"totalSize":24114175697,"msg":"Downloading snapshot..."}
{"level":30,"time":1710988921461,"pid":18800,"hostname":"mbp.local","component":"Hub","downloadedSize":2147499571,"totalSize":24114175697,"msg":"Downloading snapshot..."}
{"level":30,"time":1710988941068,"pid":18800,"hostname":"mbp.local","component":"Hub","downloadedSize":3221242283,"totalSize":24114175697,"msg":"Downloading snapshot..."}
```

This is using test bucket with transfer acceleration enabled.

**NOTE:** If the s3-bucket-name isn't valid DNS endpoint, the snapshot sync fails. This was issue before, but was not problem because we hardcoded download.farcaster.xyz:
```
{"level":30,"time":1710988763240,"pid":18380,"hostname":"mbp.local","component":"Hub","db_location":".rocks/rocks.hub._default","file_count":1,"overwrite":true,"msg":"DB is empty or overwrite is true, fetching snapshot from S3"}
{"level":50,"time":1710988763292,"pid":18380,"hostname":"mbp.local","component":"Hub","error":{"message":"getaddrinfo ENOTFOUND wasif-test3","name":"Error","stack":"Error: getaddrinfo ENOTFOUND wasif-test3
```

For now, it shouldn't be concern since default is download.farcaster.xyz which is public and available to everyone

---

## PR-Codex overview
The focus of this PR is to enhance catchup sync functionality and snapshot handling in the Hubble application.

### Detailed summary
- Improved catchup sync with snapshot feature
- Added option to enable catchup sync using S3 snapshot
- Added message limit trigger for catchup sync
- Updated default configuration settings for catchup sync
- Refactored code for snapshot handling and sync process

> The following files were skipped due to too many changes: `apps/hubble/src/hubble.ts`

---

## PR-Codex overview
This PR introduces support for catch up sync using S3 snapshot in the `Hubble` app. 

### Detailed summary
- Added support for catch up sync with S3 snapshot
- Updated CLI options for catch up sync with snapshot
- Implemented catch up sync logic with snapshot
- Defined catch up sync snapshot message limit
- Modified default configuration for catch up sync snapshot message limit

> The following files were skipped due to too many changes: `apps/hubble/src/hubble.ts`

---

## PR-Codex overview
The focus of this PR is to introduce support for using S3 snapshots for "catch up" sync in the Hubble app.

### Detailed summary
- Added support for catch up sync using S3 snapshot
- Modified CLI options for enabling catch up sync with snapshot
- Implemented logic for catch up sync with snapshot based on message count difference
- Updated default configuration for catch up sync with snapshot
- Implemented functions for handling snapshot metadata and uploading to S3

> The following files were skipped due to too many changes: `apps/hubble/src/hubble.ts`


apps/hubble/src/cli.ts
```diff
@@ -129,6 +129,15 @@ app
   .option("--enable-snapshot-to-s3", "Enable daily snapshots to be uploaded to S3. (default: disabled)")
   .option("--s3-snapshot-bucket <bucket>", "The S3 bucket to upload snapshots to")
   .option("--disable-snapshot-sync", "Disable syncing from snapshots. (default: enabled)")
+  .option("--catchup-sync-with-snapshot", "Enable catchup sync with snapshot. (default: enabled)")
+  .option(
```
sanjayprabhu, line 133: Maybe we should flip the default? Only enable in the docker compose file for hubble.sh? Also we should allow enabling/disabling via the `.env` file for `.hubble.sh`

apps/hubble/src/hubble.ts
```diff
@@ -618,8 +648,13 @@ export class Hub implements HubInterface {
     }
 
     if (this.options.resetDB === true) {
-      log.info("clearing rocksdb");
-      await this.rocksDB.clear();
+      // Sync using catchup sync clears the database, so we don't need to do it again.
+      if (this.options.catchupSyncWithSnapshot) {
+        log.info("skipping db reset as catchup sync with snapshot is enabled, which already clears the db");
```
sanjayprabhu, line 653: Why do we need to care about the snapshot within reset db? If the user passes in --resetDB it should always reset the db right?

apps/hubble/src/hubble.ts
```diff
@@ -800,7 +835,108 @@ export class Hub implements HubInterface {
...
+    if (snapshotMetadata.numMessages) {
+      delta = snapshotMetadata.numMessages - currentItemCount;
+      if (delta > limit) {
+        log.info({ delta, limit }, "catchup sync using snapshot");
+        shouldCatchupSync = true;
+      }
+    } else {
+      // Older snapshot metadata JSON may not contain database statistics (e.g. numMessages).
+      // As a result, we perform conservative calculations to determine if
+      // catchup sync using snapshot is necessary
```
sanjayprabhu, line 878: I think we don't need to handle this case since your other PR is already merged. Just need to make sure keats is running that code and has published a succesful snapshot before merging this. Saves some complexity

apps/hubble/src/hubble.ts
```diff
@@ -800,7 +835,108 @@ export class Hub implements HubInterface {
...
+        );
+        if (delta > limit) {
+          log.info({ delta, limit }, "catchup sync using snapshot");
+          shouldCatchupSync = true;
+        }
+      }
+    }
+
+    if (shouldCatchupSync) {
+      const SHUTDOWN_GRACE_PERIOD_MS = 10 * 1000; // 10 seconds
```
sanjayprabhu, line 904: Maybe do 30 seconds?

apps/hubble/src/network/sync/merkleTrie.ts
```diff
@@ -207,7 +208,11 @@ class MerkleTrie {
   public static async numItems(trie: MerkleTrie): HubAsyncResult<number> {
```
sanjayprabhu, line 208: Could you add a test for this?

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -89,12 +89,19 @@ export class StorageCache {
     this._earliestTsHashes = new Map();
 
     // Start prepopulating the cache in the background
-    this.prepopulateMessageCounts();
+    if (this._db.status === "open") {
+      this.prepopulateMessageCounts();
+    }
```
sanjayprabhu, line 94: Should we fail if it's not open? Is it possible to get here when it's not open?

apps/hubble/src/hubble.ts
Wazzymandias: we're invoking catchup sync before reset db is checked - seems wasteful to fetch the snapshot and untar it only to nuke the db again

I can revert if we want to leave as is

apps/hubble/src/storage/stores/storageCache.ts
Wazzymandias: > Is it possible to get here when it's not open

yeah, i ran into this in some nondeterministic cases

Wazzymandias: > Should we fail if it's not open?

Yeah, I can add that

apps/hubble/src/hubble.ts
sanjayprabhu: Ah that makes sense, but `catchUpSyncWithSnapshot` cli flag doesn't guarantee that we actually replace db right? We could still decide the message counts are close enough. So maybe this check should be about whether we actually did replace the db

apps/hubble/src/cli.ts
Wazzymandias: toggling between `--catchup` and `--no-catchup` is a bit different than the other flags we exposed, but updated to support this

apps/hubble/src/hubble.ts
Wazzymandias: good point, updated to actually check whether catchup sync replaced the db and only then do we skip

apps/hubble/src/hubble.ts
Wazzymandias: updated

apps/hubble/src/network/sync/merkleTrie.ts
Wazzymandias: tl;dr - No, not easily


There's a lot of tightly coupled code in the merkle trie worker. When NODE_ENV is test, the merkle trie implementation doesn't instantiate a database. 

Getting around this adds quite a bit of complexity from what I can tell. Maybe there is a simple approach but I didn't see it. 

I think the lesson here is we've dramatically overextended beyond the capabilities of Typescript/Javascript. We should just move to Rust and then testing stuff like this gets dramatically easier.

apps/hubble/src/storage/stores/storageCache.ts
Wazzymandias: added

apps/hubble/src/hubble.ts
Wazzymandias: removed this case and return error instead

apps/hubble/src/cli.ts
```diff
@@ -129,11 +129,11 @@ app
   .option("--enable-snapshot-to-s3", "Enable daily snapshots to be uploaded to S3. (default: disabled)")
   .option("--s3-snapshot-bucket <bucket>", "The S3 bucket to upload snapshots to")
   .option("--disable-snapshot-sync", "Disable syncing from snapshots. (default: enabled)")
-  .option("--catchup-sync-with-snapshot", "Enable catchup sync with snapshot. (default: enabled)")
   .option(
     "--no-catchup-sync-with-snapshot",
```
sanjayprabhu, line 133: Is this used anymore?

.changeset/violet-peas-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"@farcaster/hubble": patch
+---
+
+feat(hubble): Add support for using S3 snapshot for "catch up" sync. Snapshot metadata is used to determine whether snapshot should be used. If metadata JSON is missing attribute for message count, conservative heuristics are used to prefer using snapshot.
```
sanjayprabhu, line 5: Would make this shorter, so it fits in one line

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - name: Run Hubble
         shell: bash
-        run: docker run --name hub --detach -p2282:2282 -p2283:2283 farcasterxyz/hubble:test sh -c 'node build/cli.js identity create && HUBBLE_ARGS="start --rpc-port 2283 --ip 0.0.0.0 --gossip-port 2282 --eth-mainnet-rpc-url https://eth-mainnet.g.alchemy.com/v2/8cz__IXnQ5FK_GNYDlfooLzYhBAW7ta0 --l2-rpc-url https://opt-mainnet.g.alchemy.com/v2/3xWX-cWV-an3IPXmVCRXX51PpQzc-8iJ --network 3 --allowed-peers none" npx pm2-runtime start pm2.config.cjs'
+        run: docker run --name hub --detach -p2282:2282 -p2283:2283 farcasterxyz/hubble:test sh -c 'node build/cli.js identity create && HUBBLE_ARGS="start --rpc-port 2283 --ip 0.0.0.0 --gossip-port 2282 --eth-mainnet-rpc-url https://eth-mainnet.g.alchemy.com/v2/8cz__IXnQ5FK_GNYDlfooLzYhBAW7ta0 --l2-rpc-url https://opt-mainnet.g.alchemy.com/v2/3xWX-cWV-an3IPXmVCRXX51PpQzc-8iJ --network 3 --allowed-peers none --no-catchup-sync-with-snapshot" npx pm2-runtime start pm2.config.cjs'
```
sanjayprabhu, line 39: Don't need this any more

sanjayprabhu (approved):
Looks good to me, we should do a patch release once this is out. But before that, provide a heads up to neynar and airstack and anyone else running a replicator that they should keep this disabled for their replicator hub (or disabled in general for all hubs, their hubs shouldn't be falling behind in the first place.)

apps/hubble/pm2.config.cjs
```diff
@@ -17,6 +17,9 @@ module.exports = {
       // HACK: Allows us to pass arguments in a way recognized by the CLI flag processing library we use
       // This is only necessary when running via PM2.
       args: process.env.HUBBLE_ARGS,
+      env: {
+        "CATCHUP_SYNC_WITH_SNAPSHOT": (process.env.CATCHUP_SYNC_WITH_SNAPSHOT === "true").toString(),
```
sanjayprabhu, line 21: Actually, we should default to true here if it's not present.

apps/hubble/src/cli.ts
Wazzymandias: Commander is a confusing package. 

You can't do something like `--catchup-sync-with-snapshot false` 

This is the closest way to do it. 

TL;DR - we need both to toggle boolean flags using commander package

apps/hubble/pm2.config.cjs
Wazzymandias: I thought we wanted it to be disabled by default? Why default to true?

sanjayprabhu: Disabled for our hubs, and hubs that are not using the install script. The hubs that are using the install script, we should enable this feature.

.changeset/violet-peas-relax.md
Wazzymandias: updated

sanjayprabhu (approved):
If it works with docker, good to merge 👍

.github/workflows/ci.yml
Wazzymandias: removed

apps/hubble/pm2.config.cjs
Wazzymandias: updated approach

- docker compose will default catchup sync to true if it's not set
- user can set catchup sync through `.env` for docker compose
- if CLI flag for catchup sync is not set, we will default to false (the flag will be undefined if unset)

PR #1844, opened 2024-03-23
fix: handle edge cases of rocksdb instantiation

Wazzymandias:
## Motivation

- There were certain edge cases in instantiation that could cause snapshots to be uploaded with 0 messages
  - If snapshot flag is enabled but zero messages are found, we now return an error
  - path checks for trie db uses normalized paths and properly checks in cases where there's prefix of `.rocks`
  - If the trie DB directory exists, but hasn't loaded any messages yet, we catch the error in getting root node and return 0

## Change Summary

- If snapshot flag is enabled but zero messages are found, we now return an error
- path checks for trie db uses normalized paths and properly checks in cases where there's prefix of `.rocks`
- If the trie DB directory exists, but hasn't loaded any messages yet, we catch the error in getting root node and return 0

## Merge Checklist

## Additional Context

---

## PR-Codex overview
The focus of this PR is to enhance the `hubble` app by handling edge cases of rocksdb instantiation for snapshot uploads.

### Detailed summary
- Handle edge cases of rocksdb instantiation for snapshot uploads in `hubble`
- Synchronously fetch the number of elements in the trie DB
- Throw an error if message count is not obtained for snapshot upload
- Support catch up sync with snapshot only on mainnet
- Improve logging for catchup sync using snapshot
- Refactor trie database instantiation for consistency and error handling

### DIFF
--- a/apps/hubble/src/network/sync/merkleTrie.ts
+++ b/apps/hubble/src/network/sync/merkleTrie.ts
@@ -14,7 +14,7 @@ import {
 import { logger } from "../../utils/logger.js";
 import { getStatsdInitialization } from "../../utils/statsd.js";
 import { messageDecode } from "../../storage/db/message.js";
-import path from "path";
+import path, { dirname } from "path";
 import fs from "fs";
 
 /**
@@ -208,8 +208,25 @@ class MerkleTrie {
   public static async numItems(trie: MerkleTrie): HubAsyncResult<number> {
     // MerkleTrie has rocksdb instance, however the merkle trie worker
     // uses a separate instance under trieDb prefix which needs to be used here instead.
-    const location = `${path.basename(trie._db.location)}/${TrieDBPathPrefix}`;
-    if (!fs.existsSync(path.basename(`.rocks/${location}`))) {
+    // The trie database is instantiated with new Rocksdb, which will prefix an input path with ".rocks"
+    const fullPath = `${trie._db.location}/${TrieDBPathPrefix}`;
+    const normalizedPath = path.normalize(fullPath);
+    const parts = normalizedPath.split(path.sep);
+    // Remove the first directory. Note that the first element might be empty
+    // if the path starts with a separator, indicating it's an absolute path.
+    // In such a case, remove the second element instead.
+    if (parts[0] === "") {
+      parts.splice(1, 1); // Remove the second element for absolute paths
+    } else {
+      parts.splice(0, 1); // Remove the first element for relative paths
+    }
+
+    // Location will be the path of the trie database without the first directory, since RocksDB creates prefix
+    // of ".rocks" for every input path.
+    const location = parts.join(path.sep);
+
+    // We need to check if the full directory path exists, otherwise instantiating RocksDB will throw an error
+    if (!fs.existsSync(path.dirname(path.resolve(normalizedPath)))) {
       return ok(0);
     }
     const db = new RocksDB(location);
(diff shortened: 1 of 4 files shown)


apps/hubble/src/network/sync/merkleTrie.ts
```diff
@@ -208,8 +208,28 @@ class MerkleTrie {
   public static async numItems(trie: MerkleTrie): HubAsyncResult<number> {
     // MerkleTrie has rocksdb instance, however the merkle trie worker
     // uses a separate instance under trieDb prefix which needs to be used here instead.
-    const location = `${path.basename(trie._db.location)}/${TrieDBPathPrefix}`;
-    if (!fs.existsSync(path.basename(`.rocks/${location}`))) {
+    // The trie database is instantiated with new Rocksdb, which will prefix an input path with ".rocks"
+    const fullPath = `${trie._db.location}/${TrieDBPathPrefix}`;
```
sanjayprabhu, line 212: We should use path.join here

apps/hubble/src/network/sync/merkleTrie.ts
```diff
@@ -208,8 +208,28 @@ class MerkleTrie {
...
+    const normalizedPath = path.normalize(fullPath);
+    const parts = normalizedPath.split(path.sep);
+    // Remove the first directory. Note that the first element might be empty
+    // if the path starts with a separator, indicating it's an absolute path.
+    // In such a case, remove the second element instead.
+    if (parts[0] === "") {
+      parts.splice(1, 1); // Remove the second element for absolute paths
+    } else {
+      parts.splice(0, 1); // Remove the first element for relative paths
+    }
```
sanjayprabhu, line 222: Why do we need to handle all these cases only here? i.e. why aren't we hitting them when we open the trie db normally?

Wazzymandias: We aren't accessing using the merkle trie worker - most of the codebase just makes calls to the worker which has reference to trie db directly

apps/hubble/src/network/sync/merkleTrie.ts
Wazzymandias: done

apps/hubble/src/network/sync/merkleTrie.ts
sanjayprabhu: Right, but the worker is just doing `this._trieDb = new RocksDB(`${path.basename(dbPath)}/${TrieDBPathPrefix}`);` can't we do the same here?

Wazzymandias: this method is called on startup, before anything gets initialized 

so 
```
new RocksDB(`${path.basename(dbPath)}/${TrieDBPathPrefix}`);
```

Ends up throwing error because the parent directories may not exist.

Wazzymandias: We could try to wrap the call, i.e. using `Result.fromThrowable(new RocksDB)` but for debugging it was helpful for me to distinguish between path not existing and DB not being available to open (because of lock contention, corruption, etc) 

But I am open to simplifying it

Wazzymandias: the trie db is prefixed with `.rocks` - so we need to do some path manipulation to avoid `.rocks/.rocks` instantiation 

I've simplified the logic but at a minimum we need that to avoid random errors AFAICT

PR #1864, opened 2024-04-04
fix: Throttle storageCache population

adityapk00:
## Motivation

When a hub starts up, it attempts to pre-populate the storage cache. This is slow, since it has to basically scan every message. 

## Change Summary

- Slow down the pre-population
- Don't do too many scans in parallel.

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR focuses on fixing storage cache prepopulation throttling and adding memory usage logging with optional heap dumps. 

### Detailed summary
- Fixed throttle issue in storage cache prepopulation
- Added memory usage logging in LibP2PNode and Hub
- Implemented optional heap dumps for memory analysis


apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -93,7 +96,9 @@ export class StorageCache {
       log.error("cannot prepopulate message counts, db is not open");
       throw new HubError("unavailable.storage_failure", "cannot prepopulate message counts, db is not open");
     }
-    this.prepopulateMessageCounts();
+
+    // Run the prepopulation without waitinf for it to finish
```
Wazzymandias, line 100:
```suggestion
    // Run the prepopulation without waiting for it to finish
```

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -143,6 +148,13 @@ export class StorageCache {
   async getMessageCount(fid: number, set: UserMessagePostfix, forceFetch = true): HubAsyncResult<number> {
     const key = makeKey(fid, set);
     if (this._counts.get(key) === undefined && forceFetch) {
+      if (this._pendingMessageCountScans > MAX_PENDING_MESSAGE_COUNT_SCANS) {
+        log.error({ pendingScans: this._pendingMessageCountScans }, "too many pending message count scans");
+        return err(new HubError("unavailable.storage_failure", "too many pending message count scans"));
```
Wazzymandias, line 153: noob question: Would it be problematic if this returns error and add message ends up updating the map to set count to zero in `storageCache.ts:addMessage` ? 

```
      const count = this._counts.get(key) ?? (await this.getMessageCount(fid, set)).unwrapOr(0);
```

adityapk00: Hmm, this is another bug. I'll change it to not update the cache on error in `addMessage`. Thanks!

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -260,7 +274,16 @@ export class StorageCache {
       const set = typeToSetPostfix(message.data.type);
       const fid = message.data.fid;
       const key = makeKey(fid, set);
-      const count = this._counts.get(key) ?? (await this.getMessageCount(fid, set)).unwrapOr(0);
+      let count = this._counts.get(key);
```
Wazzymandias, line 277: Should we do similar in `removeMessage` ? 

```
  private async removeMessage(message: Message): Promise<void> {
      const set = typeToSetPostfix(message.data.type);
      const fid = message.data.fid;
      const count = this._counts.get(key) ?? (await this.getMessageCount(fid, set)).unwrapOr(0);
      if (count === 0) {
        log.error(`error: ${set} store message count is already at 0 for fid ${fid}`);
      } else {
        this._counts.set(key, count - 1);
      }
```

adityapk00: The bug is not present in removeMessage, but maybe we should do it for consistency?

Wazzymandias (approved):

PR #1868, opened 2024-04-04
perf: Use countKeysAtPrefix for storage cache counts

adityapk00:
## Motivation

Instead of using a `forEachIteratorAtPrefix` which incurs FFI + JS object creation + GC overhead, directly count the keys of a prefix in rust to speed up StorageCache calculations

## Merge Checklist

---

## PR-Codex overview
This PR focuses on improving performance in the `@farcaster/hubble` module by optimizing key counting operations.

### Detailed summary
- Introduced `rsDbCountKeysAtPrefix` function to directly count keys at a prefix in Rust
- Added `dbCountKeysAtPrefix` function in Rust to count keys with a given prefix
- Updated `countKeysAtPrefix` method in TypeScript to utilize the new Rust functions
- Refactored key counting operations for efficiency

### DIFF
--- a/apps/hubble/src/storage/stores/storageCache.ts
+++ b/apps/hubble/src/storage/stores/storageCache.ts
@@ -19,7 +19,6 @@ import { makeFidKey, makeMessagePrimaryKey, makeTsHash, typeToSetPostfix } from 
 import { bytesCompare, getFarcasterTime, HubAsyncResult } from "@farcaster/core";
 import { forEachOnChainEvent } from "../db/onChainEvent.js";
 import { addProgressBar } from "../../utils/progressBars.js";
-import { sleep } from "../../utils/crypto.js";
 
 const MAX_PENDING_MESSAGE_COUNT_SCANS = 100;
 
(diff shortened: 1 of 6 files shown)


apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -97,54 +91,9 @@ export class StorageCache {
...
-            // Sleep to allow other threads to run between each fid
-            await sleep(100);
-          }
-
-          prevFid = fid;
-          prevPostfix = postfix;
-        }
-      }
-    });
-    this.prepopulateComplete = true;
```
sanjayprabhu, line 144: We should also get rid of usages of this. It's always false now.

apps/hubble/src/storage/stores/storageCache.ts
```diff
@@ -154,11 +103,7 @@ export class StorageCache {
...
       this._pendingMessageCountScans += 1;
-
-      let total = 0;
-      await this._db.forEachIteratorByPrefix(makeMessagePrimaryKey(fid, set), () => {
-        total += 1;
-      });
+      const total = await this._db.countKeysAtPrefix(makeMessagePrimaryKey(fid, set));
 
       // Recheck the count in case it was set by another thread (i.e. no race conditions)
       if (this._counts.get(key) === undefined) {
```
sanjayprabhu, line 109: Not required for this PR, but now that this calculation is fast, we should have an LRU cache for this, otherwise this map will keep growing at numFids * numMessageTypes

sanjayprabhu (commented):
Nice, any idea how much faster this is?

Wazzymandias (approved):
🔥

apps/hubble/src/storage/stores/storageCache.ts
adityapk00: It takes <1ms on my machine, but I suspect it is slower on fargate. Probably should move it after moving to EC2. I'll add a comment to remind us.

adityapk00:
This now takes <1ms vs ~8ms before. Plus, we can also now do 4 of these in parallel. So net quite a bit faster.

PR #1869, opened 2024-04-04
feat: Opt Out Hub Diagnostics 

Wazzymandias:
## Motivation

- Hub issues can be difficult to troubleshoot because they are part of decentralized network 
- There is no guarantee subtle edge cases will be surfaced in timely manner

## Change Summary

- Add support for opt-out diagnostics reporter. Diagnostics are used to troubleshoot user issues and improve health of the network. No sensitive information is shared.
- Add `--opt-out-diagnostics` flag. If `--opt-out-diagnostic true`, no errors will be reported to the Farcaster foundation team
- Add environment variable `L2_RPC_AUTHORIZATION_HEADER` to support L2 RPC URLs where authorization header is required for event queries
 

## Merge Checklist

## Additional Context

---

## PR-Codex overview
The focus of this PR is to add opt-out diagnostics reporting and a diagnostic report URL feature to the Farcaster Hub. 

### Detailed summary
- Added opt-out diagnostics reporting feature
- Introduced a diagnostic report URL option
- Updated CLI flags and environment variables
- Implemented diagnostic reporting functionality
- Refactored diagnostic reporting logic into a separate file

> The following files were skipped due to too many changes: `apps/hubble/src/eth/l2EventsProvider.ts`, `apps/hubble/src/utils/diagnosticReportWorker.ts`


apps/hubble/package.json
```diff
@@ -73,6 +73,7 @@
     "@aws-sdk/lib-storage": "^3.504.0",
     "@chainsafe/libp2p-gossipsub": "6.1.0",
     "@chainsafe/libp2p-noise": "^11.0.0 ",
+    "@datadog/datadog-api-client": "^1.23.0",
```
sds, line 76: Rather than include this dependency here, could we instead copy+paste the type definitions from the upstream package (specifically the [parameters](https://github.com/farcasterxyz/hub-monorepo/pull/1869/files#diff-f3c2e0e7964bb1173d352311d9a9353ebf51dc92de3344eb56772fd3bed7d05fR28) that are passed to `createdEvent` and instead have this just send an HTTP request with a POST body to a URL we control (e.g. `report.farcaster.xyz`)?

This will allow us to set up and manage per-IP rate limits much more easily (i.e. doesn't require us to roll out a hub code change) and doesn't require us to expose our Datadog API key.

We could make the endpoint configurable so that users could build their own report processor if they wanted to monitor their own hubs (nice to have, but not a hard requirement for now).

apps/hubble/src/cli.ts
```diff
@@ -145,6 +146,14 @@ app
     'The host to send statsd metrics to, eg "127.0.0.1:8125". (default: disabled)',
   )
 
+  // Opt-out Diagnostics Reporting
+  .option(
+    "--opt-out-diagnostics [boolean]",
+    "Opt-out of sending diagnostics data to Warpcast. " +
```
sds, line 152: Let's use "the Farcaster foundation" instead of Warpcast here and elsewhere in the docs etc., since the `hub-monorepo` project is technically property of the foundation.

apps/hubble/src/cli.ts
```diff
@@ -597,6 +606,14 @@ app
       process.exit(1);
     }
 
+    // Opt-out Diagnostics Reporting
+    const optOut = cliOptions.optOutDiagnostics ? cliOptions.optOutDiagnostics === "true" : hubConfig.optOutDiagnostics;
+    initDiagnosticReporter({
+      optOut,
+      ...(options.hubOperatorFid && { fid: options.hubOperatorFid }),
```
sds, line 613: Here I am being a sucker and writing:

```ts
...(options.hubOperatorFid ? { fid: options.hubOperatorFid } : undefined)
```

...when this is much slicker. TIL!

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -150,11 +152,24 @@ export class L2EventsProvider {
...
   ): L2EventsProvider {
     const l2RpcUrls = l2RpcUrl.split(",");
-    const transports = l2RpcUrls.map((url) => http(url, { retryCount: 10 }));
+    const transports = l2RpcUrls.map((url) =>
+      http(url, {
+        retryCount: 10,
+        fetchOptions: {
+          ...(process.env["L2_RPC_AUTHORIZATION_HEADER"] && {
+            headers: {
+              Authorization: `${process.env["L2_RPC_AUTHORIZATION_HEADER"]}`,
```
sds, line 161: Intentional for this to be part of this PR? (fine to leave, just curious if this was supposed to be part of another PR)

sds (commented):
Looks great! Excited to get this functionality in place.

apps/hubble/src/eth/l2EventsProvider.ts
Wazzymandias: I was lazy and added it here, but can separate as well

apps/hubble/src/cli.ts
Wazzymandias: good call, updated, lmk if I missed any

apps/hubble/package.json
Wazzymandias: That's a really good idea, yeah I can make those changes

apps/hubble/src/cli.ts
```diff
@@ -145,6 +146,14 @@ app
     'The host to send statsd metrics to, eg "127.0.0.1:8125". (default: disabled)',
   )
 
+  // Opt-out Diagnostics Reporting
+  .option(
+    "--opt-out-diagnostics [boolean]",
+    "Opt-out of sending diagnostics data to the Farcaster foundation. " +
+      "Diagnostics are used to troubleshoot user issues and improve health of the network." +
+      "No sensitive information is shared with Warpcast team. (default: disabled)",
```
sds, line 154: Can rewrite to just be "No sensitive information is shared."

sanjayprabhu:
We should also support this via the .env file for the install script. And longer term we need a better solution than manually adding these flags in two places.

apps/hubble/package.json
Wazzymandias: done

PR #1907, opened 2024-04-15
feat: diff sync v2

adityapk00:
## Motivation

The current version of DiffSync (v1) is very slow. It can sync about ~150 messages/sec, and it will often not be able to catch up even it is a few million messages bit behind. To improve on DiffSync, we need to fix the following core weakness of the v1 algorithm:

## 1. No Prioritization

DiffSync v1 is a recursive algorithm, which means it processes each level of the trie, then recurses to the next level. Any prioritization can done only amongst a node’s children, which is local, but globally sub-optimal. For example:

Let's say node is missing 1M messages under the node 09183. At the first level, the algorithm will fetch all children of “0” - So it will fetch 09, see that hashes don’t match, and recurse into node 09. So far so good. 

However, now that it executes node 09, it will also fetch node 08, 07 etc… . At the next level, the nodes fetched will be 091, 081, 071 etc…, since sync parallelization only applies at the top levels of the tree. As you can see, the algorithm should be going into node 09183, but it is wasting its time fetching node 081, 071 etc… 

The nodes being fetched, in order, are: [09, 08, 07, 06, 05, 04, 03, 02, 01], [091, 081, 071, 061….] instead of [09, 091, 0918, 0983…]

The core problem is that there is no way to prioritize across levels. In our example, node 091 is more important than node 08, and node 0918 is more important than node 07. The v1 algorithm doesn’t make these prioritizations.

## 2. HASHES_PER_FETCH

DiffSync v1 has this idea of HASHES_PER_FETCH (= 128), which limited the number of hashes (messages) that we will fetch at each node. If a node had more than this number of missing messages, the algorithm would drill down into the children, even though we were already at a node where we were missing 100% of the messages. This wastes a lot of time pointlessly. 

This idea is dropped in v2, and we fetch all messages at a missing node immediately. 

## 3. Divergence Prefix

DiffSync v1 has this idea of a divergence prefix which is the highest level of the trie that diverges, with hashes indicating which nodes are missing among the children at a given prefix. However, since Hubble is merging almost 50 messages/s, each of which changes the root node’s hash, in practice, this idea is useless as the root prefix will always be different.

We drop this idea in v2, and always start the diff at the root node. 

## 4. Deterministic

The trie nodes are explored in the same deterministic order each time, and since the sync times out, some nodes are never reached, even though we might be missing some messages at that node. 

## 5. More Messages

When our node has more messages than the remote node, it is either because:

1. These are messages that have been removed by a subsequent MessageRemove, which we are missing.
2. These are old messages for an fid, which haven’t been purged yet because we don’t have the newer messages of this fid which will send the fid over storage quota and purge these messages.
3. We have messages that the remote node doesn’t have

In any case, we shouldn’t be syncing this node at all, because there is nothing for us to sync here. In cases 1 or 2, we need to sync the later nodes, and in case 3, the remote node needs to sync with us.

DiffSync v1 was wasting a lot of time at these nodes, because it was based on nodes where hashes don’t match. We fix this in v2 by scoring nodes where we have more messages as 0 (i.e., lowest priority)

## DiffSync v2

The DiffSync v2 algorithm introduces the idea of a work queue. This is a single prioritized queue of trie nodes that need to be fetched and compared. The algorithm is:

1. The root node is added to the work queue (trie node ”0”)
2. Executor fetches from the front of the queue (node “0”), and fetches the metadata for that node. The metadata includes children and the number of messages for each child.
3. Each child node is compared against our trie’s child nodes and scored based on the number of missing messages, and is added to the work queue. The work queue now looks like this: [09, 05, 02, 06…]. Node “09” has the highest delta of messages, so it is at the front of the queue.
4. If this node has 0 messages for us and >1 message for the remote, fetch and merge all of the messages (i.e., if we find a missing node, immediately fetch all messages)
5. Go to step 2. 

In the second iteration, the children of “09” are fetched, and scored. Then, the queue looks like [091, 096, 098, 091…., 08, 07,…]. Sync parallelization now kicks in, and the nodes with the most potential for new messages are at the front of the queue, and are fetched in parallel.

### Scoring Function

To score each node, we use the formula:

$$
score(node) = \frac{max(0, theirMessages - ourMessages)}{10^{10 - min(10, depth)}} 
$$

Or, as code:

```jsx
// Score the work item
const depth = TIMESTAMP_LENGTH - Math.min(TIMESTAMP_LENGTH, this.prefix.length);
const messagesDiff = Math.max(0, theirNode.numMessages - (ourNode?.numMessages ?? 0));
this.score = messagesDiff / TIMESTAMP_LENGTH ** depth;
```

The idea is that we want to normalize the scores across nodes at different depths. We find the number of extra messages the peer has at each node, and divide it by the max number of leaves that are possible below that node (upto a depth of 10, the number of timestamp nodes). This allows us to drill down quickly at leaf nodes that have a lot of missing messages.

1. All nodes where we have more messages are scored 0. This is so that they are fetched in a random order each time, since it is possible that we still have some missing messages. (See the “More messages” section above
2. All nodes at depth ≥10  have their score set to the number of excess messages the remote peer has. This is so that we fetch the nodes that have the most missing messages first.

## Parallelization

In DiffSync v2, we use upto 8 parallel threads to fetch and compare nodes. When we get to a node where we have 0 messages but the remote peer has >1 messages, we fetch all messages from that node and merge them

## Merging using Bundles

All the messages from the remote peer at a node are fetched and merged as a single bundle, allowing them to be merged faster.

## Multiple Remote Peers

In DiffSync v2, we introduce the idea of secondary remote Peers. These are upto 8 additional peers, each of which have more messages than us. When fetching missing messages at a node, we fetch from one of these secondary peers (in round-robin). This allows us to fetch lots of  messages, distributing the load across multiple peers. We do run the risk of some missing messages if this particular peer doesn’t have them, but we will catch up on the next sync. Fetching from multiple peers allows the sync to behave more like a torrent client, dramatically increasing the sync speed

## Benchmarking

With DiffSync v2, we can fetch ~1500 messages/s or about 6M messages per hour. This allows a node that has been down for one day to catch up in just one hour.

## Merge Checklist

---

## PR-Codex overview
This PR improves performance by switching to a workQ-based sync algorithm, updates default message limits, and enhances message submission handling.

### Detailed summary
- Switched to workQ based sync algorithm for performance improvement
- Updated default message limit for catchup sync
- Improved message submission handling for better performance and reliability

> The following files were skipped due to too many changes: `apps/hubble/src/hubble.ts`, `apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts`, `apps/hubble/src/network/sync/syncEngine.ts`


apps/hubble/src/hubble.ts
```diff
@@ -1453,17 +1452,11 @@ export class Hub implements HubInterface {
...
-        (e) => e,
-      );
-      if (syncResult.isErr()) {
-        log.error({ error: syncResult.error, peerId }, "Failed to sync with new peer");
-      }
+      // Sync with the first peer so we are upto date on startup.
       this.performedFirstSync = true;
+      setTimeout(async () => {
+        await ResultAsync.fromPromise(this.syncEngine.diffSyncIfRequired(this), (e) => e);
+      }, 10 * 1000);
```
Wazzymandias, line 1459: nit: can we make this a constant with unit suffix? e.g. 
```suggestion
      const DIFF_SYNC_TIMEOUT_MS = 10 * 1000;
```

apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -561,43 +614,67 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
           `Diffsync: Choosing random peer among ${peers.length} peers with more messages`,
         );
       }
-      const randomPeer = peers[Math.floor(Math.random() * peers.length)];
-      peerContact = randomPeer?.contactInfo;
-      peerId = randomPeer?.peerId;
+
+      // Pick a random peer
+      peerContact = peers[Math.floor(Math.random() * peers.length)] as PeerContact;
```
Wazzymandias, line 619: We have peers in the network who don't have gRPC port exposed - they could be reachable by p2p but not RPC 

I think we should check that the peer is `diffSyncable` before we add it as secondary

Wazzymandias: I see - if the port is closed we could short circuit the retries, but we could do that optimization later

Wazzymandias (approved):

apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -561,43 +614,67 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
...
+      secondaryContacts.push(peerContact);
+
+      // Pick a more random peers, one per sync thread. It's OK if we pick the same twice, it might
+      // happen if we don't have that many peers.
+      for (let i = 0; i < SYNC_PARALLELISM; i++) {
```
sanjayprabhu, line 624: Maybe we should cap the parallelism to peer set size? If we only have 1 peer and have 8 threads, could cause a lot of extra load. This can happen on start up/first sync. And our bootstrap nodes might be disproportionately affected.

apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -1168,189 +1178,239 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
...
+    // If we are interrupted, we will not schedule any more work
+    if (this.curSync.interruptSync) return;
+
+    // If the hashes match, we do not need to do any work
+    if (ourNode && ourNode.hash === theirNode.hash) return;
+
+    // If there are no messages to fetch, don't enque the work item
+    if (theirNode.numMessages === 0) return;
+
+    if (this.curSync.workQueue.length > MAX_WORK_QUEUE_SIZE) return;
```
sanjayprabhu, line 1194: When we hit this limit, we will stop traversing this branch of the tree right? We just wait for it to be fixed in the next round of sync?

apps/hubble/src/network/sync/syncEngine.ts
```diff
@@ -1168,189 +1178,239 @@ class SyncEngine extends TypedEmitter<SyncEvents> {
...
+      }
+
+      // Replace the work queue with only the items that are not completed
+      this.curSync.workQueue = newWorkQueue;
+
+      // If there is an opputunity to do more work, we will do it
+      if (this.curSync.executingWorkCount < SYNC_PARALLELISM) {
+        const n = Math.max(0, SYNC_PARALLELISM - this.curSync.executingWorkCount);
+        // Find the "n" best work items, and start them
+        const bestWorkItems = scheduledItems.sort((a, b) => b.score - a.score).slice(0, n);
```
sanjayprabhu, line 1237: You probably already considered this, but wondering why not use an heap or a priority queue implementation vs sorting each time?

apps/hubble/src/hubble.ts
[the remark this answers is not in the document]
GarrettJMU: Also curious on the above, but also agree -- no magic numbers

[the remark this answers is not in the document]
adityapk00: It's takes about ~10s after startup for all the bootstrap nodes to get connected, so I added that delay so for the first sync there are enough peers to pick to sync from

apps/hubble/src/network/sync/syncEngine.ts
adityapk00: Good idea. Done

apps/hubble/src/network/sync/syncEngine.ts
adityapk00: Right. I wanted a backstop so that the queue doesn't grow indefinitely

apps/hubble/src/network/sync/syncEngine.ts
adityapk00: Yeah, I tried to use a priority queue at first, but the challenge is that the library I was using (https://www.npmjs.com/package/@datastructures-js/priority-queue) didn't support "cleaning" out the completed tasks. They all assume we deque one item at a time and execute them, and didn't work well with the sync parallelization that we're using. 

I'm sure it can be done, and I will follow up in a PR where I'm doing some cleanup and making this more modular

PR #1947, opened 2024-04-25
perf: Improve getSyncMetadataByPrefix performance

adityapk00:
## Motivation

This API walks the trie to get the sync metadata at the given prefix, traversing on avg 12 nodes, reading each node from the DB. However, since we already have the prefix, we can read the final node directly from the DB, saving 11 hits to the DB for every API call! 

## Merge Checklist

---

## PR-Codex overview
This PR focuses on improving performance in `getSyncMetadataByPrefix` and enhancing testing for `MerkleTrie` operations.

### Detailed summary
- Added a new test module `merkle_trie_tests` for testing `MerkleTrie` functionality
- Updated `MerkleTrie` methods for better performance and error handling
- Refactored `TrieNode` struct and related methods for clarity and consistency

### DIFF
--- a/apps/hubble/src/addon/src/trie/merkle_trie.rs
+++ b/apps/hubble/src/addon/src/trie/merkle_trie.rs
@@ -218,6 +218,30 @@ impl MerkleTrie {
         }
     }
 
+    pub fn get_node(&self, db: &Arc<RocksDB>, prefix: &[u8]) -> Option<TrieNode> {
+        let node_key = TrieNode::make_primary_key(prefix, None);
+
+        // We will first attempt to get it from the DB cache
+        if let Some(Some(node_bytes)) = self.txn_batch.lock().unwrap().batch.get(&node_key) {
+            if let Ok(node) = TrieNode::deserialize(&node_bytes) {
+                return Some(node);
+            }
+        }
+
+        // Else, get it directly from the DB
+        if let Some(node_bytes) = db.get(&node_key).ok().flatten() {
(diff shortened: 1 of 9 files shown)


Wazzymandias (approved):
🔥 You are a mind reader

apps/hubble/src/addon/src/trie/merkle_trie.rs
```diff
@@ -218,6 +225,30 @@ impl MerkleTrie {
...
+        }
+
+        // Else, get it directly from the DB
+        if let Some(node_bytes) = self.db.get(&node_key).ok().flatten() {
+            if let Ok(node) = TrieNode::deserialize(&node_bytes) {
+                return Some(node);
+            }
+        }
+        // If not found, get it the normal way from the trie root
+        self.root
```
sanjayprabhu, line 245: When is it possible for this to return a result when the other two methods failed? It's either in the cache or the the db right? Is there a third state?

sanjayprabhu (approved):
Nice

apps/hubble/src/addon/src/trie/merkle_trie.rs
adityapk00: You are technically correct that there can only be the first 2 cases. However, I became suspicious that there might be some weird DB corruption or incorrect keys written previously because of previous bugs, so I added this fallback. 

I'm running a scan on keat's snapshot to check for this - Nothing has turned up yet, so it might be a false alarm. In any case, I have a follow up PR, and I'll remove it in that if it turns out to be nothing.

PR #1989, opened 2024-05-07
Prune messages after merge

loriopatrick:
Updates isPrunable to be getPruneAction which lets caller know if
message should be pruned (original function) or would cause a prune
action (added functionality). When applicable we prune for the given
FID after the message is merged.

Prune job is still running in the background but can be removed
in the future once it's confirmed there's no stragglers.

Benchmarking with production data, this change seems to add a 5%
overhead on average. The FID lock added to Store::prune_messages
is required for performance reasons. Without the lock, the overhead
is run dependent and takes 100%-500% longer.

----

## Motivation

Currently hubs are seeing regular latency spikes due the periodic pruning job. This update causes the hub to immediately prune messages for a given `FID` after the message is merged.

## Change Summary

Updates `isPrunable` to be `getPruneAction` which lets caller know if message should be pruned (original function) or would cause a prune action (added functionality). When applicable we prune for the given `FID` after the message is merged.

## Merge Checklist

## Additional Context

Adds a FID based lock to `Store::prune_messages`. This might be an issue for the prune job which are left running. If a prune job is taking a long time it may block merges. I suspect this won't be too different from what's happening today during prunes as the threadpool is limited to 4 threads, so a slow prune would already be blocking things.

---

## PR-Codex overview
This PR focuses on enhancing message merge operations in the `store.rs` file and adding a `PruneAction` enum in `storeEventHandler.ts`.

### Detailed summary
- Improved atomic merge operations with locks for performance
- Added `PruneAction` enum to determine pruning actions
- Updated functions for better message pruning and handling

> The following files were skipped due to too many changes: `apps/hubble/src/storage/stores/reactionStore.test.ts`, `apps/hubble/src/storage/stores/rustStoreBase.ts`, `apps/hubble/src/storage/stores/castStore.test.ts`

### DIFF
--- a/apps/hubble/src/addon/src/store/store.rs
+++ b/apps/hubble/src/addon/src/store/store.rs
@@ -807,7 +807,7 @@ impl Store {
         ts_hash: &[u8; TS_HASH_LENGTH],
         message: &Message,
     ) -> Result<Vec<u8>, HubError> {
-        // If the store supports compact state messages, we don't merge messages that exist in the compact state
+        // If the store supports compact state messages, don't merge messages older than the compact state message
         if self.store_def.compact_state_type_supported() {
             // Get the compact state message
             let compact_state_key = self.store_def.make_compact_state_add_key(message)?;
--- a/apps/hubble/src/addon/src/store/store.rs
+++ b/apps/hubble/src/addon/src/store/store.rs
@@ -616,9 +616,13 @@ impl Store {
     }
 
     pub fn merge(&self, message: &Message) -> Result<Vec<u8>, HubError> {
-        // Grab a merge lock. The typescript code does this by individual fid, but we don't have a
-        // good way of doing that efficiently here. We'll just use an array of locks, with each fid
-        // deterministically mapped to a lock.
+        // Merge operations should be atomic so we need a lock. Note, there
+        // might be extra lock conflicts due to impl.
+        //
+        // Also helps performance with RocksDB if we aren't touching the same
+        // memory space. In our case, two threads competing for locks between
+        // api calls costs more than an exclusive lock blocking the other
+        // thread entirely.
         let _fid_lock = &self.fid_locks
             [message.data.as_ref().unwrap().fid as usize % FID_LOCKS_COUNT]
             .lock()
@@ -803,7 +807,7 @@ impl Store {
         ts_hash: &[u8; TS_HASH_LENGTH],
         message: &Message,
     ) -> Result<Vec<u8>, HubError> {
-        // If the store supports compact state messages, we don't merge messages that don't exist in the compact state
+        // If the store supports compact state messages, we don't merge messages that exist in the compact state
         if self.store_def.compact_state_type_supported() {
             // Get the compact state message
             let compact_state_key = self.store_def.make_compact_state_add_key(message)?;
@@ -927,6 +931,11 @@ impl Store {
         cached_count: u64,
         units: u64,
     ) -> Result<Vec<HubEvent>, HubError> {
+        // for performance, prevent locking inside of RocksDB due to touching the same memory space (fid)
+        let _fid_lock = &self.fid_locks[fid as usize % FID_LOCKS_COUNT]
+            .lock()
+            .unwrap();
+
         let mut pruned_events = vec![];
 
         let mut count = cached_count;
--- a/apps/hubble/src/storage/stores/castStore.test.ts
+++ b/apps/hubble/src/storage/stores/castStore.test.ts
@@ -658,7 +658,7 @@ describe("revoke", () => {
   });
 });
 
-describe("mergePrunesMessages", () => {
+describe("merge should prunes messages", () => {
   let prunedMessages: Message[];
   const pruneMessageListener = (event: PruneMessageHubEvent) => {
     prunedMessages.push(event.pruneMessageBody.message);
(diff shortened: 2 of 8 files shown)


apps/hubble/src/addon/src/store/store.rs
```diff
@@ -803,7 +803,7 @@ impl Store {
         ts_hash: &[u8; TS_HASH_LENGTH],
         message: &Message,
     ) -> Result<Vec<u8>, HubError> {
-        // If the store supports compact state messages, we don't merge messages that don't exist in the compact state
```
loriopatrick, line 806: I think this comment was wrong, or am I missing something?

apps/hubble/src/storage/stores/rustStoreBase.ts
```diff
@@ -117,19 +123,26 @@ export abstract class RustStoreBase<TAdd extends Message, TRemove extends Messag
       }
     }
 
+    for (const [fid, _] of fidToPrune) {
+      const pruneResult = await this.pruneMessages(fid);
```
sanjayprabhu, line 127: This approach works, but I'm curious if it slows down the merge code path.

I suspect this is somewhat inefficient, but I don't know for sure.

e.g. current flow is:
 - check if is prunable in js (gets the storage count and earliest ts hash)
 - merge in rust
 - call the prune logic from js, which again gets the storage count etc and passes it to rust which iterates the db to find the earliest hash and prunes it.

But isPrunable already knows the storage units, message count and the earliestTsHash. One approach could be the pass in the earliestTsHash if it would cause a prune, and have the store prune the earliest message within the same transaction as the merge. Then we don't have to go back and forth between js and rust and we save some db calls.

But if you think perf impact is not that bad, then this is definitely simpler code wise.

apps/hubble/src/addon/src/store/store.rs
sanjayprabhu: I think it should say: "we don't merge messages older than the compact state message and don't exist within it". Compact state message is like a checkpoint, you cannot go back and add messages that weren't present at the time of compaction.

apps/hubble/src/storage/stores/rustStoreBase.ts
loriopatrick: I'll update my benchmarking branch to see how much overhead the pruneMessage call adds to try and get some real numbers.

Reading `getMessageCount`, it seems like the value is cached so the second load shouldn't be too bad.

> One approach could be the pass in the earliestTsHash if it would cause a prune, and have the store prune the earliest message within the same transaction as the merge.

I like the idea of that. When you add a single message it can optionally remove the latest stored item in the same operation. Is the `earliestTsHash` for a particular `FID`  & store cached somewhere on the javascript side? If not, we'd probably still need to hit the db again. It would also change `mergeMessages` a bit as you'd need to do a remove for each merge rather than for each unique FID at the end (probably doesn't matter that much).

-----

Here's an alternative design that would probably help with perf. We keep the prune job but `mergeMessages` simply helps the job track which FIDs on which stores need pruning. The job runs more frequently and only targets what needs to be updated. That would avoid the extra logic on the hot path and make the prune job run much faster. Thoughts? Might just be worth making that change as the design is still fairly simple.

sanjayprabhu: Yes, isPrunable fetches the earliest tsHash from cache. I think a design where we can perform the prune at the same time as the merge so we don't even need the job would be better. 

If that is too complicated, I prefer the current approach of doing the prune within the merge call but outside the tx. I don't think we should do the third of just having the job run more frequently. I think we should eventually get rid of the job entirely.

The first approach above is more effort, so maybe let's decide based on the benchmark results?

apps/hubble/src/addon/src/store/store.rs
```diff
@@ -615,14 +636,19 @@ impl Store {
         Ok(())
     }
 
-    pub fn merge(&self, message: &Message) -> Result<Vec<u8>, HubError> {
-        // Grab a merge lock. The typescript code does this by individual fid, but we don't have a
+    pub fn merge(&self, message: &Message, max_mesages_for_fid: u32) -> Result<Vec<u8>, HubError> {
```
loriopatrick, line 639: New prune logic is going here.

todo: move argument `max_mesages_for_fid` to `FidInfoCache`

apps/hubble/src/storage/stores/castStore.test.ts
```diff
@@ -745,34 +751,44 @@ describe("pruneMessages", () => {
...
     });
 
-    test("prunes earliest add messages with bundles", async () => {
-      const messages = [add1, add2, add3, add4, add5];
-      const mergeResults = await sizePrunedStore.mergeMessages(messages);
-      expect(mergeResults.size).toEqual(messages.length);
-      for (const result of mergeResults.values()) {
-        expect(result.isOk()).toBeTruthy();
-      }
+    test("earlier remove messages gets pruned", async () => {
```
sanjayprabhu, line 754: This test seems duplicated? One of these should test the mergeMessages (bundles) code path

apps/hubble/src/storage/stores/castStore.test.ts
```diff
@@ -658,7 +658,7 @@ describe("revoke", () => {
   });
 });
 
-describe("pruneMessages", () => {
+describe("mergePrunesMessages", () => {
```
sanjayprabhu, line 661: When we use camel case, these map one-to-one with method names. I'd probably just make this a short sentence instead.

apps/hubble/src/addon/src/store/store.rs
```diff
@@ -803,7 +807,7 @@ impl Store {
         ts_hash: &[u8; TS_HASH_LENGTH],
         message: &Message,
     ) -> Result<Vec<u8>, HubError> {
-        // If the store supports compact state messages, we don't merge messages that don't exist in the compact state
+        // If the store supports compact state messages, we don't merge messages that exist in the compact state
```
sanjayprabhu, line 810: "don't merge messages older than the compact state message"

sanjayprabhu (commented):
Looks good, I would just update the tests to cover both the `mergeMessages` and `mergeMessage` codepath. Don't have to extensive, just one of the tests using the mergeMessages should be good.

apps/hubble/src/storage/stores/castStore.test.ts
loriopatrick: Oops. A lot of the tests are similar across stores, I must have copied / migrated the test over incorrectly. Will fix :+1:

sds:
Closing due to age, and not actively being pursued. Thanks!

PR #1996, opened 2024-05-13
fix: remove unnecessary database reads when merging casts

FrederikBolding:
## Motivation

Since the base store implementation already handles conflicts for casts correctly: https://github.com/farcasterxyz/hub-monorepo/blob/main/apps/hubble/src/addon/src/store/store.rs#L195-L274 it feels wasteful to potentially use two additional DB lookups checking for an identical edge-case.

This PR removes these two additional lookups and lets the base store logic handle casts in the same way that all other stores are handled.

## Change Summary

- Remove two unnecessary database reads when merging casts

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR focuses on removing unnecessary database reads when merging casts to improve efficiency and accuracy.

### Detailed summary
- Removed unnecessary database reads when merging casts
- Updated error message for conflicts with a more recent remove
- Refactored the `find_merge_add_conflicts` function to remove unnecessary checks
- Updated test cases to reflect the new error message for conflicts

### DIFF
--- a/apps/hubble/src/addon/src/store/cast_store.rs
+++ b/apps/hubble/src/addon/src/store/cast_store.rs
@@ -93,29 +93,9 @@ impl StoreDef for CastStoreDef {
 
     fn find_merge_add_conflicts(
         &self,
-        db: &RocksDB,
-        message: &protos::Message,
+        _db: &RocksDB,
+        _message: &protos::Message,
     ) -> Result<(), super::store::HubError> {
-        // Look up the remove tsHash for this cast
-        let remove_key = self.make_remove_key(message)?;
-        // If remove tsHash exists, fail because this cast has already been removed
-        if let Ok(Some(_)) = db.get(&remove_key) {
-            return Err(HubError {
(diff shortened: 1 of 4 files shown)


apps/hubble/src/addon/src/store/cast_store.rs
```diff
@@ -93,29 +93,9 @@ impl StoreDef for CastStoreDef {
 
     fn find_merge_add_conflicts(
         &self,
-        db: &RocksDB,
-        message: &protos::Message,
+        _db: &RocksDB,
+        _message: &protos::Message,
     ) -> Result<(), super::store::HubError> {
```
sanjayprabhu, line 98: Hmm, so one difference is that we'll now accept messages with higher timestamps for the same hash, when before we would be rejecting them. In practice, I'm not sure this is actually possible since the timestamp is part of the hash. And we don't have a specific test for this case. I think it's probably fine? But let me double check the protocol spec again and make sure.

sanjayprabhu (approved):

Issue #863, opened 2023-04-12
getHubRpcClient "Named export 'grpc' not found"  in hub-web


zachterrell57:
**What is the bug?**
Getting the following error when trying to run the `hub-web` example:

SyntaxError: Named export 'grpc' not found. The requested module '@improbable-eng/grpc-web' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@improbable-eng/grpc-web';
const { grpc } = pkg;

This error happened while generating the page. Any console logs will be displayed in the terminal window.
Call Stack
<unknown>
file:///Users/zachterrell/Developer/farcaster/farcaster.network/web/node_modules/@farcaster/hub-web/dist/index.mjs (2640)
ModuleJob._instantiate
node:internal/modules/esm/module_job (128:21)
async ModuleJob.run
node:internal/modules/esm/module_job (194:5)
async ESMLoader.import
node:internal/modules/esm/loader (409:24)
async importModuleDynamicallyWrapper
node:internal/vm/module (437:15)

**How can it be reproduced? (optional)**
Running Next.JS 12.2.4 with Typescript, @farcaster/hub-web": "^0.2.4

```typescript
import { getHubRpcClient } from '@farcaster/hub-web'
const fetchData = async () => {
    const client = getHubRpcClient('https://testnet1.farcaster.xyz:2285')
    const castsResult = await client.getCastsByFid({ fid: 15 })

    castsResult.map((casts) => casts.messages.map((cast) => console.log(cast.data?.castAddBody?.text)))
  }

  fetchData()
```

pfletcherhill:
The issue as I understand it is that `@improbable-eng/grpc-web` is a CommonJS module, but the [generated `rpc.ts` file from ts-proto](https://github.com/farcasterxyz/hub-monorepo/blob/main/packages/hub-web/src/generated/rpc.ts) as well as our [`client.ts` file](https://github.com/farcasterxyz/hub-monorepo/blob/main/packages/hub-web/src/client.ts) imports the named export `grpc` from it rather than a default export. This is problematic in some environments, but not all—for instance, you can use the package fine when using Vite. If you have cycles to help, you could try patching the rpc.ts and client.ts files to use the default export, but if that doesn't work we may have to look at using a grpc-web library that uses esm instead.

pfletcherhill:
[...]

PR #1999, opened 2024-05-14
fix(hub-web import): undo change that breaks build in certain environments

hiporox:
## Motivation

The most recent change to this file reintroduced the issue identified in https://github.com/farcasterxyz/hub-monorepo/issues/863

This undoes that breaking change

## Change Summary

Remove curly braces to make import into default instead of named to fix an import error in certain builds

## Merge Checklist

---

## PR-Codex overview
This PR updates the import statement in `packages/hub-web/src/generated/rpc.ts` to fix an import error that was causing build failures.

### Detailed summary
- Updated import statement for `grpc` from "@improbable-eng/grpc-web" to `grpcWeb`
- Renamed `grpc` variable to `grpcWeb.grpc` to maintain functionality

### DIFF
--- a/.changeset/five-kings-love.md
+++ b/.changeset/five-kings-love.md
@@ -1,0 +1,5 @@
+---
+"@farcaster/hub-web": minor
+---
+
+fix import error that breaks some builds
--- a/packages/hub-web/src/generated/rpc.ts
+++ b/packages/hub-web/src/generated/rpc.ts
@@ -1,5 +1,5 @@
 /* eslint-disable */
-import { grpc } from "@improbable-eng/grpc-web";
+import grpcWeb from "@improbable-eng/grpc-web";
 import { BrowserHeaders } from "browser-headers";
 import { Observable } from "rxjs";
 import { share } from "rxjs/operators";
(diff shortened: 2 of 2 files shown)


hiporox:
Since this is a problem in generated files, it might crop up again. I can't find a way to solve this issue with the generator, so as @pfletcherhill mentioned in #863, fixing this might require using a different package for gRPC.

I also have to assume that there is someway to configure the Next.js compiler to work the way that Vite does, but I cannot find information for how to do so

hiporox:
It is worth noting that [@improbable-eng/grpc-web](https://github.com/improbable-eng/grpc-web/) is in maintenance mode as of their [last commit over a year ago](https://github.com/improbable-eng/grpc-web/commit/1d9bbb09a0990bdaff0e37499570dbc7d6e58ce8). They suggest moving to the official [grpc/grpc-web](https://github.com/grpc/grpc-web/) package. It does appear that package is ESM compatible

hellno:
just ran into this problem, thanks for providing a fix so quickly @hiporox 🙏

hiporox:
@hellno For the time being, `@farcaster/hub-web@0.8.4` is the last release before the breaking change happened, so you can use that one

sanjayprabhu:
Ah, thank you. Looks like there was a regression, and we accidentally removed some code that was handling this when protoc is run. More future proof fix here: https://github.com/farcasterxyz/hub-monorepo/pull/2006.

PR #2012, opened 2024-05-18
perf: remove by signer index

FrederikBolding:
## Motivation

The BySigner index is taking up a lot of space and is only used for quickly removing messages after a signer is revoked. As a trade-off, this index can be removed in favor of iterating over more messages when revoking.

## Change Summary

Remove code that populates the "BySigner" index and replace functions that used the index with new implementations that iterate over all messages from a given FID instead.

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR updates message handling in the `hubble` app by refactoring key generation and message processing logic.

### Detailed summary
- Refactored key generation functions in `message.ts`
- Updated message handling in `engine/index.ts`
- Removed unused functions and imports

> The following files were skipped due to too many changes: `apps/hubble/src/storage/db/message.ts`

### DIFF
--- a/apps/hubble/src/addon/src/store/message.rs
+++ b/apps/hubble/src/addon/src/store/message.rs
@@ -92,6 +92,8 @@ pub enum UserPostfix {
     // NOTE: If you add a new message type, make sure that it is only used to store Message protobufs.
     // If you need to store an index, use one of the UserPostfix values below (>86).
     /** Index records (must be 86-255) */
+    // Deprecated
+    // BySigner = 86, // Index message by its signer
 
     /** CastStore add and remove sets */
     CastAdds = 87,
--- a/apps/hubble/src/addon/src/store/message.rs
+++ b/apps/hubble/src/addon/src/store/message.rs
@@ -92,7 +92,6 @@ pub enum UserPostfix {
     // NOTE: If you add a new message type, make sure that it is only used to store Message protobufs.
     // If you need to store an index, use one of the UserPostfix values below (>86).
     /** Index records (must be 86-255) */
-    BySigner = 86, // Index message by its signer
 
     /** CastStore add and remove sets */
     CastAdds = 87,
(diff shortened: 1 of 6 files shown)


apps/hubble/src/storage/db/message.ts
```diff
@@ -215,44 +206,41 @@ export const getPageIteratorOptsByPrefix = (prefix: Buffer, pageOptions: PageOpt
...
-    const tsHash = new Uint8Array(key as Buffer).slice(tsHashOffset);
+  await db.forEachIteratorByPrefix(prefix, (key, value) => {
+    if (!value) {
+      return false;
+    }
+
+    const type = key.readUint8(prefix.length);
+
+    // We assume the iterator runs sequentially and thus we can stop when we pass 85, as that is the max non-index type
+    if (type > 85) {
```
sanjayprabhu, line 228: We should make this a constant like `MAX_MESSAGE_RECORD` or similar

apps/hubble/src/storage/engine/index.ts
```diff
@@ -452,49 +452,60 @@ class Engine extends TypedEmitter<EngineEvents> {
...
         default: {
           return err(new HubError("bad_request.invalid_param", "invalid message type"));
         }
       }
     };
 
-    await this._db.forEachIteratorByPrefix(prefix, async (key) => {
-      const revokeResult = await revokeMessageByKey(key as Buffer);
+    await this._db.forEachIteratorByPrefix(prefix, async (key, value) => {
+      if (!value) {
```
sanjayprabhu, line 491: I think `getAllMessagesBySigner` is not actually used anywhere except tests, so we should change it to `revokeMessagesBySigner` and call it here. If we do need to `getAllMessagesBySigner` later, we can refactor that that point.

sanjayprabhu (commented):
Nice. This should help a lot.

We need a migration to delete the old keys, but that can be a separate PR.

apps/hubble/src/storage/engine/index.ts
FrederikBolding: Hmm, not sure exactly how you would move `revokeMessagesBySigner` to `message.ts` given that it relies on `this._linkStore` etc. Wouldn't it be cleaner to keep the revocation living in `engine/index.ts`?

If `getAllMessagesBySigner` is not used at all, perhaps it can just be deleted?

apps/hubble/src/addon/src/store/message.rs
```diff
@@ -92,7 +92,6 @@ pub enum UserPostfix {
     // NOTE: If you add a new message type, make sure that it is only used to store Message protobufs.
     // If you need to store an index, use one of the UserPostfix values below (>86).
     /** Index records (must be 86-255) */
-    BySigner = 86, // Index message by its signer
```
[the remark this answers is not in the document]
FrederikBolding: 👍

apps/hubble/src/storage/db/message.ts
FrederikBolding: This already existed, cleaned up!

sanjayprabhu:
Moved to https://github.com/farcasterxyz/hub-monorepo/pull/2019, which also handles deleting existing index data. Thank you!

Issue #2001, opened 2024-05-15
bug(hubble): Fetch fname transfers when merge fails due to "username not registered"


sanjayprabhu:
**What is the bug?**
Hubble only polls the fname server for new names every 5 seconds. This means, if a user registers an fid and an fid and submits UserDataAdd messages within 5 seconds, the username message will be rejected since the hub has not picked it up yet.

We should detect this failure, and check the transfer history for that fid and retry the message automatically. We should also put in some protections so this cannot be used to ddos the fname server (simple check like only x calls/min to the fname server should do, user registration is not high traffic)

**How can it be reproduced? (optional)**
Include steps, code samples, replits, screenshots and anything else that would be helpful to reproduce the problem. 

**Additional context (optional)**

henryphaver:
Good finding @sanjayprabhu – We are also getting multiple reports on this from both developers and users. – Because of this issue no username is shown, but instead !123456 which renders the UX useless in almost all cases. Hope this gets prioritized.

0xAegir:
+1 This is biting us as well.

sanjayprabhu:
Are these fnames eventually fixed or are they still missing? Could you provide some examples? Trying to determine if this is related to the issue above or a separate issue. The above issue happens only for new registrations and should be fixed within a few minutes.

andrewhong5297:
+10 this is making the query experience on Dune a bit buggy, a few users have brought this up

henryphaver:
@sanjayprabhu Actually now 2/3 fixed already by themselves. - Will check if the third one is pending refresh on test system. - If not then been stuck for longer. Will report back.

PR #2013, opened 2024-05-22
fix: retry fname transfer fetch on fail

FrederikBolding:
## Motivation

Potential solution for https://github.com/farcasterxyz/hub-monorepo/issues/2001. Adds a mechanism to re-fetch fname transfers for a given name when merging a `UserDataAdd` that fails due to missing proof.

## Change Summary

Refetches fname transfers when merging a `UserDataAdd` that errors out due to a missing proof. Also moves some code around for testing purposes.

## Merge Checklist

## Additional Context

---

## PR-Codex overview
The focus of this PR is to enhance the `@farcaster/hubble` module by improving transfer retry logic and adding a new `FNameRegistryClientInterface` with related functionalities.

### Detailed summary
- Added retry logic for fetching transfers on failed merge
- Introduced `FNameRegistryClientInterface` with methods for managing transfers
- Updated `MockFnameRegistryClient` with transfer handling functionality

> The following files were skipped due to too many changes: `apps/hubble/src/storage/engine/index.test.ts`

### DIFF
--- a/apps/hubble/src/storage/engine/index.test.ts
+++ b/apps/hubble/src/storage/engine/index.test.ts
@@ -74,20 +74,6 @@ const fname = Factories.Fname.build();
 const signer = Factories.Ed25519Signer.build();
 const custodySigner = Factories.Eip712Signer.build();
 
-const transferEvents: FNameTransfer[] = [
-  {
-    id: 1,
-    username: Buffer.from(fname).toString("utf-8"),
-    from: 0,
-    to: fid,
-    timestamp: 1686291736947,
-    owner: account.address,
-    server_signature: "",
-  },
-];
-
-fNameClient.setTransfersToReturn([[], transferEvents]);
-
 let custodySignerKey: Uint8Array;
 let signerKey: Uint8Array;
 let userNameProof: UserNameProof;
@@ -102,6 +88,7 @@ let verificationAdd: VerificationAddAddressMessage;
 let userDataAdd: UserDataAddMessage;
 
 beforeAll(async () => {
+  fNameClient.setTransfersToReturn([]);
   // This forces the provider to fetch the signer address
   await fNameProvider.start();
   await fNameProvider.stop();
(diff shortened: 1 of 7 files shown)


apps/hubble/src/storage/engine/index.ts
```diff
@@ -911,7 +915,14 @@ class Engine extends TypedEmitter<EngineEvents> {
         return err(validatedFname.error);
       }
 
-      return ResultAsync.fromPromise(this._userDataStore.getUserNameProof(name), (e) => e as HubError);
+      const result = await ResultAsync.fromPromise(this._userDataStore.getUserNameProof(name), (e) => e as HubError);
+
+      if (result.isErr() && retries > 0 && this._fNameRegistryEventsProvider) {
```
sanjayprabhu, line 920: We have memory based rate limiter in rateLimits.ts. I think we should use it and set it to something low, like 60/min so hubs don't ddos the fname server.

sanjayprabhu, line 920: We should check that the error code is specifically `not_found`

sanjayprabhu (commented):
Nice

apps/hubble/src/storage/engine/index.ts
FrederikBolding: Fixed!

FrederikBolding: Hmm, its not entirely obvious to me how you want to do the rate limiting? Should it be per FID? Should it be only for the retry functionality or in `FNameRegistryEventsProvider` in general?

sanjayprabhu: I think it should be global (not per fid) and it should only be for the retry via the merge (within this function above). Since this is only one that can be user triggered.

FrederikBolding: Fixed up!

apps/hubble/src/storage/engine/index.test.ts
```diff
@@ -510,11 +511,28 @@ describe("mergeMessage", () => {
...
           expect(result._unsafeUnwrapErr().message).toMatch("is not registered");
         });
 
+        test("retries and succeeds when username proof has not been merged yet", async () => {
+          fNameClient.setTransfersToReturn([]);
+
+          // This is a bit of a hack
+          // Since all of the tests run on the same engine it takes less than 60 attempts to get rate limited
+          for (let i = 0; i < 58; i++) {
+            const result = await engine.mergeMessage(usernameAdd);
```
sanjayprabhu, line 526: Hmm, this is going to causes come flaky tests. One way to work around this would be to call it 61 times, but only look at the error code of the last one and assert that it matches the rate limit error.

sanjayprabhu (approved):

apps/hubble/src/storage/engine/index.test.ts
FrederikBolding: Good point, addressed!

PR #2073, opened 2024-06-19
cli tool for computing sync health and exploring causes for message diffs

aditiharini:
## Motivation

We want a better metric than current number of messages (i.e. "sync percent") for measuring sync health. It doesn't account for the fact that there are valid reasons for temporary displacements in message counts (i.e. compaction, small network delays). Instead, we want to take some chunk of time in the past and count up messages in that chunk and compare the counts as a somewhat more evolved sync metric. 

## Change Summary

CLI command that computes sync health between one node and peers, explores nonzero diffs, and prints out summarized, structured output. 

```
❯ node build/cli.js sync-health --max-num-peers 50 --primary-node lamia.farcaster.xyz:2283 --start-time-ofday 11:30:00 --stop-time-ofday 11:35:00
❯ cat health.out  |  jq
{
  "startTime": "2024-06-21T15:15:00.000Z",
  "stopTime": "2024-06-21T15:17:00.000Z",
  "primary": "hoyt.farcaster.xyz:2283",
  "peer": "84.247.160.59:2283",
  "primaryMessageCount": 5311,
  "peerMessageCount": 4930,
  "diff": 381,
  "diffPercentage": 0.0717379024665788,
  "numSuccessToPeer": 378,
  "numErrorToPeer": 6,
  "successTypesToPeer": [
    null,
    3,
    2,
    6
  ],
  "errorMessagesToPeer": [
    "no storage",
    "invalid signer: signer 0x02c2f67b36cec88270462f95d7c553d13bc339be61c02db591c9144fcf2592dd not found for fid 681926"
  ],
  "numSuccessToPrimary": 0,
  "numErrorToPrimary": 0,
  "successTypesToPrimary": [],
  "errorMessagesToPrimary": []
}
```

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR introduces a CLI tool `hubble` for measuring sync health. It includes commands to measure sync health and calculates message statistics between nodes.

### Detailed summary
- Added `sync-health` command to measure sync health
- Introduced classes for message stats and sync health stats
- Implemented functions to query message counts and compute stats
- Added functions to pick peers and compute sync IDs

### DIFF
--- a/apps/hubble/src/utils/syncHealth.ts
+++ b/apps/hubble/src/utils/syncHealth.ts
@@ -8,7 +8,6 @@ import {
   HubRpcClient,
   TrieNodeMetadataResponse,
   Message,
-  HubError,
 } from "@farcaster/hub-nodejs";
 
 import { appendFile } from "fs/promises";
@@ -28,7 +27,7 @@ class SyncHealthMessageStats {
   }
 
   computeDiff = () => {
-    return this.primaryNumMessages - this.peerNumMessages;
+    return Math.abs(this.primaryNumMessages - this.peerNumMessages);
   };
 
   computeDiffPercentage = () => {
@@ -42,12 +41,12 @@ class Stats {
   resultsUploadingToPrimary: HubResult<Message>[];
   primary: string;
   peer: string;
-  startTime: number;
-  stopTime: number;
+  startTime: Date;
+  stopTime: Date;
 
   constructor(
-    startTime: number,
-    stopTime: number,
+    startTime: Date,
+    stopTime: Date,
     primary: string,
     peer: string,
     syncHealthMessageStats: SyncHealthMessageStats,
@@ -105,8 +104,8 @@ class Stats {
 
   serializedSummary = () => {
     return JSON.stringify({
-      startTime: new Date(this.startTime),
-      stopTime: new Date(this.stopTime),
+      startTime: this.startTime,
+      stopTime: this.stopTime,
       primary: this.primary,
       peer: this.peer,
       primaryMessageCount: this.syncHealthMessageStats.primaryNumMessages,
@@ -127,8 +126,9 @@ class Stats {
 
 const RPC_TIMEOUT_SECONDS = 2;
 
-const getTimePrefix = (time: number) => {
-  return toFarcasterTime(time).map((farcasterTime) => {
+const getTimePrefix = (date: Date) => {
+  const unixTime = date.getTime();
+  return toFarcasterTime(unixTime).map((farcasterTime) => {
     return Buffer.from(timestampToPaddedTimestampPrefix(farcasterTime));
   });
 };
@@ -162,7 +162,7 @@ const traverseRange = async (
   startTimePrefix: Buffer,
   stopTimePrefix: Buffer,
   rpcClient: HubRpcClient,
-  f: (node: TrieNodeMetadataResponse) => undefined,
+  onTrieNode: (node: TrieNodeMetadataResponse) => undefined,
 ) => {
   const metadata = await getMetadata(Buffer.from(node.prefix), rpcClient);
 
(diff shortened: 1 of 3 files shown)


apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+    this.primaryNumMessages = primaryNumMessages;
+    this.peerNumMessages = peerNumMessages;
+  }
+
+  computeDiff = () => {
+    return this.primaryNumMessages - this.peerNumMessages;
+  };
+
+  computeDiffPercentage = () => {
+    return this.computeDiff() / this.primaryNumMessages;
```
sanjayprabhu, line 35: Do we need to take the absolute here?

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+const isPrefix = (prefix1: Buffer, prefix2: Buffer): boolean => {
+  return prefix1.toString().startsWith(prefix2.toString());
+};
+
+const traverseRange = async (
+  node: TrieNodeMetadataResponse,
+  startTimePrefix: Buffer,
+  stopTimePrefix: Buffer,
+  rpcClient: HubRpcClient,
+  f: (node: TrieNodeMetadataResponse) => undefined,
```
sanjayprabhu, line 165: Would name this something clearer. `onTrieNode` or similar

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+
+  for (const child of metadata.value.children) {
+    const childValue = Buffer.from(child.prefix);
+    if (Buffer.compare(childValue, startTimePrefix) === 0) {
+      f(child);
+    } else if (isPrefix(startTimePrefix, childValue) || isPrefix(stopTimePrefix, childValue)) {
+      await traverseRange(child, startTimePrefix, stopTimePrefix, rpcClient, f);
+    } else if (Buffer.compare(childValue, startTimePrefix) === 1 && Buffer.compare(childValue, stopTimePrefix) === -1) {
+      f(child);
+    }
```
sanjayprabhu, line 181: Add comments for each branch explaining what we're checking for

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+      numErrorToPrimary: this.errorResults("Primary").length,
+      successTypesToPrimary: this.successTypes("Primary"),
+      errorMessagesToPrimary: this.errorReasons("Primary"),
+    });
+  };
+}
+
+const RPC_TIMEOUT_SECONDS = 2;
+
+const getTimePrefix = (time: number) => {
```
sanjayprabhu, line 130: I would call this something like `timestampToPrefix`. And clarify in a comment that time is seconds and not using farcaster time (or maybe accept a date object if it's simpler)

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+  const stopTime = parseTime(stopTimeOfDay);
+
+  if (startTime === undefined || stopTime === undefined) {
+    console.log("Unable to parse time, must specify as HH:MM:SS");
+    return;
+  }
+
+  console.log("Start time", new Date(startTime), "Stop time", stopTimeOfDay, new Date(stopTime));
+
+  const primaryRpcClient = getSSLHubRpcClient(primaryNode);
```
sanjayprabhu, line 451: I think we need to handle both the secure and insecure case. This will fail if the primary hub doesn't support TLS

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+            // Sync health is us relative to peer. If the sync health is high, means we have more messages. If it's low, we have less.
+            const score = syncHealthStats.value.computeDiff();
+
+            // Useful to see progress
+            console.log("Computed sync health score", score);
+
+            let aggregateStats;
+            if (score !== 0) {
+              console.log("Investigating diff");
+              const result = await investigateDiff(primaryRpcClient, peerRpcClient, startTime, stopTime);
```
sanjayprabhu, line 503: We should make this optional. i.e. the script is read only by default, and only prints the sync health number. But if we pass in the option to write to a file, it can do the investigation and provide a more detailed output

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+    results.push(result);
+  }
+
+  return ok(results);
+};
+
+const uniqueSyncIds = (mySyncIds: Uint8Array[], otherSyncIds: Uint8Array[]) => {
+  const idsOnlyInPrimary = [];
+
+  // This is really slow. It's n^2 in the number of sync ids. It seems somwhat complicated to figure out how to hash a sync id or get a string representation that can be hashed.
```
sanjayprabhu, line 368: You can convert the sync ids to hex strings, add them to sets and subtract one from the other. No need to hash them, they are already unique

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -0,0 +1,539 @@
...
+
+const computeSyncHealthMessageStats = async (
+  startTime: number,
+  stopTime: number,
+  primaryRpcClient: HubRpcClient,
+  peer2RpcClient: HubRpcClient,
+  getNumMessagesInSpan: (rpcClient: HubRpcClient, startTime: number, stopTime: number) => Promise<HubResult<number>>,
+) => {
+  const numMessagesPrimary = await getNumMessagesInSpan(primaryRpcClient, startTime, stopTime);
+  const numMessagesPeer = await getNumMessagesInSpan(peer2RpcClient, startTime, stopTime);
```
sanjayprabhu, line 272: We're using the unoptimized version? Does the optimized version still need work?

aditiharini: Ah we're only using the optimized version. This is the name of the function that's provided as a parameter. Let me update naming.

apps/hubble/src/utils/syncHealth.ts
aditiharini: Is the string conversion encoding-agnostic? That was what I wasn't able to convince myself of. I wasn't sure if some information in the buffer would be lost in the string encoding process if e.g. i used utf-8.

sanjayprabhu: Use `bytesToHexString` and it will be fine.

sanjayprabhu (approved):

Issue #2099, opened 2024-06-26
bug(hubble): TypeScript type error with clearInterval in Node.js environment


hiroyukikumazawa:
**What is the bug?**
during build, typeScript raises the following type error:
```
No overload matches this call.
Overload 1 of 2, '(intervalId: string | number | Timeout | undefined): void', gave the following error:
Argument of type 'Timer | undefined' is not assignable to parameter of type 'string | number | Timeout | undefined'.
Type 'Timer' is not assignable to type 'string | number | Timeout | undefined'.
Property '[Symbol.dispose]' is missing in type 'Timer' but required in type 'Timeout'.
Overload 2 of 2, '(id: number | undefined): void', gave the following error:
Argument of type 'Timer | undefined' is not assignable to parameter of type 'number | undefined'.
Type 'Timer' is not assignable to type 'number'.ts(2769)
```

sds:
Thanks for the report! Addressed in #2110.

hiroyukikumazawa:
> Thanks for the report! Addressed in #2110.

welcome

PR #2100, opened 2024-06-26
fix: timer type issue for clearInterval compatibility

hiroyukikumazawa:
refs: #2099

---

## PR-Codex overview
This PR updates the `clearInterval` method in the `stop` function of `hubble.ts` to explicitly define the type as `NodeJS.Timeout`.

### Detailed summary
- Updated `clearInterval` method to define type as `NodeJS.Timeout` for `contactTimer` in `stop` function


sds:
Thanks for opening a PR! 

Looks like that line was removed in #2110, so this is no longer necessary. Will close this as no longer needed, but thank you for your contribution.

PR #2124, opened 2024-07-05
fix: full state events should not include deletes

CassOnMars:
## Why is this change needed?

The current processing of full state events results in an unbounded number of deleted messages, making event consumption downstream unpredictable and prone to getting stuck on singular large events. Because merge events historically applied to deltas, and deltas were generally small in effect size, this was a reasonable consideration, but for full state (compact) messages, it is implicit on what is added (everything in it) and what is not (everything not in it), making the deleted messages something implicitly handled by downstream consumers.

## Merge Checklist

---

## PR-Codex overview
This PR focuses on fixing the behavior of full state events in the `store.rs` file and test cases related to compact state messages in the `linkStoreCompactState.test.ts`.

### Detailed summary
- Fixed full state events to exclude deleted messages in `store.rs`
- Updated test cases to reflect changes in handling deleted messages for compact state messages in `linkStoreCompactState.test.ts`

### DIFF
--- a/.changeset/sixty-shoes-occur.md
+++ b/.changeset/sixty-shoes-occur.md
@@ -1,0 +1,5 @@
+---
+"@farcaster/hubble": minor
+---
+
+fix: full state events should not include deleted messages
--- a/apps/hubble/src/addon/src/store/store.rs
+++ b/apps/hubble/src/addon/src/store/store.rs
@@ -325,7 +325,18 @@ pub trait StoreDef: Send + Sync {
             r#type: HubEventType::MergeMessage as i32,
             body: Some(hub_event::Body::MergeMessageBody(MergeMessageBody {
                 message: Some(message.clone()),
-                deleted_messages: merge_conflicts,
+                deleted_messages: match &message.data {
+                    Some(data) => {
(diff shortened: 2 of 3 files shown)


sanjayprabhu (commented):
I like this in theory, but we need to update shuttle to handle not relying on `deleted_messages` for compact state messages first, and let other event stream consumers know about this change and give them time to migrate. So far, we've tried hard to make the compact state messages transparent, but now developers need to be aware of how to handle it.

For shuttle, I think we need to replicate the functionality within this transaction https://github.com/farcasterxyz/hub-monorepo/blob/dab95118e0f69bbcad15c49c7a156bffd7e1e093/packages/shuttle/src/shuttle/hubEventProcessor.ts#L51-L64. We have the right abstractions, we just need the list of deleted messages from the db and we can do exactly what the existing if block is doing for those messages.

CassOnMars:
so I understand correctly, the suggestion is to first implement a deletion at the shuttle level to invoke a mass deletion _marking_ of messages transactionally for the set difference in the subset of messages under the relevant types?

sanjayprabhu:
Yes

PR #2141, opened 2024-07-09
feat: make message processor handle removals for compact state explicitly, warn on mismatching deletes

CassOnMars:
## Why is this change needed?

Follow up from review of #2124, we want to confirm first that moving the responsibility of message processing marking messages as deleted for the set difference from full-state-bearing messages is a no-op, and while we're at it, might as well combine the potential thousands of SQL updates into a single query.

## Merge Checklist

---

## PR-Codex overview
This PR introduces explicit handling of message removals for compact state in the message processor, along with warnings for mismatching deletes.

### Detailed summary
- Added handling for removals of compact state messages
- Warns on mismatching deletes
- Introduced new methods for processing compact state messages

### DIFF
--- a/packages/shuttle/src/shuttle/hubEventProcessor.ts
+++ b/packages/shuttle/src/shuttle/hubEventProcessor.ts
@@ -13,11 +13,13 @@ import {
   isVerificationAddAddressMessage,
   isVerificationRemoveMessage,
   Message,
+  MessageType,
 } from "@farcaster/hub-nodejs";
 import { DB } from "./db";
 import { MessageProcessor } from "./messageProcessor";
 import { MessageHandler, MessageState, StoreMessageOperation } from "./";
 import { log } from "../log";
+import { bytesToHex } from "../utils";
 
 export class HubEventProcessor {
   static async processHubEvent(db: DB, event: HubEvent, handler: MessageHandler) {
@@ -49,7 +51,7 @@ export class HubEventProcessor {
     wasMissed = false,
   ) {
     await db.transaction().execute(async (trx) => {
-      if (deletedMessages.length > 0) {
+      if (deletedMessages.length > 0 && (operation !== "merge" || !MessageProcessor.isCompactStateMessage(message))) {
         await Promise.all(
           deletedMessages.map(async (deletedMessage) => {
             const isNew = await MessageProcessor.storeMessage(deletedMessage, trx, "delete", log);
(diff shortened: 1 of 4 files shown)


packages/shuttle/src/shuttle/hubEventProcessor.ts
```diff
@@ -49,14 +51,32 @@ export class HubEventProcessor {
...
-      if (deletedMessages.length > 0) {
+      if (deletedMessages.length > 0 && (operation !== "merge" || !MessageProcessor.isCompactStateMessage(message))) {
         await Promise.all(
           deletedMessages.map(async (deletedMessage) => {
             const isNew = await MessageProcessor.storeMessage(deletedMessage, trx, "delete", log);
             const state = this.getMessageState(deletedMessage, "delete");
             await handler.handleMessageMerge(deletedMessage, trx, "delete", state, isNew, wasMissed);
           }),
         );
+      } else if (operation === "merge" && MessageProcessor.isCompactStateMessage(message)) {
```
sanjayprabhu, line 62: We should add a flag here so we can fall back to the old logic without downgrading the package version. Just in case there are unforeseen issues.

packages/shuttle/src/shuttle/hubEventProcessor.ts
```diff
@@ -49,14 +51,32 @@ export class HubEventProcessor {
...
+        const affectedMessages = await MessageProcessor.deleteDifferenceMessages(message, trx, log);
+        await Promise.all(
+          deletedMessages.map(async (deletedMessage) => {
```
sanjayprabhu, line 65: Should we be referencing deleted messages within this branch? If we do, then it will require another package update before we can safely deprecate populating this filed in the hub. 

If we only use it for logging and not for actually storing messages, we can deprecate the field without doing another  round of package updates

packages/shuttle/src/shuttle/messageProcessor.ts
```diff
@@ -85,4 +86,84 @@ export class MessageProcessor {
...
+      .returning(["hash"])
+      .where(this.getConflictCriteria(message))
+      .execute();
+
+    const updateSet = new Set(result.map((r) => bytesToHex(r.hash)));
+    const resultMap = new Map<`0x${string}`, boolean>();
+
+    for (const row of potentialConflicts) {
+      const hash = bytesToHex(row.hash);
+      resultMap.set(hash, updateSet.has(hash));
```
sanjayprabhu, line 131: When would the potential conflicts and actual conflicts be different? Shouldn't they always be the same?

packages/shuttle/src/shuttle/hubEventProcessor.ts
```diff
@@ -62,19 +67,14 @@ export class HubEventProcessor {
...
         await Promise.all(
-          deletedMessages.map(async (deletedMessage) => {
-            let isNew = affectedMessages.get(bytesToHex(deletedMessage.hash));
-            if (isNew === undefined) {
-              log.warn(`Message ${bytesToHex(message.hash)} was not already in set, inserting as deleted row.`);
-              isNew = await MessageProcessor.storeMessage(deletedMessage, trx, "delete", log);
-            }
+          affectedMessages.map(async (deletedMessage) => {
             const state = this.getMessageState(deletedMessage, "delete");
             await handler.handleMessageMerge(deletedMessage, trx, "delete", state, isNew, wasMissed);
```
sanjayprabhu, line 72: Hmm, I think these types are different. `deleteDifferenceMessages` is returning the `Message` row from the table, but `handleMessageMerge` expects a `Message` protobuf. I think we need to deserialize the raw bytes from the row and pass in the decoded object.

CassOnMars: I think you may have caught this in an inconsistent state between updates, it is deserializing the raw bytes: https://github.com/farcasterxyz/hub-monorepo/blob/d456b0689276d6cc2e0196d214991cbd7909cbdf/packages/shuttle/src/shuttle/messageProcessor.ts#L120

sanjayprabhu: Ah, cool. Looks like there are some test failures still? PR looks good apart from that.

CassOnMars: Yeah, working through that 🫡

PR #2142, opened 2024-07-09
feat: add hub service agreement - there will be no rewards for running a hub

Wazzymandias:
## Why is this change needed?

⚠️ THERE WILL BE NO REWARDS FOR RUNNING A HUB

## Merge Checklist

---

## PR-Codex overview
This PR adds a hub service agreement feature to the Hubble script, requiring hub operators to agree to terms before starting the hub.

### Detailed summary
- Added function `prompt_for_hub_operator_agreement` to prompt hub operators to agree to terms
- Prompt displays a warning about no rewards for running the hub
- Operators must type "Yes" to proceed
- Implemented input validation for agreement
- Integration of agreement prompt before starting or upgrading the hub

### DIFF
--- a/scripts/hubble.sh
+++ b/scripts/hubble.sh
@@ -120,6 +120,26 @@ fetch_latest_docker_compose_and_dashboard() {
     fetch_file_from_repo "$GRAFANA_INI_PATH" "grafana/grafana.ini"
 }
 
+prompt_for_hub_operator_agreement() {
+  # Check if stdin is a terminal
+  if [ -t 0 ]; then
+    while true; do
+        printf "⚠️  IMPORTANT: You will NOT get any rewards for running this hub\n"
+        printf "> Please type \"Yes\" to continue: "
+        read -r response
+        case $(printf "%s" "$response" | tr '[:upper:]' '[:lower:]') in
+            yes|y)
+                printf "✅ You have agreed to the terms of service. Proceeding with hub startup...\n"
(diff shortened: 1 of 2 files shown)


scripts/hubble.sh
```diff
@@ -543,6 +563,9 @@ if [ "$1" == "upgrade" ]; then
     # Call the function to set the COMPOSE_CMD variable
     set_compose_command
 
+    # Prompt for hub operator agreement
+    prompt_for_hub_operator_agreement
```
CassOnMars, line 567: do we want this to be confirmed with every upgrade?

Wazzymandias: good catch - i can maybe store it in the `.env` in compatible format, if that gets nuked they brought it upon themselves to get re-prompted?

sds: Probably don't want to require confirmation if it would interrupt auto-upgrading, which many folks have enabled?

Wazzymandias: i updated it to prompt if script is being run from terminal, can update if there are concerns or suggestions on approach

sds (approved):

scripts/hubble.sh
sanjayprabhu: Nice, yeah, we definitely don't want to break auto-upgrade.

sanjayprabhu (approved):

PR #2149, opened 2024-07-10
feat(shuttle): Make shuttle and example-app capable of using postgres schema other than "public"

tybook:
## Why is this change needed?

This allows using shuttle to target e.g. multi-tenant dbs. Shouldn't be a breaking change because new parameters on public functions are optional. 

## Merge Checklist

---

## PR-Codex overview
This PR enables `shuttle` and its `example-app` to use a custom PostgreSQL schema. 

### Detailed summary
- Added support for custom PostgreSQL schema in shuttle and example-app
- Updated migrations to create functions with `CREATE OR REPLACE`
- Updated dependencies in `package.json`
- Added dotenv package for environment variables
- Updated README with installation instructions for node 21
- Modified database client creation to include schema parameter
- Updated table row count estimation function to include schema parameter

### DIFF
--- a/packages/shuttle/src/example-app/app.ts
+++ b/packages/shuttle/src/example-app/app.ts
@@ -25,6 +25,7 @@ import {
   HUB_SSL,
   MAX_FID,
   POSTGRES_URL,
+  POSTGRES_SCHEMA,
   REDIS_URL,
   SHARD_INDEX,
   TOTAL_SHARDS,
@@ -40,13 +41,21 @@ const hubId = "shuttle";
 
 export class App implements MessageHandler {
   private readonly db: DB;
+  private readonly dbSchema: string;
   private hubSubscriber: HubSubscriber;
   private streamConsumer: HubEventStreamConsumer;
   public redis: RedisClient;
   private readonly hubId;
 
-  constructor(db: DB, redis: RedisClient, hubSubscriber: HubSubscriber, streamConsumer: HubEventStreamConsumer) {
+  constructor(
+    db: DB,
+    dbSchema: string,
+    redis: RedisClient,
+    hubSubscriber: HubSubscriber,
+    streamConsumer: HubEventStreamConsumer,
+  ) {
     this.db = db;
+    this.dbSchema = dbSchema;
     this.redis = redis;
     this.hubSubscriber = hubSubscriber;
     this.hubId = hubId;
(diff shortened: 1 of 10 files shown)


packages/shuttle/src/example-app/app.ts
```diff
@@ -40,13 +41,21 @@ const hubId = "shuttle";
...
+  private readonly dbSchema: string;
   private hubSubscriber: HubSubscriber;
   private streamConsumer: HubEventStreamConsumer;
   public redis: RedisClient;
   private readonly hubId;
 
-  constructor(db: DB, redis: RedisClient, hubSubscriber: HubSubscriber, streamConsumer: HubEventStreamConsumer) {
+  constructor(
+    db: DB,
+    dbSchema: string,
```
sanjayprabhu, line 52: If you could add this to the end, and default to `public`, we can make it a more transparent change to the consumers

packages/shuttle/src/example-app/app.ts
```diff
@@ -55,13 +64,14 @@ export class App implements MessageHandler {
 
   static create(
     dbUrl: string,
+    dbSchema: string,
```
sanjayprabhu, line 67: same here

sanjayprabhu (commented):
Looks good to me, would just change the function params so we don't break existing consumers.

cc @sds

packages/shuttle/src/example-app/app.ts
tybook: People shouldn't be consuming the example-app functions directly, though, no?

sds: We should probably change this to take a configuration object so order of params doesn't matter. Would be a breaking change but we could do a minor release to signal potential incompatibilities.

sds: (that change should happen in a separate PR—for now, let's just do what @sanjayprabhu suggested)

sanjayprabhu: @tybook ah, you're right. I know they are using the code, but yeah upgrading this won't break them. we're probably fine. I missed that this was within the app.

sanjayprabhu (approved):

tybook:
Just pushed fix for integration tests

sanjayprabhu (approved):

PR #2171, opened 2024-07-12
feat: Assign shards using jump consistent hashing

sds:
## Why is this change needed?

While modulo-based arithmetic for assigning shards is simple and works fine, it does not work well when the number of shards changes.

For example, when increasing the number of shards from 3 → 4 using modulo-based shard assignment, ~83% of FIDs will be assigned to a different shard. With consistent hashing, only ~25% will be assigned to a different shard.

Jump consistent hashing is a particular approach to consistent hashing that is stateless (requires no additional tracking of shard assignments), which is ideal for hubs, since we don't want hubs to have to manage per-client tracking of shards.

## Merge Checklist

---

## PR-Codex overview
This PR focuses on implementing the jump consistent hashing algorithm for shard assignment in the Hubble app. 

### Detailed summary
- Implemented `jumpConsistentHash` algorithm for shard assignment based on FID
- Updated shard assignment logic in `Server` class
- Updated tests in `eventService.test.ts` for shard-based event handling

### DIFF
--- a/.changeset/few-points-fold.md
+++ b/.changeset/few-points-fold.md
@@ -1,0 +1,5 @@
+---
+"@farcaster/hubble": minor
+---
+
+feat: Assign shards using jump consistent hashing algorithm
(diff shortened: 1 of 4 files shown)


sanjayprabhu (approved):
Separately, we might want shuttle to rewind 1-5 seconds when it resumes after restart just to be safe. Will create a ticket.

PR #2191, opened 2024-07-16
chore: upgrade libp2p to 0.44.0

varunsrin:
## Why is this change needed?

Upgrades libp2p one minor version up ([Release Notes](https://github.com/libp2p/js-libp2p/releases/tag/v0.44.0) | [Migration Guide](https://github.com/libp2p/js-libp2p/blob/main/doc/migrations/v0.43-v0.44.md))

### Breaking Change in Autodial + Connection Manager Behavior

libp2p's ConnectionManager now has a default minConnection threshold of 50. If the number of active connections is < this number, it will attempt to redial peers that are in the store. 

Since we only disconnect peers and do not remove them from our store, the naive upgrade causes us to reconnect immediately to peers unexpectedly. We now set the threshold explicitly to 0 to prevent this. We should consider also removing peers from the peer store in a future upgrade. 

## Merge Checklist

---

## PR-Codex overview
This PR upgrades `libp2p` to version 0.44.0, updates `@chainsafe/libp2p-gossipsub` to 6.2.0, and introduces `@libp2p/interface-connection-gater` version 2.0.1.

### Detailed summary
- Upgraded `libp2p` to 0.44.0
- Updated `@chainsafe/libp2p-gossipsub` to 6.2.0
- Added `@libp2p/interface-connection-gater` version 2.0.1

> The following files were skipped due to too many changes: `yarn.lock`


sanjayprabhu:
The intention behind `removePeerFromAddressBook` is to kick the peer from the network. We use this in two places:
1. To prevent connecting to "invalid" peers (not in allow list, below min version etc)
2. To prevent connecting to peers that don't have a reachable rpc port

So, I think the the fact that autodial would override this is the bug and I don't think we were aware of that when it was implemented. We should remove from the peerStore if that will fix it.

CassOnMars (approved):

sanjayprabhu (approved):

Wazzymandias (approved):

Issue #2167, opened 2024-07-12
Shuttle live service is throwing "unique constraint error" (not a bug, but can be improved imo)


chetankashetti:
**What is the bug?**
Shuttle live subscription will intermittently throws unique constraint error. 
since its throwing error, it can cause application to crash. ideally if we can ignore that error that would be great.
![image](https://github.com/user-attachments/assets/2d7b85a5-524f-4e83-95eb-ac36edcf62a2)

**How can it be reproduced?**
If same message hash is processed twice this should appear.

tybook:
Running into this as well with latest version (v0.5.3) of shuttle, with near-identical fork of the example-app. I'm not sure this is "not a bug" like the original title here says, though. 

Getting this on a near-identical fork of the example-app, with redis being completely flushed beforehand:
```
$ TABLES='profiles' yarn start start
yarn run v1.22.22
warning package.json: No license field
$ tsx src/app/app.ts start
[16:09:54.574] INFO (48515): Creating app connecting to: postgres://postgres:postgres@localhost:5432/farcaster, redis://default:@localhost:6379, 54.234.220.140:2283
[16:09:54.586] INFO (48515): Writing to tables: profiles
[16:09:54.586] INFO (48515): Starting shuttle
[16:09:54.702] INFO (48515): Migration "001_initial_migration" was executed successfully
[16:09:54.702] INFO (48515): Migration "002_casts" was executed successfully
[16:09:54.702] INFO (48515): Migration "003_profile" was executed successfully
[16:09:54.702] INFO (48515): Migrations up to date
[16:09:54.703] INFO (48515): Starting HubSubscriber shuttle
[16:09:54.719] INFO (48515): HubSubscriber shuttle connected to hub
[16:09:54.720] INFO (48515): HubSubscriber shuttle Found last hub event ID: 456197822128128
[16:09:54.722] INFO (48515): HubSubscriber shuttle subscribed to hub events (types [9,1,6,2,3], shard: 0/0)
[16:10:04.723] INFO (48515): Starting stream consumer
[16:10:04.818] INFO (48515): created message (merge) 0x3960248996caabc5dfeacb2f3cc770cbda50e55e (type 5)
[16:10:04.818] INFO (48515): created message (merge) 0x9d4f560f6dd860e20cbaf2dd72d1aaf56705b7d3 (type 5)
[...]

PR #2197, opened 2024-07-17
fix: on conflict criteria is ambiguous and cannot be used for upsert

CassOnMars:
## Why is this change needed?

Shuttle defines the messages table with two unique constraints, one for the hash, another for the triple of hash, fid and type. Postgres does not allow multiple ON CONFLICT criteria, and ambiguous conflicts are essentially surfaced as uniqueness violation failures in spite of ON CONFLICT criteria given. This change alters the merge to instead attempt insert, doing nothing on any conflict, then resolving with an update if necessary and the original conflict resolution criteria is satisfied.

Addresses Issue #2167 

## Merge Checklist

---

## PR-Codex overview
This PR updates the conflict resolution logic in `MessageProcessor` to fix ambiguity issues during upsert operations.

### Detailed summary
- Updated conflict resolution logic to use `doNothing()` method instead of complex conditions
- Improved handling of conflict criteria for upsert operations

### DIFF
--- a/packages/shuttle/src/shuttle/messageProcessor.ts
+++ b/packages/shuttle/src/shuttle/messageProcessor.ts
@@ -47,7 +47,7 @@ export class MessageProcessor {
     }
 
     // @ts-ignore
-    const result = await trx
+    let result = await trx
       .insertInto("messages")
       .values({
         fid: message.data.fid,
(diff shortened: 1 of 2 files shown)


packages/shuttle/src/shuttle/messageProcessor.ts
```diff
@@ -62,28 +62,35 @@ export class MessageProcessor {
         ...opData,
       })
       .returning(["id"])
-      .onConflict((oc) =>
-        oc
-          .columns(["hash", "fid", "type"])
```
sds, line 67: If I'm understanding the issue correctly, we could have probably replaced this line with `.columns(["hash"])` and solved the problem, but this approach is perhaps more comprehensive (though it's impossible to have a conflict on hash-fid-type without having a conflict on hash.

That said, I suppose it's technically the case that we allow two messages to have the same hash as long as they are of different types and/or from ;different FIDs.

sds (approved):

packages/shuttle/src/shuttle/messageProcessor.ts
CassOnMars: the problem is that postgres seems to handle it ambiguously when there's two conflict conditions, but you cannot bifurcate or aggregate conflict handlers. seems odd since they allow where clauses in the conflict scope

PR #2201, opened 2024-07-22
fix(core): add faker back as a dependency as it is used

### LINKED ISSUE #2031: bug(core) Faker library is bundled instead of being set as dependency
**What is the bug?**
The issue has been introduced by https://github.com/farcasterxyz/hub-monorepo/pull/961
Instead of removing the import of Faker lib, removing it from the package.json dependency when it is actually used by farcaster/core made things worse as now the entire faker lib is bundled with farcaster/core.
As a consequence, it will be imported multiple times if other modules use it, and it generates a SWC error in NextJs build (at least), as it creates a conflict of identifier.
As a consequence, projects using farcaster/core have to disable SwcMinify for NextJs compilation. This is a bad practice and will not be possible anymore with NextJs15.

You can see an example on the opensource project Herocast (cc @hellno )
https://github.com/hero-org/herocast/blob/fb5d35350aa1214e915d4f0fa0c6ed7ba1f36fea/next.config.mjs#L4

Here is a Farcaster conversation on the topic :
https://warpcast.com/patxol.eth/0x00a51f22

You can try to build Herocast without `swcMinify:false`
You will get the follosing error :
image

We are able to reproduce it on the F4R codebase as well.
It might be interresting to consider extracting from 'core' any usage of Faker, but as a first step re-introducing it as a normal module dependency should fix compiler issues.

porenes:
## Why is this change needed?

Faker has been removed from the dependencies on the farcaster/core library by mistake, as it is used in factory.ts .
Re-introducing it will prevent if from being bundled.
Fixes #2031 

## Merge Checklist

---

## PR-Codex overview
This PR re-introduces the `faker` dependency in the `@farcaster/core` package.

### Detailed summary
- Re-added `@faker-js/faker` dependency in `@farcaster/core` package.json
- Removed `@faker-js/faker` from devDependencies


porenes:
@sds I am not sure how to go further with this PR.
I need a maintainer to launch the remaining workflows

stephancill:
Bump. The frames.js render lib depends on @farcaster/core and this causes build issues in next.js - would be great to get this fixed

porenes:
Thanks for support @stephancill !
I know @hellno has the issue as well, the only option at the moment is to bypass swc minify in nextjs config, which will not be an option wiht Next 15.

sds:
Sorry for the delay, thanks for your patience. This has been released in `@farcaster/core@0.15.4`

PR #2223, opened 2024-07-29
feat: add a job to measure sync health periodically

aditiharini:
## Why is this change needed?

There have been reports of missing messages from several users. We'd like to proactively detect when hubs are missing messages and identify why. 

For reviewers: 
- The diff in `syncHealth.ts` is a bit noisy because I removed a level of indentation in `printSyncHealth`
- I did not test the job, I just tested the `computeSyncHealthMessageStats` function which does the somewhat complicated querying.
- I did not incorporate into datadog, I just added logging. I thought it made sense to make sure that the pipeline works via logging as a first step and integrate into datadog as a follow up. 
- I did not write the code to select hubs in a smart way or to snap to the 10min interval that we want. Also thought this could be a follow-up. 

## Merge Checklist

---

## PR-Codex overview
This PR adds a sync health measurement job to the `@farcaster/hubble` package.

### Detailed summary
- Added a sync health measurement job to the `@farcaster/hubble` package
- Updated functions related to sync health computation and metadata retrieval
- Implemented a scheduler for measuring sync health at intervals

### DIFF
--- a/apps/hubble/src/network/sync/syncHealthJob.ts
+++ b/apps/hubble/src/network/sync/syncHealthJob.ts
@@ -9,7 +9,7 @@ import {
 import { HubInterface } from "hubble.js";
 import { peerIdFromString } from "@libp2p/peer-id";
 import { parseAddress } from "../../utils/p2p.js";
-import { Multiaddr } from "@multiformats/multiaddr";
+import { multiaddr, Multiaddr } from "@multiformats/multiaddr";
 
 const log = logger.child({
   component: "SyncHealth",
@@ -24,7 +24,7 @@ export class MeasureSyncHealthJobScheduler {
   private _startSecondsAgo = 60 * 15;
   private _spanSeconds = 60 * 10;
   private _hub: HubInterface;
-  private _peersInScope: Multiaddr[];
+  private _peersInScope: string[];
 
   constructor(syncEngine: SyncEngine, hub: HubInterface) {
     this._metadataRetriever = new SyncEngineMetadataRetriever(syncEngine);
@@ -51,23 +51,18 @@ export class MeasureSyncHealthJobScheduler {
   }
 
   peersInScope() {
-    const extraPeers =
-      process.env["SYNC_HEALTH_PEERS"]
-        ?.split(",")
-        .map((a) => {
-          const multiaddr = parseAddress(a);
-          if (multiaddr.isErr()) {
-            logger.warn(
-              { errorCode: multiaddr.error.errCode, message: multiaddr.error.message },
-              "Couldn't parse extra sync health peer address address, ignoring",
-            );
-          }
-          return multiaddr;
-        })
-        .filter((a) => a.isOk())
-        .map((a) => a._unsafeUnwrap()) ?? [];
+    const peers = process.env["SYNC_HEALTH_PEER_IDS"]?.split(",") ?? [];
 
-    return [...extraPeers, ...this._hub.bootstrapAddrs()];
+    for (const multiaddr of this._hub.bootstrapAddrs()) {
+      const peerId = multiaddr.getPeerId();
+      if (!peerId) {
+        log.info({ multiaddr }, "Couldn't get peerid for multiaddr");
+      } else {
+        peers.push(peerId);
+      }
+    }
+
+    return peers;
   }
 
   async doJobs() {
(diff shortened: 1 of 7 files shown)


apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  status(): SchedulerStatus {
+    return this._cronTask ? "started" : "stopped";
+  }
+
+  async doJobs() {
+    log.info({}, "Starting compute sync health job");
+
+    const peerIds = Array.from(this._metadataRetriever._syncEngine.getCurrentHubPeerContacts());
```
sanjayprabhu, line 52: I'm not sure choosing  random peers each time is going to be that informative. 

We should allow users to customize which peers they care about via an env var. If the env var is not set, it can default to the bootstrap peers and this hub itself.

apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -0,0 +1,99 @@
...
+    for (const [peerId, contactInfo] of peersToContact) {
+      const rpcClient = await this._hub.getRPCClientForPeer(peerIdFromString(peerId), contactInfo);
+
+      if (rpcClient === undefined) {
+        log.info("Couldn't get rpc client, skipping peer", peerId, contactInfo);
+        continue;
+      }
+
+      const peerMetadataRetriever = new RpcMetadataRetriever(rpcClient);
+      computeSyncHealthMessageStats;
```
sanjayprabhu, line 73: There's a lot of missing test coverage. Once the PR is ready to go, I would add some tests for this file as well as for `syncHealth.ts`

apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      log.info(
+        {
+          hubNumMessages: syncHealthMessageStats.value.primaryNumMessages,
+          peerNumMessages: syncHealthMessageStats.value.peerNumMessages,
+          syncHealth: syncHealthMessageStats.value.computeDiff(),
+          syncHealthPercentage: syncHealthMessageStats.value.computeDiffPercentage(),
+          contactInfo,
+        },
+        "Computed sync health stats for peer",
```
sanjayprabhu, line 95: Include a short, unique and easily searchable string in this line so we can find this in datadog. Something like `SyncHealth` or `SyncHealthStats`

apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -0,0 +1,99 @@
...
+        continue;
+      }
+
+      log.info(
+        {
+          hubNumMessages: syncHealthMessageStats.value.primaryNumMessages,
+          peerNumMessages: syncHealthMessageStats.value.peerNumMessages,
+          syncHealth: syncHealthMessageStats.value.computeDiff(),
+          syncHealthPercentage: syncHealthMessageStats.value.computeDiffPercentage(),
+          contactInfo,
```
sanjayprabhu, line 93: Just do `peerId` no need for the entire contact info.

apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      if (syncHealthMessageStats.isErr()) {
+        log.info(syncHealthMessageStats.error, "Error computing sync stats");
+        continue;
+      }
+
+      log.info(
+        {
+          hubNumMessages: syncHealthMessageStats.value.primaryNumMessages,
+          peerNumMessages: syncHealthMessageStats.value.peerNumMessages,
```
sanjayprabhu, line 90: ourMessages and theirMessages is the existing convention

apps/hubble/src/network/sync/syncHealthJob.ts
aditiharini: I think we agreed offline that this level of testing is sufficient given the risk associated with this job. Maybe let's push this PR out with the current level of testing and revisit adding more tests later if we want?

apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -0,0 +1,130 @@
...
+          theirNumMessages: syncHealthMessageStats.value.peerNumMessages,
+          syncHealth: syncHealthMessageStats.value.computeDiff(),
+          syncHealthPercentage: syncHealthMessageStats.value.computeDiffPercentage(),
+          peerId,
+        },
+        "Computed SyncHealth stats for peer",
+      );
+    }
+
+    log.info("Finished SyncHealth job");
```
sanjayprabhu, line 128: No need for this file, since the previous line is the last one in the job anyway.

apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -0,0 +1,130 @@
...
+  status(): SchedulerStatus {
+    return this._cronTask ? "started" : "stopped";
+  }
+
+  peersInScope() {
+    const extraPeers =
+      process.env["SYNC_HEALTH_PEERS"]
+        ?.split(",")
+        .map((a) => {
+          const multiaddr = parseAddress(a);
```
sanjayprabhu, line 58: Does this need to be a multiaddr? Or does it work with just a peer id?

sanjayprabhu (approved):

sanjayprabhu (approved):

PR #2229, opened 2024-07-31
fix: Always assign on-chain events to shard 0 for linear ordering

sds:
## Why is this change needed?

Sharding works when the ordering of events in different shards are independent of each other.

In the case of onchain events and username proofs, however, these events can affect multiple FIDs at a time
(e.g. transfers) which means depending on which shard they are assigned to, they might get processed out of order.

Fix this by always using the same shard (shard 0) for onchain events + username proofs. This isn't perfect since it means shard 0 is a SPOF, but that's better than a potential race condition in the processing of events, and the volume of onchain events is much lower than regular hub events.

## Merge Checklist

---

## PR-Codex overview
This PR focuses on ensuring onchain and username proofs are always assigned to the same shard for linear ordering.

### Detailed summary
- Updated logic to assign onchain and username proofs to the same shard for linear ordering
- Added a function to check if an event is ordered
- Adjusted tests to reflect the new shard assignment logic

### DIFF
--- a/.changeset/modern-cows-argue.md
+++ b/.changeset/modern-cows-argue.md
@@ -1,0 +1,5 @@
+---
+"@farcaster/hubble": patch
+---
+
+fix: Ensure onchain + username proofs are always assigned to the same shard for linear ordering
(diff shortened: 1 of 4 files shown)


sanjayprabhu (approved):
Worth adding a test.

PR #2232, opened 2024-08-01
fix(shuttle): Add missing `await` to per-iteration `sleep()` call

tybook:
## Why is this change needed?

The fix for high CPU usage due to busy-waiting (release v0.5.9) wasn't effective because `sleep()` returns a Promise which wasn't being `await`ed, so the containing `while` loop was still iterating unchecked thousands of times per second. This should fix for good.

## Merge Checklist

---

## PR-Codex overview
This PR focuses on fixing a bug in the `shuttle` package where an `await` was missing in the `sleep()` call, potentially causing CPU thrashing.

### Detailed summary
- Added missing `await` to `sleep()` call in `eventStream.ts` to prevent CPU thrashing.

### DIFF
--- a/.changeset/fresh-dragons-compare.md
+++ b/.changeset/fresh-dragons-compare.md
@@ -1,0 +1,5 @@
+---
+"@farcaster/shuttle": patch
+---
+
+fix(shuttle): Add missing `await` to per-iteration `sleep()` call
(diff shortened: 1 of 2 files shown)


sds:
Thanks! Sorry for the back and forth.

PR #2238, opened 2024-08-04
submit missing messages via the sync health job and output richer logs

aditiharini:
## Why is this change needed?
1. The sync score masks some differences between which sync ids are on each node. 
2. It's hard to investigate sync health with the information we have right now. 

Actually querying for and submitting missing messages will reveal differences in the actual messages on the nodes and  logging the results will give us information about why nodes are out of sync. 

I also lumped in a small change to fix vscode config for rust analyzer. 

## Merge Checklist

---

## PR-Codex overview
The focus of this PR is to enhance the sync health job in the `hubble` application by adding features like comparing peers and submitting messages.

### Detailed summary
- Added options for peers, username, and password in the sync health job CLI
- Refactored metadata retrievers to support hub interactions
- Implemented message submission and processing in the sync health job
- Enhanced sync health computation and message stats gathering

### DIFF
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -958,6 +958,8 @@ app
   .option("--primary-node <host:port>", "Node to measure all peers against (required)", "hoyt.farcaster.xyz:2283")
   .option("--outfile <filename>", "File to output measurements to", "health.out")
   .option("--peers <ip:port,...>", "Peers to compare with (default: pick random connected peers)")
+  .option("--username <username>", "Username for primary node")
+  .option("--password <password>", "Password for primary node")
   .action(async (cliOptions) => {
     await printSyncHealth(
       cliOptions.startTimeOfday,
@@ -966,6 +968,8 @@ app
       cliOptions.primaryNode,
       cliOptions.outfile,
       cliOptions.peers ? cliOptions.peers.split(",") : undefined,
+      cliOptions.username,
+      cliOptions.password,
     );
   });
 
--- a/apps/hubble/src/cli.ts
+++ b/apps/hubble/src/cli.ts
@@ -957,6 +957,7 @@ app
   .option("--max-num-peers <count>", "Maximum number of peers to measure for", "20")
   .option("--primary-node <host:port>", "Node to measure all peers against (required)", "hoyt.farcaster.xyz:2283")
   .option("--outfile <filename>", "File to output measurements to", "health.out")
+  .option("--peers <ip:port,...>", "Peers to compare with (default: pick random connected peers)")
   .action(async (cliOptions) => {
     await printSyncHealth(
       cliOptions.startTimeOfday,
@@ -964,6 +965,7 @@ app
       cliOptions.maxNumPeers,
       cliOptions.primaryNode,
       cliOptions.outfile,
+      cliOptions.peers ? cliOptions.peers.split(",") : undefined,
     );
   });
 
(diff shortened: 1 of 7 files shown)


apps/hubble/src/cli.ts
```diff
@@ -957,13 +957,19 @@ app
   .option("--max-num-peers <count>", "Maximum number of peers to measure for", "20")
   .option("--primary-node <host:port>", "Node to measure all peers against (required)", "hoyt.farcaster.xyz:2283")
   .option("--outfile <filename>", "File to output measurements to", "health.out")
+  .option("--peers <ip:port,...>", "Peers to compare with (default: pick random connected peers)")
+  .option("--username <username>", "Username for primary node")
+  .option("--password <password>", "Password for primary node")
```
sanjayprabhu, line 962: I wouldn't add this to the cli options. This is only for debugging. Most hub runner shouldn't have to care about this

apps/hubble/src/network/sync/syncHealthJob.ts
```diff
@@ -101,12 +130,64 @@ export class MeasureSyncHealthJobScheduler {
...
+        new Date(startTime),
+        new Date(stopTime),
+      );
+
+      if (syncIds.isErr()) {
+        log.info({ error: syncIds.error }, "Error computing diverging sync ids");
+        continue;
+      }
+
+      const resultsPushingToUs = await tryPushingMissingMessages(
```
sanjayprabhu, line 145: Not sure about submitting to the remote peer. I think almost all the peers we currently care about have auth enabled, so it will fail. When this is deployed, ~8K hubs will end up submitting to the same 3 bootstrap nodes. I think we should only submit to ourselves. The remote peer, when it run it's own job will handle it's side of things.

It's fine to enable this for debugging, but it should not be turned on for all hubs.

apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts
```diff
@@ -780,7 +785,23 @@ describe("Multi peer sync engine", () => {
...
+      await divergingSyncIds(metadataRetriever1, metadataRetriever2, start, stop)
+    )._unsafeUnwrap();
+
+    // There's only one message that's uniquely on engine 1. This is consistent with the message counts.
+    expect(divergingSyncIds2.idsOnlyInPrimary.length).toEqual(1);
+    expect(divergingSyncIds2.idsOnlyInPeer.length).toEqual(0);
+
+    // Show that pushing missing messages works
+    const pushResults2 = (
+      await tryPushingMissingMessages(metadataRetriever1, metadataRetriever2, divergingSyncIds2.idsOnlyInPrimary)
```
sanjayprabhu, line 798: Calling individual functions like this in an integration test is a sign that we're missing an abstraction. I would pull out a class that contains all this logic and call functions on that class instead. Ideally, it should encapsulate most of the logic in the SyncHealthJob.

apps/hubble/src/cli.ts
aditiharini: Hm, I've used this functionality repeatedly for debugging (not just for this feature). I think that it's difficult to maintain debugging functionality in a branch that's not checked in. That branch needs to stay rebased and will grow over time. If we think that others can use this for debugging, is there value in checking it in?

apps/hubble/src/network/sync/syncHealthJob.ts
aditiharini: `resultsPushingToPeer` is where we push to the remote peer and I gated it by keeping track of which peers we get auth errors from and not repeatedly trying to submit to those. This sounds like it should solve the problem of a bunch of hubs trying to submit to our auth-gated hubs and getting rejected. 

Neynar seems to not have auth enabled and I've gotten useful signal from submitting to Neynar. Are we concerned about the performance implications of the submissions caused by this job?

aditiharini: I'll get rid of the code for pushing to the peer. I think we can add this in a follow-up if we want it.

apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts
aditiharini: Resolved, agreed the class is nicer!

sanjayprabhu (approved):

PR #2240, opened 2024-08-07
fix: fix sync id retrieval for sync health job

aditiharini:
## Why is this change needed?

The `getAllSyncIdsByPrefix` rpc returns a max of 1024 sync ids and we weren't accounting for the fact that the list returned by the rpc might be truncated. As a result, we weren't looking at all relevant sync ids when trying to submit missing messages. 

I lumped some cleanups/improvements to the sync health command in too and an update to the vscode config for formatting rust code. 

## Merge Checklist

---

## PR-Codex overview
This PR focuses on enhancing sync health tracking in the `hubble` app. 

### Detailed summary
- Added `MAX_VALUES_RETURNED_PER_SYNC_ID_REQUEST` constant
- Updated `trie_node.rs` and `server.ts` with new constant value
- Modified sync health computations and message stats
- Introduced `sync-health` source for data submission
- Improved peer selection logic for sync health
- Enhanced handling of sync IDs with a large number of entries

### DIFF
--- a/apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts
+++ b/apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts
@@ -75,9 +75,9 @@ beforeAll(async () => {
 
 describe("Multi peer sync engine", () => {
   jest.setTimeout(TEST_TIMEOUT_LONG);
-  const addMessagesWithTimeDelta = async (engine: Engine, timeDelta: number[]) => {
+  const addMessagesWithTimeDelta = async (engine: Engine, timeDelta: number[], startFarcasterTime?: number) => {
     // Take care that the farcasterTime is not in the future
-    const farcasterTime = getFarcasterTime()._unsafeUnwrap() - 1000;
+    const farcasterTime = startFarcasterTime ? startFarcasterTime : getFarcasterTime()._unsafeUnwrap() - 1000;
 
     return await Promise.all(
       timeDelta.map(async (t) => {
@@ -741,7 +741,7 @@ describe("Multi peer sync engine", () => {
     }
   });
 
-  test("perform sync health computation", async () => {
+  test("sync health: basic", async () => {
     const metadataRetriever1 = new SyncEngineMetadataRetriever(hub1, syncEngine1);
     const metadataRetriever2 = new SyncEngineMetadataRetriever(hub2, syncEngine2);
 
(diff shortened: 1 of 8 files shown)


apps/hubble/src/utils/syncHealth.ts
```diff
@@ -343,14 +348,50 @@ const pickPeers = async (rpcClient: HubRpcClient, count: number): Promise<HubRes
...
+  }
+
+  const metadata = await metadataRetriever.getMetadata(Buffer.from(prefix));
+
+  if (metadata.isErr()) {
+    return err(metadata.error);
+  }
+
+  // We need to do this weird hack because the length of the results of [getAllSyncIdsByPrefix] is capped at 1024.
+  if (syncIds.value.syncIds.length === metadata.value.numMessages) {
```
sanjayprabhu, line 376: Might be cleaner to do the metadata call first, and then if the num messages is > 1024 recurse in, otherwise call getAllSyncIdsByPrefix.

Right now, for large nodes, we ignore the results of the first call.

apps/hubble/src/rpc/server.ts
```diff
@@ -76,6 +76,7 @@ export const DEFAULT_SUBSCRIBE_PERIP_LIMIT = 4; // Max 4 subscriptions per IP
 export const DEFAULT_SUBSCRIBE_GLOBAL_LIMIT = 4096; // Max 4096 subscriptions globally
 const MAX_EVENT_STREAM_SHARDS = 10;
 export const DEFAULT_SERVER_INTERNET_ADDRESS_IPV4 = "0.0.0.0";
+export const MAX_VALUES_RETURNED_PER_SYNC_ID_REQUEST = 1024;
```
sanjayprabhu, line 79: Would add a comment here and in the the rust file that this is duplicated and make sure to change it in both places.

apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts
```diff
@@ -809,6 +809,56 @@ describe("Multi peer sync engine", () => {
...
+    // Engine1 has no messages
+    await setupEngine(engine1);
+    await setupEngine(engine2);
+
+    // This is the start time from addMessagesWithTimeDelta
+    const startFarcasterTime = getFarcasterTime()._unsafeUnwrap() - 3000;
+
+    const timeDeltas = [];
+    for (let j = 0; j < 2000; j++) {
+      timeDeltas.push(j);
```
sanjayprabhu, line 833: The better way to test this is to patch the constant to have a lower number like 10, instead of adding 2K messages. This really slows down the tests.

Patching constants is also a bit annoying in tests, so cleaner way to do it is pass it in as optional argument in the constructor that defaults to the global constant but can be overridden in tests.

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -343,14 +348,50 @@ const pickPeers = async (rpcClient: HubRpcClient, count: number): Promise<HubRes
...
+  if (metadata.value.numMessages <= MAX_VALUES_RETURNED_PER_SYNC_ID_REQUEST) {
+    const syncIds = await metadataRetriever.getAllSyncIdsByPrefix(Buffer.from(prefix));
+
+    if (syncIds.isErr()) {
+      return err(syncIds.error);
+    }
+
+    return ok(syncIds.value.syncIds);
+  } else {
+    const computedSyncIds = [];
```
sanjayprabhu, line 379: Hmm, why is this branch not covered if there's an integration test for it? Is this actually being hit in tests?

apps/hubble/src/network/sync/multiPeerSyncEngine.test.ts
aditiharini: Ah, good point!

apps/hubble/src/utils/syncHealth.ts
aditiharini: Good catch, this wasn't being exercised in the test because the time range query wasn't wide enough to capture a node with a lot of messages under it.

apps/hubble/src/utils/syncHealth.ts
```diff
@@ -343,15 +348,61 @@ const pickPeers = async (rpcClient: HubRpcClient, count: number): Promise<HubRes
...
         return;
       });
   });
 };
 
-const computeSyncIdsInSpan = async (metadataRetriever: MetadataRetriever, startTime: Date, stopTime: Date) => {
+const computeSyncIdsUnderPrefix = async (
+  metadataRetriever: MetadataRetriever,
+  prefix: Uint8Array,
+  maxValuesReturnedPerSyncIdRequest: number,
```
sanjayprabhu, line 362: This and a few other functions could go into the sync health probe class. So you don't have to pass in these params. But not a big deal for now, can be cleaned up later.

sanjayprabhu (approved):

PR #2243, opened 2024-08-08
feat: create time-bucketed rate limits based on message timestamp

CassOnMars:
## Why is this change needed?

Rate limiting hits odd edge cases from synchronization, this alters rate limiters based on fids to bucket into specialized rate limits per message timestamps

## Merge Checklist


apps/hubble/src/utils/rateLimits.ts
```diff
@@ -8,23 +8,49 @@ export const SUBMIT_MESSAGE_RATE_LIMIT = {
...
+export function getRateLimiterForTotalMessagesByTimeBucket(
+  current: number,
+  timestamp: number,
+  scale: number,
+  totalMessages: number,
+  lowTTL = 60,
+  highTTL = 60 * 60 * 24,
+): RateLimiterAbstract {
+  // Grow the value in magnitude based on how far the message is away from the current time -
+  // is this old data being synchronized or is this spamming?
```
sanjayprabhu, line 39: For the sync issue, we could only apply the rate limit in submitMessage and submitMessageBundle when source == 'rpc' or source == 'gossip'.  And ignore rate limits for sync. 

It could open up theoretical attacks via maliciously constructed sync tries, but we already have that problem anyway. So, it could be a cleaner solution to trying to handle really wide time ranges in the the rate limiter

PR #2255, opened 2024-08-15
feat: add start and stop time to bulk rpcs with large volumes of data

aditiharini:
We want to be able to restrict the reconciliation job's queries to a time range of e.g. past 1-2 weeks to reduce the amount of read load on hubs. The most expensive part of the read operation is serializing the protobufs to send back to the callee and that's what we want to save on. 

Notes for reviewers:
- There are ways to optimize the time query by (1) using the indexes (2) starting at the common prefix of the start and stop time rather than iterating all the results for the fid. I think it's not worth pursuing these optimizations until we have evidence that we need them. The simple solution implemented here will eliminate the protobuf serialization and significantly less risky than the more optimized solutions. 
- `getUserDataByFid` and `getAllUserDataMessagesByFid` have the exact same implementation. I added time ranges to the one we classified as a bulk rpc but not the other one. We should consider eliminating one of these rpcs.  
- I didn't add a time range for the link compaction bulk query because (1) we don't query for link compaction messages in reconciliation (2) it seems like we don't store a lot of link compaction state-- each new link compaction message for an fid replaces the older one. 
- I'm not sure we need the time filter for user data, but I added it anyway. Happy to remove if it seems like a bad complexity tradeoff. 
- The rpc protocol change is backwards compatible. I copied the format for `TimestampFidRequest` from `FidRequest` and added new optional parameters. 

## Merge Checklist

---

## PR-Codex overview
This PR adds start/stop time filters for bulk queries and validation methods for time inputs.

### Detailed summary
- Added start/stop time filters for bulk queries in various modules
- Implemented validation methods for farcaster time inputs

> The following files were skipped due to too many changes: `apps/hubble/src/addon/src/store/store.rs`, `apps/hubble/src/rpc/server.ts`, `packages/hub-web/src/generated/request_response.ts`, `packages/hub-nodejs/src/generated/request_response.ts`, `packages/core/src/protobufs/generated/request_response.ts`, `packages/hub-web/src/generated/rpc.ts`, `apps/hubble/src/storage/engine/index.ts`, `packages/hub-nodejs/src/generated/rpc.ts`, `apps/hubble/src/rpc/test/bulkService.test.ts`

### DIFF
--- a/apps/hubble/src/rpc/server.ts
+++ b/apps/hubble/src/rpc/server.ts
@@ -952,7 +952,7 @@ export default class Server {
 
         const { fid, pageSize, pageToken, reverse } = call.request;
 
-        const userDataResult = await this.engine?.getUserDataByFid(fid, undefined, undefined, {
+        const userDataResult = await this.engine?.getUserDataByFid(fid, {
           pageSize,
           pageToken,
           reverse,
@@ -1202,11 +1202,16 @@ export default class Server {
         log.debug({ method: "getAllCastMessagesByFid", req: call.request }, `RPC call from ${peer}`);
 
         const { fid, pageSize, pageToken, reverse, startTimestamp, stopTimestamp } = call.request;
-        const result = await this.engine?.getAllCastMessagesByFid(fid, startTimestamp, stopTimestamp, {
-          pageSize,
-          pageToken,
-          reverse,
-        });
+        const result = await this.engine?.getAllCastMessagesByFid(
+          fid,
+          {
+            pageSize,
+            pageToken,
+            reverse,
+          },
+          startTimestamp,
+          stopTimestamp,
+        );
         result?.match(
           (page: MessagesPage<CastAddMessage | CastRemoveMessage>) => {
             callback(null, messagesPageToResponse(page));
@@ -1221,11 +1226,16 @@ export default class Server {
         log.debug({ method: "getAllReactionMessagesByFid", req: call.request }, `RPC call from ${peer}`);
 
         const { fid, pageSize, pageToken, reverse, startTimestamp, stopTimestamp } = call.request;
-        const result = await this.engine?.getAllReactionMessagesByFid(fid, startTimestamp, stopTimestamp, {
-          pageSize,
-          pageToken,
-          reverse,
-        });
+        const result = await this.engine?.getAllReactionMessagesByFid(
+          fid,
+          {
+            pageSize,
+            pageToken,
+            reverse,
+          },
+          startTimestamp,
+          stopTimestamp,
+        );
         result?.match(
           (page: MessagesPage<ReactionAddMessage | ReactionRemoveMessage>) => {
             callback(null, messagesPageToResponse(page));
@@ -1240,11 +1250,16 @@ export default class Server {
         log.debug({ method: "getAllVerificationMessagesByFid", req: call.request }, `RPC call from ${peer}`);
 
         const { fid, pageSize, pageToken, reverse, startTimestamp, stopTimestamp } = call.request;
-        const result = await this.engine?.getAllVerificationMessagesByFid(fid, startTimestamp, stopTimestamp, {
-          pageSize,
-          pageToken,
-          reverse,
-        });
+        const result = await this.engine?.getAllVerificationMessagesByFid(
+          fid,
+          {
+            pageSize,
+            pageToken,
+            reverse,
+          },
+          startTimestamp,
+          stopTimestamp,
+        );
         result?.match(
           (page: MessagesPage<VerificationAddAddressMessage | VerificationRemoveMessage>) => {
             callback(null, messagesPageToResponse(page));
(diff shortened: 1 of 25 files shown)


apps/hubble/src/storage/stores/userDataStore.ts
```diff
@@ -68,8 +68,13 @@ class UserDataStore extends RustStoreBase<UserDataAddMessage, never> {
   }
 
   /** Finds all UserDataAdd messages for an fid */
-  async getUserDataAddsByFid(fid: number, pageOptions: PageOptions = {}): Promise<MessagesPage<UserDataAddMessage>> {
-    const messages_page = await rsGetUserDataAddsByFid(this._rustStore, fid, pageOptions);
```
sanjayprabhu, line 72: I think this can also use the `getAllMessagesByFid` method from the base store?

protobufs/schemas/request_response.proto
```diff
@@ -92,6 +92,15 @@ message FidRequest {
   optional bool reverse = 4;
 }
 
+message TimestampFidRequest {
```
sanjayprabhu, line 95: Minor nit, maybe this should be `FidTimestampRequest`

apps/hubble/src/rpc/server.ts
```diff
@@ -952,7 +952,7 @@ export default class Server {
 
         const { fid, pageSize, pageToken, reverse } = call.request;
 
-        const userDataResult = await this.engine?.getUserDataByFid(fid, {
+        const userDataResult = await this.engine?.getUserDataByFid(fid, undefined, undefined, {
```
sanjayprabhu, line 955: Hmm, the order of arguments is a bit awkward. Given we almost always have page options but only rarely we do we specify timestamps, we should probably but the timestamp at the end.

apps/hubble/src/rpc/server.ts
```diff
@@ -1201,8 +1201,8 @@ export default class Server {
         const peer = Result.fromThrowable(() => call.getPeer())().unwrapOr("unknown");
         log.debug({ method: "getAllCastMessagesByFid", req: call.request }, `RPC call from ${peer}`);
 
-        const { fid, pageSize, pageToken, reverse } = call.request;
-        const result = await this.engine?.getAllCastMessagesByFid(fid, {
+        const { fid, pageSize, pageToken, reverse, startTimestamp, stopTimestamp } = call.request;
+        const result = await this.engine?.getAllCastMessagesByFid(fid, startTimestamp, stopTimestamp, {
```
sanjayprabhu, line 1205: I think we should validate that the start and stop timestamps are valid farcaster time. People might pass in unix epoch or millisecond resolution, we should catch that. Also might want to make sure passing in 0 for startTimestamp works.

apps/hubble/src/storage/stores/userDataStore.ts
aditiharini: There's one additional filter on this one and the rust api for `getAllMessagesByFid` doesn't expose a filter function. I think maybe we keep this as-is for now and introduce that refactor later on if we want to?

apps/hubble/src/rpc/server.ts
aditiharini: Good point, added the validation. I think it's not critical to make sure that 0 for start timestamp works because if someone doesn't care about start time they should leave it empty and that behavior is tested well.

aditiharini: Lmk if there are some specific reasons you think the 0 start time is a valid thing to test, happy to add tests if valuable.

apps/hubble/src/storage/engine/index.ts
```diff
@@ -854,8 +916,28 @@ class Engine extends TypedEmitter<EngineEvents> {
...
+
+      validatedStartTime = validatedStartTimeResult.value;
+    }
+
+    let validatedStopTime;
+    if (stopTime) {
+      const validatedStopTimeResult = validations.validateFarcasterTime(stopTime);
+      if (validatedStopTimeResult.isErr()) {
+        return err(validatedStopTimeResult.error);
+      }
```
sanjayprabhu, line 934: Hmm, this is copied too many times. Maybe worth adding a function for this. But not a huge deal.

packages/core/src/validations.ts
```diff
@@ -197,6 +197,22 @@ export const validateEd25519PublicKey = (publicKey?: Uint8Array | null): HubResu
   return ok(publicKey);
 };
 
+export const validateFarcasterTime = (farcasterTime: number) => {
```
sanjayprabhu, line 200: Could you add some tests please? In particular, we should make sure 0 is valid.

apps/hubble/src/rpc/server.ts
sanjayprabhu: With public API endpoints, we can't rely on what people are expected to do. We should handle adversarial cases and non-obvious cases. And 0 is actually an obvious case, where people might use it as a default value. I think it should work fine, but I would still add a test both at the validation function level and at the server.test.ts level.

aditiharini: That's a fair point. I added more tests.

sanjayprabhu (approved):

PR #2290, opened 2024-09-03
feat: add a new metric for tracking count of successfully merged messages

aditiharini:
## Why is this change needed?

We're currently using hubble.hub.merge_message.count for the message count metric and it  looks skewed because some hubs have more messages submitted as bundles than others and each bundle only count as a single message rather than as the number of individual messages inside them. Create a new metric to track successfully merged messages (we already have a metric for errors). 

## Merge Checklist

---

## PR-Codex overview
The focus of this PR is to increment a statsd metric when a message submission is successful in the `Hub` class of the Hubble app.

### Detailed summary
- Increment statsd metric on successful message submission in `submit_message` method.
- Added statsd metric incrementation in the `mergeResult` match block.

### DIFF
--- a/apps/hubble/src/hubble.ts
+++ b/apps/hubble/src/hubble.ts
@@ -1942,6 +1942,7 @@ export class Hub implements HubInterface {
             `source:${source}`,
           ];
           infoLogs.push(`[${parts.join("|")}]`);
+          statsd().increment("submit_message.success", 1, { message_type: type, source: source ?? "unknown-source" });
         },
         (e) => {
           const parts = [
@@ -1974,7 +1975,6 @@ export class Hub implements HubInterface {
     // Convert the merge results to an Array of HubResults with the key
     const finalResults: HubResult<number>[] = [];
     const finalFailures = new Map<string, number>();
-    const totalTimeMilis = Date.now() - start;
     let success = 0;
     for (let i = 0; i < allResults.size; i++) {
       const result = allResults.get(i) as HubResult<number>;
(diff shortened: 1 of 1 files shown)


apps/hubble/src/hubble.ts
```diff
@@ -1985,10 +1986,10 @@ export class Hub implements HubInterface {
         finalFailures.set(errCode, count + 1);
       }
       finalResults.push(result);
-    }
 
-    const totalTimeMilis = Date.now() - start;
-    statsd().timing("hub.merge_message", totalTimeMilis / finalResults.length);
+      // Register one metric event per message that was merged
+      statsd().timing("hub.merge_message", totalTimeMilis / allResults.size);
```
sanjayprabhu, line 1991: I think the timing should just count how much time was spent in this particular iteration of the loop now?

sanjayprabhu: Ah, I see. We should keep this outside the loop then and do `statsd().increment` on a different metric. Timing is meant to capture the time spent (it just also happens to expose a `.count` in datadog/grafana that measures the number of timing calls.

sanjayprabhu: Or, actually, we should be able to update the dashboard to use this metric https://github.com/farcasterxyz/hub-monorepo/blob/cc0d0a3ecb4bcbcaecf1ea60ad720ad354c858e8/apps/hubble/src/storage/engine/index.ts#L424

aditiharini: I don't think we can use `storage.merge_messages`. It also tracks the timing over an aggregate list of messages. R.e. the `hub.merge_message` metric-- if we're measuring latency for a single message and not the aggregate set, I feel like we should register it the appropriate number of times. But yeah, I get that it's a bit of a hack to use a timing metric fo a count. I can just create a new metric.

sanjayprabhu (approved):

PR #2298, opened 2024-09-06
feat: request missing on chain events on error

aditiharini:
## Why is this change needed?

Our hubs regularly fail to get into sync because some of them are missing onchain events. This feature will have us request missing onchain events if there are errors related to missing on chain events when messages are submitted. 

Deployed to hoyt for testing: [Logs](https://app.datadoghq.com/logs?query=%22Attempting%20to%20retryEventsForFid%22%20OR%20%22Finished%20retryEventsForFid%22%20OR%20%22cacheOnChainEvent%22%20&agg_m=count&agg_m_source=base&agg_q=%40fid&agg_q_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=time%2Cdesc&top_n=30&top_o=top&viz=stream&x_missing=true&from_ts=1726433975166&to_ts=1726433977504&live=false)
## Merge Checklist

---

## PR-Codex overview
The focus of this PR is to enhance error handling and retry mechanisms for on-chain events in the Hubble application.

### Detailed summary
- Added feature to request missing on-chain events on related submit message errors
- Improved error code handling for validation failures and unknown signers
- Implemented retry mechanism for on-chain events by fid
- Enhanced error messages and error handling logic

> The following files were skipped due to too many changes: `apps/hubble/src/storage/engine/index.ts`, `apps/hubble/src/eth/l2EventsProvider.ts`

### DIFF
--- a/apps/hubble/src/eth/l2EventsProvider.test.ts
+++ b/apps/hubble/src/eth/l2EventsProvider.test.ts
@@ -7,8 +7,15 @@ import { deployStorageRegistry, publicClient, testClient, walletClientWithAccoun
 import { accounts } from "../test/constants.js";
 import { sleep } from "../utils/crypto.js";
 import { L2EventsProvider, OptimismConstants } from "./l2EventsProvider.js";
-import { Transport } from "viem";
-import { getBlockNumber, mine, simulateContract, waitForTransactionReceipt, writeContract } from "viem/actions";
+import { CreateContractEventFilterParameters, Transport } from "viem";
+import {
+  createContractEventFilter,
+  getBlockNumber,
+  mine,
+  simulateContract,
+  waitForTransactionReceipt,
+  writeContract,
+} from "viem/actions";
 import OnChainEventStore from "../storage/stores/onChainEventStore.js";
 import StoreEventHandler from "../storage/stores/storeEventHandler.js";
 
@@ -196,6 +203,46 @@ describe("process events", () => {
       expect(events.length).toEqual(1);
       expect(events[0]?.fid).toEqual(1);
       expect(events[0]?.storageRentEventBody?.units).toEqual(1);
+    },
+    TEST_TIMEOUT_LONG,
+  );
+
+  // TODO(aditi): Impl and add tests for key registry and id registry
+  test(
+    "retry by fid",
+    async () => {
+      const rentSim = await simulateContract(publicClient, {
+        address: storageRegistryAddress,
+        abi: StorageRegistry.abi,
+        functionName: "credit",
+        account: accounts[0].address,
+        args: [BigInt(1), BigInt(1)],
+      });
+
+      const rentHash = await writeContract(walletClientWithAccount, rentSim.request);
+      const rentTrx = await waitForTransactionReceipt(publicClient, { hash: rentHash });
+      await sleep(1000); // allow time for the rent event to be polled for
+      await mine(testClient, { blocks: L2EventsProvider.numConfirmations });
+      await waitForBlock(Number(rentTrx.blockNumber) + L2EventsProvider.numConfirmations);
+
+      const events1 = await onChainEventStore.getOnChainEvents(OnChainEventType.EVENT_TYPE_STORAGE_RENT, 1);
+      expect(events1.length).toEqual(1);
+      expect(events1[0]?.fid).toEqual(1);
+      expect(events1[0]?.storageRentEventBody?.units).toEqual(1);
+
+      await OnChainEventStore.clearEvents(db);
+      const events2 = await onChainEventStore.getOnChainEvents(OnChainEventType.EVENT_TYPE_STORAGE_RENT, 1);
+
+      expect(events2.length).toEqual(0);
+
+      await l2EventsProvider.retryEventsForFid(1);
+      await sleep(1000); // allow time for the rent event to be polled for
+
+      const events3 = await onChainEventStore.getOnChainEvents(OnChainEventType.EVENT_TYPE_STORAGE_RENT, 1);
+
+      expect(events3.length).toEqual(1);
+      expect(events3[0]?.fid).toEqual(1);
+      expect(events3[0]?.storageRentEventBody?.units).toEqual(1);
     },
     TEST_TIMEOUT_LONG,
   );
--- a/apps/hubble/src/eth/l2EventsProvider.test.ts
+++ b/apps/hubble/src/eth/l2EventsProvider.test.ts
@@ -200,7 +200,6 @@ describe("process events", () => {
     TEST_TIMEOUT_LONG,
   );
 
-  // TODO(aditi): It's pretty high overhead to set up tests with the id registry and key registry -- need to deploy contracts (need bytecode). Testing via the storage contract tests most of the meaningful logic.
   test(
     "retry by fid",
     async () => {
--- a/apps/hubble/src/eth/l2EventsProvider.test.ts
+++ b/apps/hubble/src/eth/l2EventsProvider.test.ts
@@ -7,15 +7,8 @@ import { deployStorageRegistry, publicClient, testClient, walletClientWithAccoun
 import { accounts } from "../test/constants.js";
 import { sleep } from "../utils/crypto.js";
 import { L2EventsProvider, OptimismConstants } from "./l2EventsProvider.js";
-import { CreateContractEventFilterParameters, Transport } from "viem";
-import {
-  createContractEventFilter,
-  getBlockNumber,
-  mine,
-  simulateContract,
-  waitForTransactionReceipt,
-  writeContract,
-} from "viem/actions";
+import { Transport } from "viem";
+import { getBlockNumber, mine, simulateContract, waitForTransactionReceipt, writeContract } from "viem/actions";
 import OnChainEventStore from "../storage/stores/onChainEventStore.js";
 import StoreEventHandler from "../storage/stores/storeEventHandler.js";
 
@@ -207,7 +200,7 @@ describe("process events", () => {
     TEST_TIMEOUT_LONG,
   );
 
-  // TODO(aditi): Impl and add tests for key registry and id registry
+  // TODO(aditi): It's pretty high overhead to set up tests with the id registry and key registry -- need to deploy contracts (need bytecode). Testing via the storage contract tests most of the meaningful logic.
   test(
     "retry by fid",
     async () => {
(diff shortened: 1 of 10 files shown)


apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -661,6 +669,39 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
     }
   }
 
+  public async retryEventsForFid(fid: number) {
+    if (this._fidRetryDedupMap.has(fid)) {
+      return;
+    }
+
+    // Let's never remove. No need to retry same fid many times. If we allow retrying multiple times, we need to rate limit.
+    this._fidRetryDedupMap.set(fid, true);
```
sanjayprabhu, line 678: We should have some limit on this either by size or by time. Otherwise it could bloat memory.

apps/hubble/src/storage/engine/index.ts
```diff
@@ -332,6 +336,18 @@ class Engine extends TypedEmitter<EngineEvents> {
...
+      if (result.isErr()) {
+        // Try to request on chain event if it's missing
+        if (
+          result.error.errCode === "bad_request.no_storage" ||
+          "bad_request.missing_signer" ||
+          "bad_request.missing_fid"
+        ) {
+          // TODO(aditi): Add timeout
+          // TODO(aditi): Do we just want to request all or only the appropriate event for the error message. Seems worth just requesting all. Simplifies the dedup logic too.
+          await this._l2EventsProvider?.retryEventsForFid(fid);
```
sanjayprabhu, line 348: We should not await here, it will block merging other messages.

packages/core/src/errors.ts
```diff
@@ -69,9 +69,12 @@ export type HubErrorCode =
   | "bad_request.parse_failure"
   | "bad_request.invalid_param"
   | "bad_request.validation_failure"
+  | "bad_request.missing_signer"
   | "bad_request.duplicate"
   | "bad_request.conflict"
   | "bad_request.prunable"
+  | "bad_request.no_storage"
+  | "bad_request.missing_fid"
```
sanjayprabhu, line 77: "invalid" or "unknown" might be better than "missing"

apps/hubble/src/eth/l2EventsProvider.test.ts
```diff
@@ -199,4 +199,53 @@ describe("process events", () => {
     },
     TEST_TIMEOUT_LONG,
   );
+
+  // TODO(aditi): It's pretty high overhead to set up tests with the id registry and key registry -- need to deploy contracts (need bytecode). Testing via the storage contract tests most of the meaningful logic.
```
sanjayprabhu, line 203: I think this is fine for now. The overhead is worth it if it can be shared with other tests, but until then this is fine.

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -661,6 +669,39 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
...
+
+    try {
+      // The viem API requires an event kind if you want to provide filters by indexed arguments
+      // TODO(aditi): Do we want batching?
+      await this.syncHistoricalEvents(0, this._lastBlockNumber, this._lastBlockNumber, {
```
sanjayprabhu, line 683: This is a really large batch size, and I'm not sure alchemy or infura would allow such high number. I would hardcode to 100 (assuming the result is ordered reverse chronologically, please verify).

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -661,6 +669,39 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
...
+        eventKind: "Storage",
+        eventName: "Rent",
+        fid,
+      });
+      // TODO(aditi): Do we want to do more event types
```
sanjayprabhu, line 688: Actually, we should do all events. Don't filter by event name when fetching.

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -737,11 +778,81 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
...
+      toBlock: toBlockBigInt,
+      eventName: eventSpecificArgs?.eventName,
+      args: {
+        fid: fidBigInt,
+      },
+      strict: true,
+    };
+  }
+
+  // TODO(aditi): Make sure we don't make multiple requests for a single fid
```
sanjayprabhu, line 808: Hmm, this is not a problem now, but when you remove the await when calling the retry function, it will become an issue. You might need to add a lock around the duplicate check. I would add a test case when submitting 10 messages via mergeMessages for the same fid, it only calls retry once.

apps/hubble/src/storage/engine/index.ts
```diff
@@ -332,6 +336,18 @@ class Engine extends TypedEmitter<EngineEvents> {
...
         consumeRateLimitByKey(`${fid}`, limiter);
       }
+      if (result.isErr()) {
+        // Try to request on chain event if it's missing
+        if (
+          result.error.errCode === "bad_request.no_storage" ||
+          "bad_request.missing_signer" ||
+          "bad_request.missing_fid"
+        ) {
+          // TODO(aditi): Add timeout
```
sanjayprabhu, line 346: Not at this level. Should be within the provider.

apps/hubble/src/storage/engine/index.ts
```diff
@@ -332,6 +336,18 @@ class Engine extends TypedEmitter<EngineEvents> {
...
+          // TODO(aditi): Do we just want to request all or only the appropriate event for the error message. Seems worth just requesting all. Simplifies the dedup logic too.
```
sanjayprabhu, line 347: Yes, all is better

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -661,6 +669,39 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
     }
   }
 
+  public async retryEventsForFid(fid: number) {
```
sanjayprabhu, line 672: Worth adding a metric and a log line for events successfully processed as part of a retry

apps/hubble/src/eth/l2EventsProvider.ts
aditiharini: Discussed offline-- added comments.

apps/hubble/src/eth/l2EventsProvider.ts
aditiharini: Discussed offline-- added other common event kinds.

apps/hubble/src/storage/engine/index.ts
aditiharini: Would you be able to double check my approach on setting the timeout? I honestly couldn't find a lot of other examples in the codebase and this approach feels a bit awkard/imprecise.

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -671,34 +677,60 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
 
   public async retryEventsForFid(fid: number) {
     if (this._fidRetryDedupMap.has(fid)) {
-      return;
+      return; //
```
sanjayprabhu, line 680: Empty comment?

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -671,34 +677,60 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
...
         eventKind: "KeyRegistry",
         eventName: "Add",
         fid,
       });
+
+      await this.syncHistoricalEvents(0, this._lastBlockNumber, this._lastBlockNumber, {
+        eventKind: "KeyRegistry",
+        eventName: "Remove",
+        fid,
+      });
```
sanjayprabhu, line 727: Let's just put this in a dict like `{"KeyRegistry": ["Add", "Remove"]}` and iterate instead of duplicating the lines.

apps/hubble/src/eth/l2EventsProvider.ts
```diff
@@ -671,34 +677,60 @@ export class L2EventsProvider<chain extends Chain = Chain, transport extends Tra
...
 
+    log.info({ fid }, `Attempting to retryEventsForFid ${fid}`);
+    statsd().increment("l2events.retriesByFid.attempts", 1, { fid: fid.toString() });
+
+    // Let's not remove. No need to retry same fid many times. If we allow retrying multiple times, we need to rate limit. The map will get cleared periodically to prevent memory bloat.
     try {
-      // The viem API requires an event kind if you want to provide filters by indexed arguments
-      // TODO(aditi): Do we want batching?
+      // Set a timeout of 10 seconds per retry. We don't expect a lot of retries.
+      setTimeout(() => {
```
sanjayprabhu, line 691: Ah, yeah, this is not clean. Better to use the timeout within the viem api. We should already have one actually. Just verify it will be respected.

aditiharini: There's a default timeout of 10s at the transport level. We haven't overridden that.  I think this is fine for now. 

```
    const opMainnetRpcUrls = options.l2RpcUrl.split(",");
    const opTransports = opMainnetRpcUrls.map((url) => http(url, { retryCount: 3, retryDelay: 500 }));
    const opClient = createPublicClient({
      chain: optimism,
      transport: fallback(opTransports, { rank: options.rankRpcs ?? false }),
    });
```

apps/hubble/src/eth/l2EventsProvider.ts
aditiharini: This is currently pretty hard to test. We mock the l2 events provider in most tests -- which I think is a good practice. The l2 events tests are slow and the setup for each test is really complicated. It's hard to test the logic 

I don't think I understand the failure mode here though. All this code is in JS-land so we don't need to worry about concurrency issues right? Are you worried about the JS runtime switching to a new task between the `has` and `set` calls? I also don't think it's the biggest issue if we end up with multiple concurrent calls.

sanjayprabhu (approved):

apps/hubble/src/storage/engine/index.ts
```diff
@@ -279,45 +283,58 @@ class Engine extends TypedEmitter<EngineEvents> {
...
-          mergeResults.set(i, err(new HubError("unavailable", `rate limit exceeded for FID ${fid}`)));
-          return;
+        const result = await this.computeMergeResult(message, i);
+        if (result.isErr()) {
+          mergeResults.set(i, result);
+          // Try to request on chain event if it's missing
+          if (
+            result.error.errCode === "bad_request.no_storage" ||
+            "bad_request.unknown_signer" ||
+            "bad_request.missing_fid"
```
carmen0208, line 329: Q: Should that be `bad_request.unknown_fid` as the name suggested in line 1237

PR #2335, opened 2024-09-26
fix: Ignore local/loopback traffic in IP connection limiter

sds:
## Why is this change needed?

This allows the use of reverse proxies.

## Merge Checklist

---

## PR-Codex overview
This PR focuses on enhancing the connection limiting functionality by ignoring local loopback IP traffic and modifying a test case.

### Detailed summary
- Updated `rateLimits.ts` to ignore both `127.0.0.1` and `::1` for local loopback traffic.
- Changed the connection limit check in `server.ts` to exclude local IPs from the limit enforcement.
- Modified the test in `eventService.test.ts` to skip the test case for excessive subscriptions.

### DIFF
--- a/apps/hubble/src/rpc/server.ts
+++ b/apps/hubble/src/rpc/server.ts
@@ -319,7 +319,7 @@ class IpConnectionLimiter {
     const ip = extractIPAddress(peerString) ?? "unknown";
 
     const connections = this.ipConnections.get(ip) ?? 0;
-    if (connections >= this.perIpLimit) {
+    if (ip !== "127.0.0.1" && ip !== "::1" && connections >= this.perIpLimit) {
       return err(new Error(`Too many connections from this IP: ${ip}`));
     }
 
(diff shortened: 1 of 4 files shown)


sanjayprabhu (approved):

apps/hubble/src/rpc/server.ts
```diff
@@ -319,7 +319,7 @@ class IpConnectionLimiter {
     const ip = extractIPAddress(peerString) ?? "unknown";
 
     const connections = this.ipConnections.get(ip) ?? 0;
-    if (connections >= this.perIpLimit) {
+    if (ip !== "127.0.0.1" && connections >= this.perIpLimit) {
```
sanjayprabhu, line 322: What about ipv6?

sds: It's a fair point, I just don't think people are using IPV6 for loopback communication. I'll follow up however.

PR #2341, opened 2024-09-27
feat(shuttle): Support deleting messages on DB that are missing on hub

tybook:
There was existing support for handling messages that were present on hub and missing on DB, but not the opposite. From talking with Sanjay, it should be correct/safe to delete these messages from the DB.

## Why is this change needed?

Shuttle only reconciled in one direction, even though everything appears 95% written to support both directions. This completes that. I know from adding logging to our Shuttle-powered app that that are a significant number of messages present in our DB that are missing in the hub. 

## Merge Checklist

---

## PR-Codex overview
This PR introduces functionality to delete messages from the database that are not present on the hub, enhancing the message handling logic in the `shuttle` package.

### Detailed summary
- Updated `handleMissingMessage` method in `HubEventProcessor` to accept an optional `missingInHub` parameter.
- Added logic to delete messages if `missingInHub` is true; otherwise, it merges them.
- Adjusted calls to `handleMissingMessage` in `app.ts` to pass the new parameter.

### DIFF
--- a/packages/shuttle/src/example-app/app.ts
+++ b/packages/shuttle/src/example-app/app.ts
@@ -220,7 +220,7 @@ export class App implements MessageHandler {
         fid,
         async (message, missingInDb, prunedInDb, revokedInDb) => {
           if (missingInDb) {
-            await HubEventProcessor.handleMissingMessage(this.db, message, this);
+            await HubEventProcessor.handleMissingMessage(this.db, message, this, !missingInDb);
           } else if (prunedInDb || revokedInDb) {
             const messageDesc = prunedInDb ? "pruned" : revokedInDb ? "revoked" : "existing";
             log.info(`Reconciled ${messageDesc} message ${bytesToHexString(message.hash)._unsafeUnwrap()}`);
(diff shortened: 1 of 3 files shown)


packages/shuttle/src/example-app/app.ts
```diff
@@ -220,15 +220,15 @@ export class App implements MessageHandler {
         fid,
         async (message, missingInDb, prunedInDb, revokedInDb) => {
           if (missingInDb) {
-            await HubEventProcessor.handleMissingMessage(this.db, message, this);
+            await HubEventProcessor.handleMissingMessage(this.db, message, this, !missingInDb);
```
sanjayprabhu, line 223: Why `!missingInDb`? Shouldn't this always be false?

tybook: `!missinInDb` is always `false` here (see the `if` condition above). I thought putting it this way would be a bit of self-documentation but if it's just confusing I can replace with `false`.

tybook: Fixed

sanjayprabhu (approved):

PR #2347, opened 2024-10-02
fix: add request length limit for getAllMessagesBySyncIds

aditiharini:
We've observed hub memory usage and disk reads spike over the last couple days. We'd like to put a size bound on `getAllMessagesBySyncIds` because it currently accepts an unlimited number of sync ids and could end up exhausting hub resources. 

## Merge Checklist

---

## PR-Codex overview
This PR focuses on adding a request length limit for the `getAllMessagesBySyncIds` function in the `server.ts` file and refactoring the message retrieval logic in `syncHealth.ts` to handle multiple requests efficiently.

### Detailed summary
- Added a check for the length of `syncIds` in `getAllMessagesBySyncIds` to enforce a maximum limit.
- Changed the message retrieval logic in `syncHealth.ts` to accumulate messages from multiple requests into `allMessages`.
- Updated the error handling to ensure consistent error responses.

### DIFF
--- a/apps/hubble/src/rpc/server.ts
+++ b/apps/hubble/src/rpc/server.ts
@@ -719,6 +719,10 @@ export default class Server {
   }
 
   public async getAllMessagesBySyncIds(request: SyncIds) {
+    if (request.syncIds.length > MAX_VALUES_RETURNED_PER_SYNC_ID_REQUEST) {
+      return err(new HubError("bad_request.validation_failure", "Too many sync ids provided"));
+    }
+
     const syncIds = request.syncIds.map((syncId) => SyncId.fromBytes(syncId));
     const messagesResult = await this.syncEngine?.getAllMessagesBySyncIds(syncIds);
     if (messagesResult?.isErr()) {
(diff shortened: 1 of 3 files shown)


apps/hubble/src/rpc/server.ts
```diff
@@ -719,6 +719,10 @@ export default class Server {
   }
 
   public async getAllMessagesBySyncIds(request: SyncIds) {
+    if (request.syncIds.length > MAX_VALUES_RETURNED_PER_SYNC_ID_REQUEST) {
```
sanjayprabhu, line 722: We should also do this for `getAllSyncIdsByPrefix`, and audit the other RPC to see if they are vulnerable to the same problem.

aditiharini: `getAllSyncIdsByPrefix` already has a limit that we enforce internally in `TrieNode.get_all_values`. Audited the others and all of them seem fine. The bulk rpcs have pagination and some page size limit so we don't need to do this.

PR #2371, opened 2024-10-21
chore: introduce a pattern to help standardize log tag names 

aditiharini:
It's hard to filter logs by tag if we name the same tag different things in different places. Introduce a pattern for constructing logs that pushes us toward standardizing log tag names. 

## Merge Checklist

---

## PR-Codex overview
This PR focuses on standardizing log tag names by introducing a `Tags` class, which is used throughout the codebase to create consistent log entries.

### Detailed summary
- Introduced `Tags` class in `apps/hubble/src/utils/logger.ts` for standardized log tags.
- Updated logging in `apps/hubble/src/hubble.ts`, `apps/hubble/src/network/p2p/gossipNode.ts`, and `apps/hubble/src/network/sync/syncHealthJob.ts` to use `Tags`.
- Added `TagFields` type definition for better structure of log fields.

### DIFF
--- a/apps/hubble/src/utils/logger.ts
+++ b/apps/hubble/src/utils/logger.ts
@@ -144,6 +144,7 @@ export type TagFields = {
   [key: string]: any;
 };
 
+// Always go through this class to construct log tags so that tag names are standardized.
 export class Tags {
   private fields: TagFields;
 
--- a/apps/hubble/src/utils/logger.ts
+++ b/apps/hubble/src/utils/logger.ts
@@ -151,10 +151,6 @@ export class Tags {
     this.fields = fields;
   }
 
-  merge(tags: Tags) {
-    this.fields = { ...this.fields, ...tags.fields };
-  }
-
   toString() {
     const extraFields = Object.fromEntries(
       Object.entries(this.fields).filter(([key, value]) => {
--- a/apps/hubble/src/utils/logger.ts
+++ b/apps/hubble/src/utils/logger.ts
@@ -134,51 +134,46 @@ class BufferedLogger {
   }
 }
 
+export type TagFields = {
+  fid?: number;
+  timestamp?: number;
+  peerId?: string;
+  messageType?: MessageType;
+  message?: Message;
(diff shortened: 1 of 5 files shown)


apps/hubble/src/utils/logger.ts
```diff
@@ -134,6 +134,54 @@ class BufferedLogger {
...
+
+  addPeerId(peerId: string) {
+    return new Tags({ ...this.fields, peerId });
+  }
+
+  addErrorMessage(errMsg: string) {
+    return new Tags({ ...this.fields, errMsg });
+  }
+
+  addMessageFields(message: Message) {
```
sds, line 164: Not a strong opinion, but it might make sense to implement an adapter for the logging instance that instead of passing a `Tags` object, it accepts an options object which knows how to handle the different types you want to annotate the tags with. For example:

```ts
log.info({ fid: ..., timestamp: ..., message: ..., error: ... });
```

So you would define the `info` method like:

```ts
info(obj?: LogTags, logMsg?: string) {
  const tags = new Tags(obj); // Define logic for special handling of keys
  return this.log.info(tags.build(), logMsg);
}
```

You would also allow a subset of keys to be provided, e.g.
```ts
type LogTags = {
  message?: Partial<Message>; // Can provide any subset of fields for message
  peerId?: string;
  error?: Error;
}

...
class Tags {
  constructor(obj: LogTags) {
    ...
    // Extract `obj` into top-level keys (since `message` etc. will be deeply nested)
  }
}
```

sds (commented):
Supportive, but there's a way we can structure to not require as much verbosity when providing tags.

apps/hubble/src/utils/logger.ts
aditiharini: I think it might make sense to use the type with all the fields on it rather than the builder pattern though (which is very verbose).

sds: That seems reasonable to me. Supportive!

sds (approved):

PR #2439, opened 2024-12-13
fix: Fix compatibility issue with replaceAll for older Node.js versio…

mdqst:
## Why is this change needed?

In the previous implementation, the `replaceAll` method was used, which was introduced in ECMAScript 2021. This causes compatibility issues in older versions of Node.js or other environments that do not support this method.

To ensure broader compatibility, I have replaced `replaceAll(" ", "")` with `replace(/ /g, "")`, which uses regular expressions and works in older environments.

This change should resolve any issues for users running on older versions of Node.js or environments that do not support `replaceAll`. 

## Merge Checklist

---

## PR-Codex overview
This PR focuses on improving the way whitespace is removed from the `line` variable when checking for the undesired datasource entry. It changes the method used to replace spaces in the `line`.

### Detailed summary
- Replaced `line.replaceAll(" ", "")` with `line.replace(/ /g, "")` to remove spaces from the `line` variable.

### DIFF
--- a/apps/hubble/scripts/grafanadash.cjs
+++ b/apps/hubble/scripts/grafanadash.cjs
@@ -20,7 +20,7 @@ module.exports = function grafana() {
       lineNumber++;
 
       // Check if the current line contains the undesired datasource entry
-      if (line.includes('"datasource":') && !line.replaceAll(" ", "").includes('"datasource":"Graphite"')) {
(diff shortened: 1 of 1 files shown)


sds:
Thanks for the PR. We support Node 20+, and don't plan to support older versions.

